PR #7, opened 2015-05-29
(WIP) records + merkledag specs

jbenet:
includes also keychain types.

this is all very much WIP


merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+  bytes Data = 1; // opaque user data
+
+  // user extensions start at 50
+}
+```
+
+### Serialized Format
+
+
+(TODO remove this? use only protobuf?)
```
whyrusleeping: I think we are okay using only protobuf. since we version tag the repos and the clients and protocols, we can write in logic to handle multicodec OR protobuf later if for some reason we have to switch away from protobuf due to it being proven to be stealing money from the poor, or some other terrible thing.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+A (distributed) `record` is a piece of data meant to be transmitted and stored in various computers across a network. It carries a `value` external to the record-keeping system, which clients of the record system use. For example: a namespace record may store the value of a name. All record systems include a notion of _record validity_, which allows users of the record system to verify whether the value of a record is correct and valid under the user's circumstances. For example, a record's validity may depend upon a cryptographic signature, a range of spacetime, et-cetera.
+
+### Record System
+
+A (distributed) `record system` is a protocol which defines a method for crafting, formating, distributing, verifying, and using records over computer networks. (e.g. the Domain Name System).
+
+
+- `crafting` - construction of a record (the process of calculating the values of a record)
+- `formating` - serializing of a record into a bitstring.
```
whyrusleeping: formatting doesnt really feel like the right word here

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+// Marshal/Unmarshal specifies a way to code the record
+type Marshal(r *Record) ([]byte, error)
+type Unmarshal(r *Record, []byte) (error)
+```
+
+### Interface Example
+
+For example, Alice and Bob earlier could use the following interface:
+
+```protobuf
```
whyrusleeping: this isnt quite protobuf

whyrusleeping:
@jbenet is there anything on telling which of two records is the 'most valid' ?

keychain/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+```go
+// ProofOfControl proves a certain key is under control of a prover.
+var ProofOfControl = "proof-of-control"
+
+// ProofOfWork proves an amount of work was expended by a prover.
+var ProofOfWork = "proof-of-work"
+
+// ProofOfStorage proves certain data is possessed by prover.
+var ProofOfStorage = "proof-of-storage"
```
wking: The distinction between Proof of Storage and Proof of Retrievability wasn't immediately obvious to me.  After searching a bit I feel like I now have a better idea for how Proof of Storage works, but linking out to whatever you consider canonical docs or useful review articles for each proof type would be nice.

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+Reviewers:
+
+- [Jeromy Johnson](github.com/whyrusleeping)
+
+* * *
+
+This is the [Spec](../) for the IPFS Merkledag.
+
+## Definitions
```
wking: Can these just go into our all-IPFS glossary (#ipfs/notes#7).  Or maybe that's what this is?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `hash` - throughout this document, the word `hash` refers specifically to cryptographic hash functions, such as sha3.
+- `dag` - directed acyclic graph
+- `merkle-link` - a link (graph edge) between two objects, which is (a) represented by the hash of the target object, and (b) embedded in the source object. merkle-links construct graphs (dags) whose links are content-addressed, and authenticated.
```
wking: What does “authenticated” mean here?  Is it just “the keys are verifyable by hashing the value”?  I don't see much space between that and “content addressed”.

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `merkledag` - the merkledag is a directed acyclic graph whose links are merkle-links (hashes of the content). It is a hash tree, and (under a very lose definition) a merkle tree. Alternative names: the merkle-web, the merkle-forest, the merkle-chain.
```
wking: “lose” → “loose”.  And is the definition looseness just DAG vs. tree (i.e. number of roots)?

keychain/README.md
whyrusleeping: I also wonder where these proofs are used

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `multihash` - the [multihash](https://github.com/jbenet/multihash) format / protocol.
+- `ipfs object` - a node in the ipfs merkledag. It represents a singular entity.
```
wking: The “singular entity” sentence doesn't add anything for me.  What point were you trying to clarify with it?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `merkledag format` - the format that ipfs objects are expressed with.
+- `link segment` or `link table` - the part of the merkledag format that expresses links to other objects.
+- `data segment` - the part of the merkledag format that expresses object data.
+- `protobuf` - protocol buffers, a serialization encoding.
```
wking: Link to https://developers.google.com/protocol-buffers/?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+- `hash` - throughout this document, the word `hash` refers specifically to cryptographic hash functions, such as sha3.
+- `dag` - directed acyclic graph
+- `merkle-link` - a link (graph edge) between two objects, which is (a) represented by the hash of the target object, and (b) embedded in the source object. merkle-links construct graphs (dags) whose links are content-addressed, and authenticated.
+- `merkledag` - the merkledag is a directed acyclic graph whose links are merkle-links (hashes of the content). It is a hash tree, and (under a very lose definition) a merkle tree. Alternative names: the merkle-web, the merkle-forest, the merkle-chain.
+- `multihash` - the [multihash](https://github.com/jbenet/multihash) format / protocol.
+- `ipfs object` - a node in the ipfs merkledag. It represents a singular entity.
+- `merkledag format` - the format that ipfs objects are expressed with.
+- `link segment` or `link table` - the part of the merkledag format that expresses links to other objects.
+- `data segment` - the part of the merkledag format that expresses object data.
```
wking: To clarify “object data”, I'd prefer “the part of the merkledag format that expresses any non-link data”.

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `protobuf` - protocol buffers, a serialization encoding.
+- `multicodec` - a self-describing, generalized serialization format.
+
+## Abstract
+
+The _ipfs merkledag_ is a directed acyclic graph whose edges are merkle-links. This means that links to objects can authenticate the objects themselves, and that every object contains a secure representation of its children.
+
+This is a powerful primitive for distributed systems computations. The merkledag simplifies distributed protocols by providing an append-only authenticated datastructure. Parties can communicate and exchange secure references (merkle-links) to objects. The references are enough to verify the correctness of the object at a later time, which allows the objects themselves to be served over untrusted channels. Merkledags also allow the branching of a datastructure and subsequent merging, as in the version control system git. More generally, merkledags simplify the construction of Secure CRDTs, which enable distributed, convergent, commutative computation in an authenticated, secure way.
```
wking: Link CRDTs to http://en.wikipedia.org/wiki/Conflict-free_replicated_data_type for folks like me who aren't familiar with the acronym?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+## The Format
+
+The IPFS merkledag format is very simple. It serves as a thin waist for more complex applications and data structure transports. Therefore, it aims to be as simple and small as possible.
```
wking: “waist” → “base” or “layer”?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- `dag` - directed acyclic graph
+- `merkle-link` - a link (graph edge) between two objects, which is (a) represented by the hash of the target object, and (b) embedded in the source object. merkle-links construct graphs (dags) whose links are content-addressed, and authenticated.
+- `merkledag` - the merkledag is a directed acyclic graph whose links are merkle-links (hashes of the content). It is a hash tree, and (under a very lose definition) a merkle tree. Alternative names: the merkle-web, the merkle-forest, the merkle-chain.
+- `multihash` - the [multihash](https://github.com/jbenet/multihash) format / protocol.
+- `ipfs object` - a node in the ipfs merkledag. It represents a singular entity.
+- `merkledag format` - the format that ipfs objects are expressed with.
+- `link segment` or `link table` - the part of the merkledag format that expresses links to other objects.
+- `data segment` - the part of the merkledag format that expresses object data.
+- `protobuf` - protocol buffers, a serialization encoding.
+- `multicodec` - a self-describing, generalized serialization format.
```
wking: Is there an existing spec for this?  Searching around for “multicodec” didn't turn up any obvious hits.

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+This kind of modularity enables complicated and powerful applications to be built with little effort on top of a common base. All the complexity of authentication, distribution, replication, routing, and transport can be pulled in from other protocols and tools. This type of modularity is what made the layered internet -- the TCP/IP stack -- so tremendously powerful.
+
+### Web of Data Structures
+
+In a sense, IPFS is a "web of data-structures", with the merkledag as the common denominator. Agreeing upon a format allows linking different authenticated datastructures to each other, enabling sophisticated distributed applications to easily construct, distribute, and link their data.
+
+### Linked Data
+
+The merkledag is a type of Linked-Data. The links do not follow the standard URI format, and instead opt for a more general and flexible UNIX filesystem path format, but the power is all there. One can trivially map formats like JSON-LD directly onto IPFS (IPFS-LD), making IPFS applications capable of using the full-power of the semantic web.
```
wking: Typing via entries in Links seems like an additional layer on top of the Merkle-DAG base.  If we don't have type links as first-class citizens (see discussion in ipfs/ipfs#36), I think we should punt this section to a higher-level doc, so it gets the same treatment as other useful things you can build using the Merkle-DAG base.

merkledag/README.md
wking: Never mind, the images you show below explain the “waist” analogy.

records/README.md
wking: “serializing” does seem like the right word, so maybe just swap “formating” and “serializing”?  Or:

    - `serializing` - packing a record into a bitstring.

?

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+### Record System
+
+A (distributed) `record system` is a protocol which defines a method for crafting, formating, distributing, verifying, and using records over computer networks. (e.g. the Domain Name System).
+
+
+- `crafting` - construction of a record (the process of calculating the values of a record)
+- `formating` - serializing of a record into a bitstring.
+- `distributing` - transportation of a record from one set of computers to another.
+- `verifying` - checking a record's values to ensure correctness and validity.
```
wking: I find it odd to explicitly list a serialization step without also listing an deserialization/parsing step.  Or is the need for that step so obvious that it's not worth listing?

merkledag/README.md
whyrusleeping: in a dag, multiple nodes can have the same child. Thats not usually part of the definition of a tree.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+
+### Validity Schemes
+
+A `validity scheme` is a small sub-protocol that defines how a record's _validity_ is to be calculated. `Validity` is the quality of a record being usable by a user at a particular set of circumstances (e.g. a range of time). It is distinct from `correctness` in that `correctness` governs whether the record was correctly constructed, and `validity` governs whether a record may still be used. All _valid_ records must be _correct_. For simplicity, the process of checking correctness is included in the `validity scheme`.
```
wking: Maybe use the words “syntax” and “semantics” or “policy” here to help draw the line between “correctness” and “validity”.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+For example, suppose Alice and Bob want to store records on a public bulletin board. To make sure their records are not tampered with, Alice and Bob decide they will include cryptographic signatures. This can ensure correctness. Further, they also agree to add new records every day, to detect whether their records are being replayed or censored. Thus, their `validity scheme` might be:
```
wking: I'm having trouble connecting signatures with “correctness”.  I'd expect “correctness” means “can I deserialize this record and find any fields required by this spec”.  If you make choices on top of this spec (e.g. you decide to sign your records), then the presence of a signature, the validity of the signature, the status of the signing key in your trust store, etc. are all part of “validity”.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+```go
+Record Node {
+  Scheme   Link // link to a validity scheme
+  Value    Link // link to an object representing the value.
+  Validity Data // data needed to satisfy the validity scheme
+}
+```
+
+To achieve good performance, record storage and transfer should bundle all the necessary objects and transmit them together. While "the record object" is only one of the dag objects, "the full record" means a bundle of all objects needed to fully represent, verify, and use the record. (This recommendation does not necessarily include data that records _describe_, for example an ipfs provider record (which signals to consumers that certain data is available) would not include the data itself as part of "the full record").
```
wking: Maybe drop “and use” from the “full record” definition, since that's so wishy.  To get past verification you'll need the Record, linked Scheme, and any further objects needed to validate the Validity data (which you may also want to store as links?  E.g. to public keys, etc?).  The only thing the “full record” does not need to bundle is stuff down the Value link.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+### The Interface
+
+The IPRS interface is below. It has a few types and functions. We use the Go language to express it, but this is language agnostic.
+
+```go
+// Record is the base type. user can define other types that
+// extend Record.
+type Record struct {
```
wking: I'm not sure how the `Record` type matches up with the `Record` node defined above.  For example, they seem to swap the link/data-ness of the value, and shift the generic `Validity` to a specific `Signature`.  Personally, I think it makes more sense to require `Value` to be a link to an IPFS object, although for small values the serving node may decide to bundle the value object into the “full record” for efficiency.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+  return recordProtobuf.Unmarshal(r, d)
+}
+```
+
+## Example Record Types
+
+For ease of use, IPRS implementations should include a set of common record types:
+
+- signed, valid within a datetime range
+- signed, expiring after a Time-To-Live
```
wking: Is this distinct from “valid within a datetime range”?  It seems like it's the same structure with an empty start date.

wking: Ah, I see there are notes about the distinction in the per-type sections below.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func Unmarshal(r *Record, d []byte) (error) {
+  return recordProtobuf.Unmarshal(r, d)
+}
+```
+
+## Example Record Types
+
+For ease of use, IPRS implementations should include a set of common record types:
```
wking: If all of these records require a PKI, it seems sensible to at least suggest a common infrastructure, and point out that participating nodes are free to refuse to host or transmit records that don't have valid signatures in that infrastructure.  This might be a good place to spec out the keyspace limitations as well, since requiring IPRS keys to start with a multihash for the signing public key is how we handle namespacing keys to avoid attackers maliciously clobbering your valid records.  So having a “use PK-namespaced IPRS keys and records signed using PKI $X” limitation seems like a reasonable requirement to me, although it can be a separate layer on top of the unsigned base protocol if you like.

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+This record type uses digital signatures (and thus a PKI) and TTLs (and thus a TI). It establishes that a record is valid for a certain amount of time after a particular event. For example, an event may be "upon receipt" to specify that a record is valid for a given amount of time after a processor first receives it. This is equivalent to the way DNS sets expiries.
+
+### Signed, based on ancestry (chain)
+
+This record type uses digital signatures (and thus a PKI) and merkle-links to other, previous records. It establishes that the "most recent" record (merkle-ordered) is the most valid. This functions similar to a git commit chain.
+
+### Signed, with cryptographic freshness
+
+This record type uses digital signatures (and thus a PKI) and a cryptographic notion of freshness (and therefore a TI). It establishes that records are only valid if within some threshold of recent time. It is similar to a TTL.
```
wking: This is also similar to the “based on ancestry” approach.  If you're using signatures to exclude malicious authors, then all you need for validity is a way to order potential records.  That can be either lookups in an ancestry chain (potentially slow) or comparing timestamps (fast).  This also only requires that the publisher has a consistent view of time (or record index, or whatever), and does not require the publisher and consumer to share a consistent view of time.  You only need a publisher/consumer TI if you also want an expiration time for records, but I imagine lots of folks will want a record that is valid forever but override-able by subsequent publication (i.e. “please use this record forever, unless you have access to a newer record that I've also published”).

records/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+This record type uses digital signatures (and thus a PKI) and timestamps (and thus a TI). It establishes that a record is valid during a particular datetime range. 0 (beginning of time), and infinity (end of time) can express unbounded validity.
+
+### Signed, expiring after a Time-To-Live
+
+This record type uses digital signatures (and thus a PKI) and TTLs (and thus a TI). It establishes that a record is valid for a certain amount of time after a particular event. For example, an event may be "upon receipt" to specify that a record is valid for a given amount of time after a processor first receives it. This is equivalent to the way DNS sets expiries.
+
+### Signed, based on ancestry (chain)
+
+This record type uses digital signatures (and thus a PKI) and merkle-links to other, previous records. It establishes that the "most recent" record (merkle-ordered) is the most valid. This functions similar to a git commit chain.
```
wking: Verifying this with a “full record” type transmission is going to be difficult.  Do you expect transactions to ship the entire history (this seems slow and wasteful)?  Or is the request going to have special information (e.g. “give me any records for <key> that are descendents of <last-value>”, this seems to require too much local storage on the record-serving node).  This is probably going to be a case where we break the “should” recommendation for full record bundling, and just leave it up to the client to fetch any intervening records.

merkledag/README.md
wking: On Fri, May 29, 2015 at 10:12:36AM -0700, Jeromy Johnson wrote:
> in a dag, multiple nodes can have the same child. Thats not usually
> part of the definition of a tree.

Ah, thanks.

keychain/README.md
jbenet: > but linking out to whatever you consider canonical docs or useful review articles for each proof type would be nice.

Yeah this is super raw and early. almost removed this section.

> I also wonder where these proofs are used

- this part (the proof types) is very very WIP. just a thought.
- the idea is to have "proof objects" that we can point to, and have their type tell us how to process them.
- a proof of storage could potentially be used in a provider record, to prevent spam. and they would be used in other things, like filecoin.

merkledag/README.md
jbenet: It's useful to have definitions for a spec in the same doc cause definitions can afford to be as specific as needed for this doc. a glossary one would be more general.

agree these (or at least some) should go into the glossary

merkledag/README.md
jbenet: content addressed data structures can be used to make authenticated data structures. 

authenticated data structures are a crypto idea. it's a data structure which can be manipulated and verified by agents who do not trust each other.

jbenet: amiller has a nice definition here: http://www.cs.umd.edu/~mwh/papers/gpads.pdf

merkledag/README.md
jbenet: An independent entity.

merkledag/README.md
jbenet: not yet

merkledag/README.md
jbenet: it's a refernce to the IP "thin waist" or "hourglass" model

merkledag/README.md
wking: On Fri, May 29, 2015 at 11:34:14AM -0700, Juan Batiz-Benet wrote:
> > +- `ipfs object` - a node in the ipfs merkledag. It represents a
> >    singular entity.
> 
> An independent entity. 

So “indepenently addressable”?  Or fetchable?  Hashable?  Verifiable?
All of the above?

merkledag/README.md
```diff
@@ -0,0 +1,144 @@
...
+- bitcoin uses a merkledag to construct the blockchain, a shared append-only ledger with convergent distributed consensus.
+- Tahoe-LAFS uses a merkledag to construct a secure, distributed, capability filesystem based on the least-authority principle.
+
+(NOTE: please suggest other systems to reference here.)
+
+### Thin Waist of Data Structures
+
+At its core, IPFS provides the merkledag as a primitive (or "internet layer") to build sophisticated applications easily. It is a "thin-waist" for secure, distributed applications, which -- by agreeing to follow the common format -- can then run across any replication, routing, and transport protocols. To draw an analogy, this is like the "thin-waist" IP provided to connect hosts across medium-specific networks.
+
+![](mdag.waist.png)
```
wking: I continue to be confused by the naming for our naming systems ;).  After ipfs/go-ipfs#1208, I was under the impression that both DNS links and DHT links fell under the “IPNS” umbrella, but in this image there are parallel entries for DNS and IPNS.  After this PR, perhaps IPRS is the umbrella and IPNS is specific to links on IPFS's DHT?

merkledag/README.md
jbenet: i'm increasingly dissatisfied by the annoyances of protobuf's shortcomings. have to trick it into doing things, like self description or streams.

records/README.md
jbenet: sure sgtm

records/README.md
jbenet: these aren't steps, but "parts".  "formatting" or "serializing" already implies deserializing.

records/README.md
jbenet: i'm drawing a distinction between:
- `correct record` - this is a record crafted _correctly_ (truly by the author, etc), independent of circumstances.
- `valid record` - this is a correct record that is valid or invalid given user circumstances (like time)

records/README.md
jbenet: it doesn't exactly. i need a good notation to describe IPFS objects.

jbenet: oh and i wrote this before the keychain above.

records/README.md
jbenet: hahaha yep.

records/README.md
jbenet: >  it seems sensible to at least suggest a common infrastructure, and point out that participating nodes are free to refuse to host or transmit records that don't have valid signatures in that infrastructure.

agreed.

> ince requiring IPRS keys to start with a multihash for the signing public key is how we handle namespacing keys to avoid attackers maliciously clobbering your valid records

some records (like dht provider records) are distributed in the ipfs-dht under the key for the content they give a record for, instead of the author's key.

records/README.md
wking: On Fri, May 29, 2015 at 11:57:29AM -0700, Juan Batiz-Benet wrote:
> i'm drawing a distinction between:
> - `correct record` - this is a record crafted _correctly_ (truly by
>   the author, etc), independent of circumstances.
> - `valid record` - this is a correct record that is valid or invalid
>   given user circumstances (like time)

Ah, got it.  And I see you do mention construction and use in your
“validity scheme” paragraph, but maybe you could spell it out like
that ^ for folks like me who incorrectly jump to the syntax/semantics
distinction.

records/README.md
wking: On Fri, May 29, 2015 at 12:00:31PM -0700, Juan Batiz-Benet wrote:
> some records (like dht provider records) are distributed in the
> ipfs-dht under the key for the content they give a record for,
> instead of the author's key.

Oh, right.  I guess they'll need a separate section on how they avoid
collisions/forgery and aggregate provider records for a given
resource.

records/README.md
jbenet: >  Or is the request going to have special information (e.g. “give me any records for that are descendents of ”, this seems to require too much local storage on the record-serving node).

It may be possible to bundle with bitswap replication / query notation -- which will be something like git ref notation and xpath... -- (things like `want <path>/**/*` or `want <path>~10` OR go the full xpath route... oof.)

>  This is probably going to be a case where we break the “should” recommendation for full record bundling, and just leave it up to the client to fetch any intervening records.

Yeah, agreed. already there's the case that a user will likely have the keys used to sign records with, so dont need to send. 

i wonder if it makes sense to come up with a "bundle" abstraction in bitswap (or other replication protocol) that expresses "I'm about to send you this graph (all the keys)", and then starts sending. meanwhile, the receiver -- after getting the first message -- can quickly ack the keys it does have and prevent the send. the simpler (but maybe less well-performing) option is to just wait till the user requests things.

#### just some examples to get people thinking:






@krl this is another concerns for _requiring_ records to _always be_ based on ancestry. the big one though, is that records can be lost, so the chain may not be complete and have to deal with that.

wking: On Fri, May 29, 2015 at 12:17:08PM -0700, Juan Batiz-Benet wrote:
> > Or is the request going to have special information (e.g. “give me
> > any records for that are descendents of ”, this seems to require
> > too much local storage on the record-serving node).
> 
> It may be possible to bundle with bitswap replication / query
> notation -- which will be something like git ref notation and
> xpath... -- (things like `want <path>/**/*` or `want <path>~10` OR
> go the full xpath route... oof.)

That sounds nice, but my main concern is that in the absence of
senders knowing what the requester will ask for they won't know which
auxiliary records to cache locally.  But I'm not familiar with bitswap
enough to say for sure.  Maybe you earn points by helping the
requester track down any records you're missing, and/or you benefit by
caching repeatedly-requested nodes follow-up objects locally.

This isn't a problem wit Git, etc., because the sending node always
has all the records available locally.  It's probably worth looking
into whether and how Git supports serving from a shallow clone, since
that's very similar to what we have in an IPRS node.

jbenet: in general, for small records like these, _waiting_ is way worse than _sending it multiple times_.  because these are in the hot-path of resolving for roots, many extra RTTs are killer here.

wking: On Fri, May 29, 2015 at 12:50:33PM -0700, Juan Batiz-Benet wrote:
> > +### Signed, based on ancestry (chain)
> > +
> > +This record type uses digital signatures (and thus a PKI) and
> >  merkle-links to other, previous records. It establishes that the
> >  "most recent" record (merkle-ordered) is the most valid. This
> >  functions similar to a git commit chain.
> 
> in general, for small records like these, _waiting_ is way worse
> than _sending it multiple times_.  because these are in the hot-path
> of resolving for roots, many extra RTTs are killer here.

I get that, but for something where validity could potentitially
require the entire commit history behind the tip record, defining “the
hot-path” sounds like it's going to require the sender to have:

* a periodically-updated memory of what records clients are requesting
  (e.g. “I'm often asked for <object> after requests for <key>, so I
  could just send that optimistically”) and
* some incentive to supply those records directly instead of leaving
  the client to fetch them itself (e.g. “In fact, I'm asked so
  frequently for <object> that I might as well cache it locally to
  inexpensively gain bitswap points that I can cash-in to get other
  objects I want”).

merkledag/README.md
whyrusleeping: we should like, mirror that page in ipfs and link to the hash :smile:

records/README.md
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Validator is a function that returns whether a record is valid.
+// Users provide their own Validator implementations.
+type Validator func(r *Record) (bool, error)
+
+// Order is a function that sorts two records based on validity.
+// This means that one record should be preferred over the other.
+// there must be a total order. if return is 0, then a == b.
+// Return value is -1, 0, 1.
```
wking: Why restrict the return values?  I'd recommend consumers just use `> 0` instead of `== 1` when acting on this information.

records/README.md
```diff
@@ -0,0 +1,244 @@
...
+type Order func(a, b *Record) int
```
wking: For non-trivial ordering (e.g. walking the commit history), how will we handle timeouts and/or errors?  Should this take an additional `Context` argument and return an additional error?  This consideration also applies to the `Validator` function.

records/README.md
```diff
@@ -0,0 +1,244 @@
...
+// extend Record.
+type Record struct {
+  Scheme    Link // link to the validity scheme
+  Signature Link // link to a cryptographic signature over the rest of record
+  Value     Data // an opaque value
+}
+
+// Validator is a function that returns whether a record is valid.
+// Users provide their own Validator implementations.
+type Validator func(r *Record) (bool, error)
```
wking: Function name should probably be a verb (e.g. `Validate`), although see also [reasons to noun functions][1].

[1]: http://www.grinchcentral.com/function-names-to-verb-or-not-to-verb

jbenet:
this should help describe what i mean in the keychain stuff:

jbenet:
(yay diagrams. if anyone knows a _good_ web based programmatic + drag and drop diagram tool, lmk. (most of them suck)

records/README.md
jbenet: yeah i mean this as a general protocol interface, not a go-ipfs specific interface. (i just find golang interface notation very nice to write)

wking: On Fri, May 29, 2015 at 04:00:47PM -0700, Juan Batiz-Benet wrote:
> > +type Order func(a, b *Record) int
> 
> yeah i mean this as a general protocol interface, not a go-ipfs
> specific interface. (i just find golang interface notation very nice
> to write)

Hmm.  So are things like error handling too low level for the general
protocol API?  In that case, maybe drop the error return from the
`Validator` and (un)marshaling functions?  Or maybe define the general
protocol interface as language-agnostic REST endpoints?

jbenet: Oh error handling makes sense. i meant for the `context` part

records/README.md
krl: > @krl this is another concerns for requiring records to always be based on ancestry. the big one though, is that records can be lost, so the chain may not be complete and have to deal with that.

If records can be lost, it's not really a chain structure. Compare this with blockchains, all the security properties would fall apart if blocks could be lost.

In what case would this be more reasonable in ipns than in bitcoin?

jbenet: @krl 
- records _can always be lost_. even in bitcoin. bitcoin tries to ensure records are not lost by attaching a gigantic (billions of USD) monetary cost to losing records. ipns does not.
- ipns is _not always a chain_, it doesn't claim to be.
- ipns must be able to support high throughput mutability -- ie it should not _impose_ any problem on high throughput use cases. high throughput, to give an example, means 1 million records/mutations per second. In a case like this, people _will not_ store all the records, willingly breaking any imposed chains. it is better not to impose a chain in this case.
- If we're going to call anything a record chain, it should be in a context where it is _strictly_ encouraged to be -- and remain -- a full chain, so that guarantees can be preserved. Not all uses of ipns require this.

krl: > If we're going to call anything a record chain, it should be in a context where it is strictly encouraged to be -- and remain -- a full chain, so that guarantees can be preserved. Not all uses of ipns require this.
+9000

merkledag/README.md
jbenet: IPRS is not the umbrella. IPRS is just a way to create records. it doesnt say anything about IPNS or making a name system. it merely specifies ways of writing + verifying records.

There's:
- IPNS - the name system for IPFS, which can resolve names on the ipfs routing system, dns names, and more.
- IPNS identity addresses (public key names mapped to records, stored on an IPFS routing system, as specified by this IPRS spec).

yes, DNS perhaps in the future should be separated out. this is a ton of work right now and not useful because it only confuses end users when they have to worry about multiple roots and so on. today, it's fine to use `"/ipns/..."` for `"/ipns/<hash-of-pubkey>/..."` or `"/ipns/<dns-domain-name/..."`. and really helps keep the number of logical symbols people have to worry about when they first encounter IPFS to a small number. 

note also that the word "IPFS" also means many things. it's not a big deal in practice.

merkledag/README.md
jbenet: this part doesnt say anything about "Typing via entries in Links". this part talks about JSON-LD, which the base merkledag format supports out of the box, because JSON-LD just adds one `@context` entry mapped to a spec written in JSON. an equivalent `IPFS-LD` would work similarly.

jbenet: anyway, this give readers a flavor of what is achievable.

records/README.md
jbenet: yeah, you're right. in general i find "validity scheme" weak and confusing. i'll try to bring these bullets in.

records/README.md
jbenet: ok sounds good

merkledag/README.md
wking: On Tue, Jun 02, 2015 at 10:35:50PM -0700, Juan Batiz-Benet wrote:
> this part doesnt say anything about "Typing via entries in
> Links". this part talks about JSON-LD, which the base merkledag
> format supports out of the box, because JSON-LD just adds one
> `@context` entry mapped to a spec written in JSON. an equivalent
> `IPFS-LD` would work similarly.

See the “collisions with filenames” discussion referenced from [1] and
continued in subsequent comments.  I don't think we need to rehash
that discussion here, but I do think that the IPFS-LD idea needs more
polishing than it has received so far before we start talking about it
in a Merkle DAG spec that otherwise restricts itself to well-tested
material.

[1]: https://github.com/ipfs/ipfs/issues/36#issuecomment-98037120

whyrusleeping:
jbenet for the records, youre thinking this right: https://gist.github.com/whyrusleeping/8f2c206ac2fbc952fea2

records/README.md
whyrusleeping: @wking this isnt a function name, its a type

whyrusleeping:
@jbenet re: a drawing tool, i use http://draw.io its pretty nice

merkledag/README.md
greglook: @jbenet are you familiar with [EDN](https://github.com/edn-format/edn)? It handles self-description pretty elegantly, and has a few binary serialization formats.

jbenet:
@whyrusleeping: not quite https://gist.github.com/whyrusleeping/8f2c206ac2fbc952fea2#comment-1466430

merkledag/README.md
jbenet: @greglook not yet! but it's made by @richhickey so it's likely right. thanks for the pointer!

jbenet: @greglook ah no, this is a text format. it's not optimized for binary rep. (unless i'm missing a page)

greglook: @jbenet see [Datomic/fressian](https://github.com/Datomic/fressian) and [cognitect/transit-format](https://github.com/cognitect/transit-format) for binary representations. The latter is not strictly EDN, but includes the same core ideas.

jbenet: great, thanks @greglook!

records/README.md
wking: On IRC just @whyrusleeping asked about breaking ties after changing
validity schemes [1].  In the dicussion through [2], I advocated for
including revocation lists in records published after an incompatible
validity change until the revoked records expired.  @jbenet worried
about handling folks who publish valid-forever records before
switching validity schemes, but I think that's just a good reason to
not use valid-forever records by default.  If users want to take that
risk and configure their publisher to create valid-forever records,
then it's on them to either:

a. Not change to an incompatible validity scheme after publishing a
   valid-forever record, or
b. Carry the valid-forever-in-the-old-validity-scheme records around
   in their revocation list for all subsequent records.

[1]: https://botbot.me/freenode/ipfs/2015-06-03/?msg=40942042&page=7
[2]: https://botbot.me/freenode/ipfs/2015-06-04/?msg=40943523&page=1

records/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+The IPRS interface is below. It has a few types and functions. We use the Go language to express it, but this is language agnostic.
+
+```go
+// Record is the base type. user can define other types that
+// extend Record.
+type Record struct {
+  Scheme    Link // link to the validity scheme
+  Signature Link // link to a cryptographic signature over the rest of record
+  Value     Data // an opaque value
```
whyrusleeping: Here value is data, and above it is a link. Above we have data for the validity information, but here there is no place for that to go. Something is wonky here. If i have my way, it would look like:

```json
{
    "Data": "validity data, to be interpreted after parsing the scheme link",
    "Links":[
        {
            "Name":"Scheme",
            "Hash": "pointer to schema definition object (really just a placeholder)",
        },
        {
            "Name":"Signature",
            "Hash": "pointer to cryptographic signature",
        },
        {
            "Name":"Value",
            "Hash": "pointer to records actual content",
        }
    ]
}
```

whyrusleeping: and below, where we have the date time thing, that info would just be another link.

whyrusleeping: Another thing, is the signature separate from the validity data? I'm okay with that, just want to make the distinction clear.

wking: On Wed, Jun 03, 2015 at 06:46:24PM -0700, Jeromy Johnson wrote:
> > +type Record struct {
> > +  Scheme    Link // link to the validity scheme
> > +  Signature Link // link to a cryptographic signature over the rest of record
> > +  Value     Data // an opaque value
> 
> Another thing, is the signature separate from the validity data? I'm
> okay with that, just want to make the distinction clear.

Just to keep the vocabulary consistent, @jbenet was putting the
signature under “correctness” not “validity” (since it's a
done-right-at-craft-time issue [1]), although they both fall under the
“validity scheme” [2].

I agree with [3] that it makes sense to allow a given validity scheme
to place correctness/validity information in additional links and/or
the data block as it sees fit, so long as it doesn't clobber the
base-Record "Scheme" or "Value" names (and do those have to be
title-cased?).

As it's currently written up, the linking for a signature is going to
be weird, since what you really want signed is *the record itself*.
The whole “signable part” business reminds me of OpenPGP with it's
“unhashed subpacket data” [4].  I played around with an OpenPGP
implementation while trying to rotate my main GnuPG key [5], and the
whole “unhashed data” business is a pain in the neck.  Can we just
make signature objects first-class citizens and allow (require?) folks
to push signatures instead of records.  Then the signature could link
to the record, which would carry the validity information and a link
to the record payload.

[1]: https://github.com/ipfs/specs/pull/7/files#r31354268
[2]: https://github.com/ipfs/specs/pull/7/files#diff-0cb9903b60857948374119aa5ca7a097R37
[3]: https://github.com/ipfs/specs/pull/7/files#r31687180
[4]: https://tools.ietf.org/html/rfc4880#section-5.2.3
[5]: http://git.tremily.us/?p=gpg-migrate.git

whyrusleeping: question 3, is there actually going to be anything in the 'schema object' ? will it even be an object? what should it contain?

records/README.md
wking: On Fri, May 29, 2015 at 12:08:36PM -0700, W. Trevor King wrote:
> Fri, May 29, 2015 at 12:00:31PM -0700, Juan Batiz-Benet:
> > some records (like dht provider records) are distributed in the
> > ipfs-dht under the key for the content they give a record for,
> > instead of the author's key.
> 
> Oh, right.  I guess they'll need a separate section on how they
> avoid collisions/forgery and aggregate provider records for a given
> resource.

So I'm not sure what the current syntax here is, but I'd expect the
provider-list record would be keyed with something like:

  <multihash-of-provided-object>/providers

the payload itself wouldn't be a signature, but would Merkle-link to
signatures where the link names were the provider IDs.  Each linked
signature would have a signed payload with links:

  Provider: <multihash-of-provider>
  Object: <multihash-of-provided-object>

So the Merkle chain of the record would be:

  record
    <provider-id>: <signature-id>
  ↓               
  signature
    Key: <provider-id>
    Signee: <providing-claim-id>
    …
  ↓
  <providing-claim-id>
    Provider: <provider-id>
    Object: <multihash-of-provided-object>

Nodes that had reason to trust each other (e.g. to not forge providing
claims or to properly currate a provider-list) wouldn't have to fetch
and verify the signed data.  Nodes that had no reason to trust each
other (e.g. fetching a provider list from an untrusted node or
receiving a new providing-claim from an untrusted node) should aquire
and check the signature before using the providing-claim for lookup or
adding it to a provider-list.  Since most nodes won't trust each
other, I expect the signature, providing-claim, public-key, etc.,
packets would be passed around in one optimistic-transmit block [1],
and probably be optimistically hosted on the same nodes that host the
provider-list itself for that purpose.

[1]: https://github.com/ipfs/specs/pull/7#discussion_r31356057

jbenet:
I'm going to merge this and continue improving.

PR #9, opened 2015-06-03
added wire protocol description

jbenet:
this is also WIP but should be good enough to kick off node-ipfs and make the necessary changes to go-ipfs.

we do need to get the spdy situation figured out in go-ipfs or switch to something else. cc @whyrusleeping

### DIFF
--- a/protocol/network/wire.md
+++ b/protocol/network/wire.md
@@ -1,0 +1,105 @@
+
+# IPFS Protocol on the Wire
+
+Authors: [Juan Benet](http://github.com/jbenet)
+
+Reviewers:
+
+
+* * *
+
+This [spec](../../) describes the IPFS protocol on the wire.
+
+As explained in the [network spec](./network), IPFS is [transport agnostic](./network/#Transport-Agnostic). We have the **hard constraint** of making IPFS work across _any_ duplex stream (an outgoing and an incoming stream pair, any arbitrary connection) and work on _any_ platform.
+
+To make this work, IPFS has to solve a few problems:
+
+- [Protocol Multiplexing](#Protocol-Multiplexing) - running multiple protocols over the same stream
+  - [multistream](#multistream) - self-describing protocol streams
+  - [multistream-select](#multistream-select) - a self-describing protocol selector
+  - [Stream Multiplexing](#Stream-Multiplexing) - running many independent streams over the same wire.
+- [Portable Encodings](#Portable-Encodings) - using portable serialization formats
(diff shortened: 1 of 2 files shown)


jbenet:
cc @diasdavid @travisperson @krl @mappum

protocol/network/wire.md
```diff
@@ -0,0 +1,105 @@
...
+- SPDY
+- QUIC
+- SSH
+
+IPFS nodes are free to support whatever stream multiplexors they wish, on top of the default one. The default one is there to enable even the simplest of nodes to speak multiple protocols at once. The default multiplexor will be HTTP/2 (or maybe QUIC?), but implementations for it are sparse, so we are beginning with SPDY. We simply select which protocol to use with a multistream header.
+
+For example:
+
+```
+/ipfs/QmdRKVhvzyATs3L6dosSb6w8hKuqfZK2SyPVqcYJ5VLYa2/multistream-select/0.3.0
```
diasdavid: Supporting several multistream-select versions is part of the desired feature set also?

protocol/network/wire.md
```diff
@@ -0,0 +1,105 @@
...
+/ipfs/Qmb4d8ZLuqnnVptqTxwqt3aFqgPYruAbfeksvRV1Ds8Gri/spdy/3
+<spdy-header-opening-a-stream-0>
+/ipfs/QmVXZiejj3sXEmxuQxF2RjmFbEiE9w7T82xDn3uYNuhbFb/ipfs-dht/0.2.3
```
diasdavid: Where do these hashes come from?

jbenet: i should probably make real examples that are resolvable. the idea is that you'd have the specs in directories that give the protocol name + version number, and then we'd publish them with ipfs, getting a resulting root hash. this allows the hashes to be resolved for and look at the spec. but yeah, _these_ particular hashes aren't resolvable because i made them up. perhaps they should be for more clear examples. (may also be relevant to link directly to commit objects once we have those.)

protocol/network/wire.md
jbenet: not necessarily, but the version is there to make it clear what version it is in a human readable sense? we don't actually need it-- the hash is enough.

protocol/network/wire.md
travisperson: That is *so* cool.

jbenet:
I'm going to merge this for now.

PR #13, opened 2015-06-23
protocol/spec.tex: First pass at an module overview diagram

wking:
Details in the commit messages.


jbenet:
@wking thanks for making these. I was looking for something more like these:

jbenet:
made others. closing this

PR #14, opened 2015-06-26
DHT Spec

diasdavid:
This is a `Work in Progress`

@jbenet I'm doing a PR now so that you can fill in some of the questions that appeared when I started writing it. They are highlighted with `_italic_` and a `?`. Thanks


protocol/routing/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+Authors: David Dias
+
+Reviewers:
+
+TODOS:
+
+-----------------------
+
+> This spec defines the routing protocol spec, covering `Peer discovery`, `Routing` and the `DHT`. The spec is a **Work In Progress**.
```
jbenet: We should separate out the DHT from Routing. Routing is an interface, that is satisfied by DHT.

protocol/routing/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+-_How offen do we issue this searches?_
+
+### Routing
+
+For impl reference, check: https://github.com/ipfs/go-ipfs/blob/master/routing/routing.go#L19-L49
+
+#### Find a peer
+
+_When searching for a peer, do we fetch the kbucket from a peer and see which peer we want to ping next or do we ask for a given Id to a peer and that peer replies to us with the best candidate (or itself if it is the case)?_
```
jbenet: > we ask for a given Id to a peer and that peer replies to us with the best candidate (or itself if it is the case)

It returns a set (3?) of best candidates I think.

protocol/routing/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+#### Ping
+
+Ping mechanism (for heartbeats). Ping a peer and log the time it took to answer.
+
+_what if the Id doesn't exist? Is there any rule for non existing peers? Should we log time for best matches as well?_
```
jbenet: > what if the Id doesn't exist?

if we exhaust the query set and haven't found the peer in question, we exit with an error.

diasdavid:
Superseded by https://github.com/ipfs/specs/pull/18

PR #18, opened 2015-07-18
Protocol Spec Update

diasdavid:
This PR contains my updates to the Network Spec and a new spec for the DHT (the one I started writing to have thoughts down before we had an 'official' one).

Right now, what is considered most 'finished' is the Network spec, my next step is finish the routing layer and polish the spec here too.

Nevertheless, here goes the PR for peer-review :)

PS: I also started adding chapter numbers, I was missing those from seeing other RFC, papers, etc, it helps skimming through the different sections :)


protocol/README.md
```diff
@@ -1,22 +1,31 @@
-# IPFS Protocol Spec (WIP!)
+IPFS Protocol Spec
+==================
 
-Authors: [@jbenet](http://github.com/jbenet)
+> **This spec is an Work In Progress (WIP)**
+
+Authors: 
+
+- [@jbenet](http://github.com/jbenet)
```
jbenet: maybe add yourself ?

diasdavid:
Follow https://github.com/ipfs/specs/pull/19

PR #19, opened 2015-08-23
libp2p spec

diasdavid:
It is getting form :) still a working progress, but leaving a PR here so people can start watching it, make comments, reviews, etc :)


jbenet:
Great progress!! :+1: 

(i can spend some time working on this too later this week.)

diasdavid:
1st
img_1478

2nd (After I realized that the current discovery only belongs to kad-routing and that peer-routing actually makes more sense to be called discovery, since we 'discover' peers for a given key)
img_1479

each of these components exposes the given interface and is free to use any other of the modules available through that interface

jbenet:
> current discovery only belongs to kad-routing

we have three "discovery" protocols at the moment:
- random walks on the dht 


> discovery

can we keep this as "peer routing"? "peer discovery" is already a thing in other protocols and it typically means "finding more peers", not "finding a specific peer in the network". in dht-land, the coral + kademlia papers describe the query as a routing process. and call it a "routing table".

jbenet:
(that said, "content routing" could be called "content discovery", but i think given the "peer routing" naming, it should be consistent. also "content discovery" might be a set of passive protocols in the future, like gossip distribution of content)

diasdavid:
Totally see your point, we should follow and keep calling it Peer Routing then :)

jbenet:
(sorry, it's really hard to find semantic alignment/fit in these things in general, and it took us many months to get to these)

diasdavid:
_incoming_

img_1480

diasdavid:
Next steps on libp2p definiton. Make libp2p-interfaces.go and use the Go lang as a way to define the interfaces.

protocol/network/3-requirements.md
```diff
@@ -0,0 +1,104 @@
...
+We take both security and performance seriously. We recognize that encryption is not viable for some in-datacenter high performance use cases.
+
+We recommend that:
+- implementations encrypt all communications by default
+- implementations are audited
+- unless absolutely necessary, users normally operate with encrypted communications only.
+
+libp2p uses cyphersuites like TLS.
+
+**NOTE:** we do not use lib2p directly, because we do not want the CA system baggage. Most libp2p implementations are very big. Since the lib2p model begins with keys, libp2p only needs to apply ciphers. This is a minimal portion of the whole TLS standard.
```
diasdavid: This doesn't hold true today and must be updated. Discussion - https://github.com/ipfs/specs/issues/29

diasdavid:
I've been receiving lot of feedback that there isn't a specification anywhere for libp2p and although that we don't have a finished specification, we do have something informative to say the least. I've learned that coming and reading from this PR doesn't come natural so what I want to do is merge this PR, adopt the organization of the document and evolve it from there, it has greatly changed from the one in Master, specially after I started implementing libp2p. Sounds good @jbenet ? :)

RichardLitt:
>  I've learned that coming and reading from this PR doesn't come natural so what I want to do is merge this PR, adopt the organization of the document and evolve it from there, it has greatly changed from the one in Master, specially after I started implementing libp2p.

100% agreed. If this is the source of truth, merge it. Reading a PR is confusing. 

> Also, it would be cool to extrapolate a whitepaper/paper for libp2p

Also yes.

protocol/network/1-introduction.md
```diff
@@ -0,0 +1,27 @@
+1 Introduction
+==============
+
+With the developement of building IPFS, the InterPlanetary FileSystem[?], we came to learn about the several challenges imposed by having to run a distributed file system on top of heterogeneous devices, with diferent network setups and capabilities. During this process, we had to revisit the whole network stack and elaborate solutions to overcome the obstacles imposed by design decisions of the several layers and protocols, without breaking compatibility or recreating technologies.
+
+In order to build this library, we focused on tackling problems independently, creating less complex solutions with powerful abstractions, that when composed, can offer an environment for a Peer-to-Peer application to work sucessfuly.
+
+## 1.1 Motivation
+
+`libp2p` is the result of the collective experience while building a distributed system, that puts the responsability on the developers on how they want their app to interop with others in the network, favoring configuration and extensibility instead of assumptions about how the network setup.
```
jbenet: "about the network setup"

protocol/network/1-introduction.md
```diff
@@ -0,0 +1,27 @@
...
+
+In essence, a peer using libp2p should be able to communicate with another peer using different transports, including connection relay, and talk over different protocols, negotiated in demand.
```
jbenet: "on demand"

protocol/network/1-introduction.md
```diff
@@ -0,0 +1,27 @@
...
+
+## 1.2 Goals
+
+Our goals for libp2p specification and its implementations are:
+
+- Enable the use of various transports (TCP, UDP, DTLS, uTP, etc)
```
jbenet: distinguish "transports" from "authenticated transports" into two bullets.
> - transports: TCP, UDP, SCTP, UDT, uTP, QUIC, SSH, etc.
> - authenticated transports: TLS, DTLS, CurveCP, SSH

protocol/network/1-introduction.md
```diff
@@ -0,0 +1,27 @@
...
+- Efficient use of sockets (connection reuse)
```
jbenet: also:
> - efficient use of underlying transport (e.g. native stream muxing, native auth, etc)

protocol/network/2-state-of-the-art.md
```diff
@@ -0,0 +1,70 @@
+2 An analysis the State of the Art in Network Stacks
+====================================================
+
+This section presents to the reader an analysis of the available protocols and arquitectures for a Network Stack. The goal is to provide the foundations from which to infer the conclusions and understand what are libp2p requirements and its designed arquitecture.
```
jbenet: "architectures"

protocol/network/2-state-of-the-art.md
```diff
@@ -0,0 +1,70 @@
...
+- Protocols started to be designed with the assumption that a developer will create a client-server application from the start.
+
+We even learned how to hide all of the complexity of a distributed system behind gateways on the Internet, using protocols that were designed to perform a point-to-point operation, such as HTTP, making it opaque for the application to see and understand how the cascade of service calls made for each request.
+
+`libp2p` offers a move towards dialer-listener interactions, from the client-server listener, where it is not implicit which of the entities, dialer or listener, has which capabilities or is enabled to perform which actions. Setting up a connection between two applications today is a multilayered problem to solve, and these connections should not have a purpose bias, instead support to several other protocols to work on top of the established connection. In a client-server model, a server sending data without a prior request from the client is known as a push model, which typically adds more complexity, in a dialer-listener model, both entities can perform requests independently.
```
jbenet: :+1:  very nicely said

protocol/network/2-state-of-the-art.md
```diff
@@ -0,0 +1,70 @@
...
+
+## 2.2 Categorizing the network stack protocols by solutions
+
+### 2.2.1 Establishing the phisical Link
```
jbenet: physical

protocol/network/2-state-of-the-art.md
```diff
@@ -0,0 +1,70 @@
...
+- ethernet
+- wifi
+- bluetooth
+- usb
+
+### 2.2.2 Addressing a machine or process
+
+- IPv4
+- IPv6
+- Hidden Addressing, like SDP
```
jbenet: - Oblivious routing systems, like Tor and I2P

protocol/network/2-state-of-the-art.md
```diff
@@ -0,0 +1,70 @@
...
+- Protocols started to be designed with the assumption that a developer will create a client-server application from the start.
+
+We even learned how to hide all of the complexity of a distributed system behind gateways on the Internet, using protocols that were designed to perform a point-to-point operation, such as HTTP, making it opaque for the application to see and understand how the cascade of service calls made for each request.
+
+`libp2p` offers a move towards dialer-listener interactions, from the client-server listener, where it is not implicit which of the entities, dialer or listener, has which capabilities or is enabled to perform which actions. Setting up a connection between two applications today is a multilayered problem to solve, and these connections should not have a purpose bias, instead support to several other protocols to work on top of the established connection. In a client-server model, a server sending data without a prior request from the client is known as a push model, which typically adds more complexity, in a dialer-listener model, both entities can perform requests independently.
+
+## 2.2 Categorizing the network stack protocols by solutions
```
jbenet: maybe a note describing why this is here:

> Before diving into the libp2p protocols, it is important to understand the large diversity of protocols already in wide use and deployment that help maintain today's simple abstractions. For example, when one thinks about an HTTP connection, one might naively just think `HTTP/TCP/IP` as the main protocols involved, but in reality many more participate, all depending on the usage, the networks involved, and so on. Protocols like `DNS, DHCP, ARP, OSPF, Ethernet, 802.11 (WiFI), ...` and many others get involved. Looking inside ISPs' own networks would reveal dozens more.
>
> Additionally, it's worth noting that the traditional 7-layer OSI model characterization does not fit libp2p, because the OSI model tends to drive towards point-to-point links between applications, whereas the libp2p protocols speak more towards various sizes of networks with various properties cooperating under various  different security models. Also, libp2p protocols can "peer at the same level", meaning that multiple protocols can run simultaneously to handle one type of responsibility, instead of one-per-layer in traditional OSI stacking (for example, `bootstrap lists`, `mDNS`, `DHT Discovery`, and `PEX` are all forms of "_Peer Discovery_" that can coexist and can synergize.

jbenet: updated the msg above o/

jbenet:
hang on, i'm reviewing it properly now. i'll post back here when done

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+==============
+
+libp2p was designed around the Unix Philosophy, creating smaller components, easier to understand and to test. These components should also be able to be swapped in order to accomodate different technologies or scenarios and also make it that it is upgradable over time.
+
+Although different Peers can support different protocols depending on their capabilities, any Peer can act as a dialer and/or a listener for connections from other Peers, connections that once established can be reused from both ends, removing the distinction between clients and servers.
+
+libp2p interface acts as a thin veneer to three subsystems that are required in order for peers to be able to communicate. These subsystems are allowed to be built on top of other subsystems as long as they respect the standardized interface. The main 3 subsystems are:
+
+- Peer Routing - Mechanism to find a Peer in a network. This Routing can be done recursively, iteratively or even in a broadcast/multicast mode.
+- Swarm - Handles everything that touches the 'opening a stream' part of libp2p, from protocol muxing, stream muxing, NAT Traversal, Connection Relaying, while being multitransport
```
jbenet: @diasdavid i think you merged "NAT Traversal" "Connection Relaying" and "Protocol Muxing" into your node swarm, but i do not think think this should be merged in the libp2p spec, nor on Go. The go swarm only handles transport multiplexing (connections/streams) to many peers. NAT and Relay are separate protocols. In Go, the "Host" handles the protocol multiplexing, and composes (uses) the NAT and Relay protocols.

So, this should not be "tree main subsystems", but rather "multiple subsystems" brought together under the `libp2p.Node` (I think "Node" is a better word than "Host" or "Peer").

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+- Distributed Record Store - A system to store provider records and relay records
```
jbenet: maybe just
- Distributed Record Store - A system to store and distribute records. Records are small entries used by other systems for signaling, establishing links, announcing peers or content, and so on. It has a similar role to DNS in the broader internet.

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+Each of these subsystem exposes a well known interface (see chapter 6 for Interfaces) and may use eachother in order to fulfil their goal. A global overview of the system is:
+
+```
+┌─────────────────────────────────────────────────────────────────────────────────┐
+│                                  libp2p                                         │
+└─────────────────────────────────────────────────────────────────────────────────┘
+┌─────────────────────────┐┌──────────────────────────┐┌──────────────────────────┐
+│       Peer Routing      ││      Swarm               ││ Distributed Record Store │
+└─────────────────────────┘└──────────────────────────┘└──────────────────────────┘
+```
```
jbenet: the "not three systems" above changes this picture to have other things (namely, NAT, and Relay). there are likely to be other protocols "that peer" here too. and the reason "NAT" and "Relay" can't be part of "the thing named Swarm here" is that "NAT" and "Relay" make use of "Peer Routing" and "Peer Routing" makes use of "NAT" and "Relay". They're really peering-protocols.

may be useful to define

> - Protocol Peering -- when two protocols make use of each other, and don't explicitly layer one on top of the other.

somewhere

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+## 4.1 Peer Routing
+
+A Peer Routing subsystem, exposes an interface to identify which peers should a message be routed in the DHT. It receives a key and must return one or more PeerInfo objects.
+
+We present two examples of possible Peer Routing subsystems, the first based on a the Kademlia DHT and the second based on mDNS. Nevertheless, other Peer Routing mechanisms might be implemented, as long as they fulfil the same expectation and interface.
```
jbenet: s/might/can be/

(might sounds like a big maybe, instead of an easy to do thing)

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+### 4.1.1 kad-routing
+
+kad-routing implements the Kademlia Routing table, where each peer holds a set of k-buckets, each of them containing several PeerInfo from other peers in the network. In order to find the whereabouts of these peers, it implements 3 discovery mechanisms:
+
+- mDNS-discovery
+- random-walk
+- bootstrap-list
```
jbenet: - bootstrap-list is not part of kad-routing
- mDNS-discovery is not part of kad-routing

these are just discovery protocols...

protocol/network/4-architecture.md
jbenet: missing also from this is the "Discovery Protocols" which also peer

jbenet: and would make the big "libp2p" thing at the top be a "libp2p.Node"

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+││             └──────────────┘└──────────────┘└──────────────┘ ││
+│└──────────────────────────────────────────────────────────────┘│
+│┌──────────────────────────────────────────────────────────────┐│
+││ mDNS-routing                                                 ││
+│└──────────────────────────────────────────────────────────────┘│
+│┌──────────────────────────────────────────────────────────────┐│
+││ other-routing-mechanisms                                     ││
+│└──────────────────────────────────────────────────────────────┘│
+└────────────────────────────────────────────────────────────────┘
+```
```
jbenet: this diagram is misleading, vertical placement tends to imply "things rely on the things below them" where here i think you want to show how protocols "peer" with each other. useful to represent them horizontally in that case.

(i know horizontal is challenged, but most networks people who read a vertical stack of protocols take that to mean well-layered dependence relations, when here that isn't the case)

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### 4.1.1.1 mDNS-discovery
+
+mDNS-discovery uses mDNS to emit beacons in the local area network to find if there are more peers available, these peers are extremely interesting because our link to them will have an low latency, plus if we are disconnected from the network, we will be leverage the content on those peers have cached.
```
jbenet: - This whole thing should be under a heading of "Discovery Protocols"  
- This should also probably include how the DiscoveryProtocol interface is mapped to mDNS.
- maybe say:
> mDNS-discovery is a Discovery Protocol that uses mDNS (link to wikipedia) over local area networks. It emits mDNS beacons to find if there are more peers available. Local area network peers are very useful to peer-to-peer protocols, as low latency links are very useful.
> mDNS-discovery is a standalone protocol and does not depend on any other libp2p protocol. mDNS-discovery can yield peers available in the local area network, without relying on other infrastructure. This is particularly useful in intranets, networks disconnected from the internet backbone, and networks who temporarily loose links.
> mDNS-discovery can be configured per-service (i.e. discover only peers participating in a specific protocol, like IPFS), and with private networks (discover peers belonging to a private network).
> We are exploring ways to make mDNS-discovery beacons encrypted (so that other nodes in the local network cannot discern what service is being used). Though the nature of mDNS will always reveal local IP addresses.
> **Privacy Note**: mDNS advertises in local area networks, which reveals IP addresses to listeners in the same local network. It is not recommended to use this with privacy-sensitive applications or oblivious routing protocols.

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### 4.1.1.2 random-walk
+
+Random walking consists on traversing the routing tables of the peers available in the network, learning about new peers along the way.
```
jbenet: - "traversing the routing tables of the peers available" is ambiguous. maybe:

> Random-Walk is a Discovery Protocol for DHTs (and other protocols with routing tables). It makes random DHT queries in order to learn about a large number of peers quickly. This causes the DHT (or other protocol) to converge much faster, at the expense of a small load at the very beginning.

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+#### 4.1.1.3 bootstrap-list
+
+A predefined set of peers available on the network in well known locations, so that a new created node can easily find more peers to connect in the network.
```
jbenet: - Bootstrap-list should be "Discovery Protocols" section

> Bootstrap-List is a Discovery Protocol that uses local storage to cache the addresses of highly stable (and somewhat trusted) peers available in the network. This allows protocols to "find the rest of the network". This is essentially the same way that DNS bootstraps itself. (though note that changing the DNS bootstrap list  --the "dot domain" addresses -- is not easy to do, by design).
> - The list should be stored in long-term local storage, whatever that means to the local node (e.g. to disk)
> - Protocols can ship a default list hardcoded or along with the standard code distribution (like DNS)
> - In most cases (and certainly in the case of IPFS) the Bootstrap-List should be user configurable, as users may wish to establish separate networks, or place their reliance and trust in specific nodes.

protocol/network/4-architecture.md
diasdavid: > @diasdavid i think you merged "NAT Traversal" "Connection Relaying" and "Protocol Muxing" into your node(js)-swarm, but i do not think think this should be merged in the libp2p spec, nor on Go. The go swarm only handles transport multiplexing (connections/streams) to many peers. 

Bringing the protocol multiplexing to something called `host` would be a simple change and essentially one more module. One question, doesn't that mean that swarm doesn't know how to multistream dialed connections into spdy?

> NAT and Relay are separate protocols. In Go, the "Host" handles the protocol multiplexing, and composes (uses) the NAT and Relay protocols.

The way I was pushing forward was that a `transport` implementation is free to be implemented with any feature it needs as long as it respects the interface dial + listen. for example, a libp2p-webrtc could be implemented with a mechanism to do the signalling through the Kademlia Router or some other server, for Swarm, it would only matter that dial(PeerInfo) would dial to that PeerInfo and if some options if preferred, then dial(PeerInfo, {relay: true}) for e.g.

What is the full extent of `Host`?(also happy to call it Node)

protocol/network/4-architecture.md
diasdavid: > the "not three systems" above changes this picture to have other things (namely, NAT, and Relay). there are likely to be other protocols "that peer" here too. and the reason "NAT" and "Relay" can't be part of "the thing named Swarm here" is that "NAT" and "Relay" make use of "Peer Routing" and "Peer Routing" makes use of "NAT" and "Relay". They're really peering-protocols.

Oh, I believe I'm starting to see what you mean. NAT and Relay can use Peer Routing because of the `contract`that any of these subsystems will respect the same interface and expectations. For e.g Kad-Record-Store uses Kad-Routing to perform it's Kademlia queries 

I actually thought this was pretty neat and was pretty happy when I understood I could break logic apart by defining expectations. libp2p becomes a big composition of other modules

protocol/network/4-architecture.md
diasdavid: True, but they are used by kad-routing. They can be on their own category too, felt this showed the connection between those implementations and the Peer Routing algorithm

jbenet: yeah makes sense to list that it _can use_ them, but doesn't make sense for it to _define_ them, as a reader would think these are just _parts_ of kad-routing instead of general things, or that _kad-routing_ always (or most of the time) uses them. in reality, many ipfs nodes out there don't use mDNS. lots of people turn it off because VPSes dont like it

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+### 4.2.1 Stream Muxer
+
+The stream muxer must implement the interface offered by [abstract-stream-muxer](https://github.com/diasdavid/abstract-stream-muxer).
+
+### 4.2.2 Protocol Muxer
+
+Protocol muxing is handled on the application level instead of the conventional way at the port level (where a different services/protocols listen at different ports). This enables us to support several protocols to be muxed in the same socket, saving the cost of doing NAT traversal for more than one port.
+
+Protocol multiplexing is done through [`multistream`](https://github.com/jbenet/multistream), a protocol to negoatiate different types of streams (protocols) using [`multicodec`](https://github.com/jbenet/multicodec)
```
jbenet: - @diasdavid: do you agree about `libp2p.Node` being the thing that does protocol muxing? i could be convinced otherwise, so we should discuss it if you feel strongly about it. I do think `libp2p.Node` should exist though, and NAT and Relay be part of _that_ instead of part of the `Swarm`.

protocol/network/4-architecture.md
```diff
@@ -0,0 +1,126 @@
...
+
+### 4.2.3 Transport
+
+### 4.2.4 Crypto
+
+### 4.2.5 Identify
+
+**Identify** is one of the protocols mounted on top of swarm, our Connection handler, however, it follows and respects the same pattern as any other protocol when it comes to mounting it on top of swarm. Identify enables us to trade listenAddrs and observedAddrs between peers, this is crucial for the working of IPFS, since every socket open implements REUSEPORT, an observedAddr by another peer can enable a third peer to connect to us, since the port will be already open and redirect to us on a NAT.
+
+### 4.2.6 Relay
```
jbenet: i think we can collect `{Identify, Relay, NAT (there's multiple sub protocols for NAT), etc}` into a heading like `"Connectivity Protocols"`, which layer over Swarm, but are part of (under) libp2p (`libp2p.Node`).

protocol/network/5-datastructures.md
```diff
@@ -0,0 +1,13 @@
...
+The network protocol deals with these datastructures:
+
+- a `PrivateKey`, the private key of a node.
+- a `PublicKey`, the public key of a node.
+- a `PeerID`, a hash of a node's public key.
+- a `Node`[1], has a PeerID, and open connections to other `Nodes`.
+- a `Connection`, a point-to-point link between two Nodes (muxes 1 or more streams)
+- a `Stream`, a duplex message channel.
+
+[1] currently called `PeerHost` in go-ipfs.
```
jbenet: - yep, :+1: on the rename. (confirming for readers)

protocol/network/4-architecture.md
diasdavid: We should definitely exchange ideas. I'm found of the modular way it is organised, I do not have any strong feelings about keeping Protocol Muxing inside Swarm, but I'm also not sure if that would make things more simpler. Currently, Protocol Muxing in Node.js Swarm has this API https://github.com/diasdavid/node-libp2p-swarm#accept-requests-on-a-specific-protocol

I feel also that `node := new (libp2p.Node)` or `var node = new Libp2p()` have a place, it is the big glue that packs all of this multi options and that can be passed around in a program. My wish is to have for IPFS custom builds with the protocols we support, one for Node.js and one for the browser, so that we can instantiate a `new BrowserLibp2p()` and it already has all of the transports and features we use in a browser.

diasdavid:
@jbenet really nice review, thank you! :D Also great to be doing this now as we are noticing some final polishing on how things are done (glad that we are doing that while we are jumping into go-libp2p too)

protocol/network/2-state-of-the-art.md
diasdavid: Thank you :):)

protocol/network/6-interfaces.md
```diff
@@ -0,0 +1,137 @@
...
+
+## 6.2 Peer Routing
+
+## 6.3 Swarm
+
+~~The network is abstracted through the swarm which presents a simplified interface for the remaining layers to have access to the network. This interface should look like:~~
+
+- `.openStream(peer, protocol)` - peer should contain the ID of the peer and its respective multiaddrs known.
+- `.registerHandler(protocol, handlerFunc)` - enable a protocol to be registered, so that another peer can open a stream to talk with us to that specific protocol
+- `.listen()` - to start listening for incoming connections and therefore opening of streams
```
jbenet: - :+1: on this interface.

how about the below. This assumes that we don't sink protocol muxing into the Swarm, but again, that's up for discussion as it may be relevant to. regardless, we want users to use a libp2p.Node, not the Swarm directly.

```go
// Node is the basic interface of libp2p. It represents a "Host" in the libp2p internet.
// It is capable of opening streams (either as a dialer or a listener) to other Nodes.
//
// The libp2p.Node uses a libp2p.Swarm internally to handle all connections and
// stream multiplexing.
//
// It is meant to have user protocols "mounted" onto the Node (as one would bind
// a socket and Listen, in a regular TCP/IP network stack). This is done with either
// the OpenStream or SetHandler functions (which represent Dialing or Listening) 
// for stream of a specific protocol. Only one "client process" may mount a protocol,
// that means that only one Handler is registered per protocol.ID. This tends to be
// the convenient thing to do, but may be up for discussion (e.g. being part of 
// multiple IPFS-DHTs, etc).
//
// The p2p.Node uses internal protocols (which are mounted on top, like any other
// protocol) such as Identify, NAT (traversal), and Relay, to achieve full connectivity
// on the networks it belongs to.
type p2p.Node interface {
  // OpenStream opens a reliable stream to peer.ID with given protocol.ID. 
  // If there is no prior connection to peer.ID, p2p.Node will attempt to establish one.
  OpenStream(peer.ID, protocol.ID)

  // SetHandler registers a function to be called whenever another peer opens a stream
  // with that specific protocol.ID. If HandlerFunc is nil (default), the handler writes an 
  // ProtocolNotFoundError back and closes the stream.
  SetHandler(protocol.ID, protocol.HandlerFunc)
  
  // Listen starts listening for incoming connections and therefore opening of streams
  // (this may need to handle the addresses, i'll look again at how go-libp2p does it).
  Listen()

  // this will need more things here, will look into go-libp2p to figure this out.
}
```

protocol/network/6-interfaces.md
```diff
@@ -0,0 +1,137 @@
...
+                              │ multiplex │
+                              │└ ─ ─ ─ ─ ┘│
+                              │┌ ─ ─ ─ ─ ┐│
+                              │ QUIC      │
+                              │└ ─ ─ ─ ─ ┘│
+                              │┌ ─ ─ ─ ─ ┐│
+                              │ others    │
+                              │└ ─ ─ ─ ─ ┘│
+                              └───────────┘
+```
```
jbenet: - i'm confused by why "multistream" and "transport" peer (should "multistream" be on top of Transport)?
- :+1: on stream muxer peering. it's weird because some transports (like QUIC) solve everything in one piece. not easy to represent this in a 2d box diagram without it looking like a mess :D )

protocol/network/6-interfaces.md
diasdavid: The interface has evolved a lil bit with the last 'Swarm revisit' to support multiprotocols, but it is essentially the same:

https://github.com/diasdavid/node-libp2p-swarm#usage

> we want users to use a libp2p.Node, not the Swarm directly.
Totally agree, we can have a 'front facing' interface from libp2p, in fact node-libp2p is just that, glue, take a peak at: 
https://github.com/diasdavid/node-libp2p#setting-everything-up

Glad to remove the prefixes and make it everything libp2p.<call>

protocol/network/7-properties.md
```diff
@@ -1,16 +1,105 @@
...
-# IPFS Protocol on the Wire
+## 7.1 Communication Model - Streams
 
-Authors: [Juan Benet](http://github.com/jbenet)
+The Network layer handles all the problems of connecting to a peer, and exposes
+simple bidirectional streams. Users can both open a new stream
+(`NewStream()`) and register a stream handler (`SetStreamHandler`). The user
+is then free to implement whatever wire messaging protocol she desires. This
+makes it easy to build peer-to-peer protocols, as the complexities of
+connectivity, multi-transport support, flow control, and so on, are handled.
```
jbenet: we should probably choose between:
```
- NewStream() or OpenStream()
- SetHandler() or SetStreamHandler()
```
if we want to emulate the plan9 socket interface, we can do:

```
DialStream(protocol.ID)
ListenStream(protocol.ID, HandlerFunc)
```

jbenet: as an aside, I will probably be implementing an interface in Go that looks like this:

```go
n := libp2p.NewNode(...all the config...)
proto, err := n.Mount("/dht") // or n.Bind
stream1, err := proto.Dial()
stream2, err := proto.Dial()
stream3, err := proto.Dial()
stream5, err := proto.Accept() // error because not listening yet
err := proto.Listen()
stream5, err := proto.Accept()
stream6, err := proto.Accept()
stream7, err := proto.Accept()
```

Aaaaand reroll (another interface)

```
// use Mount or Bind with Dial and Listen modes
n := libp2p.NewNode(...all the config...)
proto1, err := n.Bind("/echo", Listen)
proto2, err := n.Bind("/bitswap", DialAndListen)
proto3, err := n.Bind("/ping", Dial)
stream1, err := proto1.Dial() // error because proto1 is only bound to Listen
stream2, err := proto1.Dial() // ok
stream3, err := proto2.Dial() // ok
stream4, err := proto2.Accept() // ok
stream5, err := proto3.Dial() // ok
stream6, err := proto3.Accept() // error because proto2 is only bound to Dial
```
- bind is more bsd socket-y
- mount is more unix fs-y
- mount is more what _I_ like and mean.
- mount evokes filesystem endpoint pathspace, which is what we're exposing here
- mount sounds more plan9 friendly.
- i'll probably do mount unless someone can point out to me why Bind and Mount should be different steps.

cc @whyrusleeping for thoughts.

protocol/network/6-interfaces.md
diasdavid: > i'm confused by why "multistream" and "transport" peer (should "multistream" be on top of Transport)?

It seems that I really need to update this one to match what is described in the 'Revisited Swarm'. Yes, multistream is more of a companion that enables the upgrades to happen than a option for the connection.

protocol/network/4-architecture.md
jbenet: would this work for you?

```go
n := libp2p.BrowserNode()
```
or in js
```js
var libp2p = require('libp2p')
var libp2p = require('libp2p-browser') // or a thinner package for the browser
n = libp2p.BrowserNode()
```

protocol/network/7-properties.md
jbenet: @diasdavid note that the above interface doesn't need to be everywhere, but the "protocol object" (returned by `Bind` or `Mount`) is a convenience wrapper around the `Node`'s methods, which is idiomatic Go.

diasdavid: Not sure if I like Bind or Mount the better, but I do have a suggestion for proto.Dial :)

`n.Dial(PeerInfo, {protocol: <if any preference, otherwise use the first it can connect from the list of multiaddrs available and those that are supported by us>})`

jbenet: btw, if we _don't need_ the distinction between the dialer/listener it may be nice to just think of `NewStream`. but i'm not sure. im not sure we can completely remove directionality. it would sort of look like a TCP simultaneous open...

for example
```go
// in one side
n := p2p.NewNode(config)
p := n.Mount("/dht", DialAndListen)
s1, err := p.NewStream()
// s1 above and s2 here same pipe
```

```
// in the other side
n := p2p.NewNode(config)
p := n.Mount("/dht", DialAndListen)
s2, err := p.NewStream()
// s1 above and s2 here same pipe
```
- We would need some material to distinguish streams and imply that we do want simultaneous open to happen, or we want _two separate_ duplex pipes.
- It may be impossible to conveniently distinguish these.
- on the other hand, we may want to just leave that up to the user...

### \o/ another reroll \o/ 

**actually i think this is wrong and likely very confusing. interesting though. i'll leave it here for posterity. but not a serious suggestion:**

```go
n := libp2p.NewNode(...all the config...)
stream1, err := p.NewStream("/dht/1") // returns a stream mounted at /dht/1. other side can simultaneous-open to this one.

stream2, err := n.DialStream("/dht/0") // assigns next available "port number", dials out.
stream3, err := n.AcceptStream() // assigns next available "port number" (0 is implied).

// same as
p := n.Mount("/dht", DialAndListen)
s1, err := p.Dial("1")
s2, err := p.Dial("0") // assigns next port number
s3, err := p.Accept() // "0
```

protocol/network/7-properties.md
```diff
@@ -93,13 +182,7 @@ For example:
...
+### 7.5.5 Portable Encodings
 
 In order to be ubiquitous, we _must_ use hyper-portable format encodings, those that are easy to use in various other platforms. Ideally these encodings are well-tested in the wild, and widely used. There may be cases where multiple encodings have to be supported (and hence we may need a [multicodec](https://github.com/jbenet/multicodec) self-describing encoding), but this has so far not been needed.
-
 For now, we use [protobuf](https://github.com/google/protobuf) for all protocol messages exclusively, but other good candidates are [capnp](https://capnproto.org), [bson](http://bsonspec.org/), [ubjson](http://ubjson.org/).
-
-
-## Secure Communications
-
-The wire protocol is -- of course -- wrapped with encryption. We use cyphersuites similar to TLS. This is explained further in the [network spec](./#encryption).
```
jbenet: not sure where this went, but should be at least a note somewhere on this doc, as people will be looking for it. (there's also a link to section up in `7.5`)

protocol/network/8-implementations.md
```diff
@@ -0,0 +1,40 @@
+8 Implementations
+=================
+
+These are the list of known libp2p, or in another words, the components that respect the interfaces and expectations defined in the Interfaces chapter, that can be composed to make a working libp2p.
```
jbenet: langfix: 

> This is a list of known libp2p module implementations. They are components that respect the interfaces and expectations defined in the "Interfaces" chapter, and can be composed to make a working libp2p library.

protocol/network/figs/logo.svg
```diff
@@ -0,0 +1,13 @@
...
+    <title>logo</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="Page-1" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd" sketch:type="MSPage">
+        <text id="libp2p" sketch:type="MSTextLayer" font-family="Arial Black" font-size="144" font-weight="686">
+            <tspan x="0" y="158" fill="#62D1D4">lib</tspan>
+            <tspan x="191.953125" y="158" fill="#001C57">p2p</tspan>
+        </text>
+    </g>
+</svg>
```
jbenet: i love svg.

jbenet: it's like viewsource friendly images

protocol/routing/README.md
```diff
@@ -0,0 +1,74 @@
...
+  - Publish and fetch content (also providing)
+- Maintaining partial state of the network
+  - DHT
+  - kbucket
+
+## Overview
+
+The Routing Protocol is composed by three componenets, these are:
+- Interface: Our routing primitives that are offered for the user, such as finding and publishing content, including the storage and lookup of metadata (Providers).
+- Peer Discovery: Responsible for filling our kbucket with best candidates.
```
jbenet: i think discovery should be separate from routing. I should be able to run routing protocols without a single discovery protocol (or a "manual discovery" i.e. i call `ipfs swarm connect <ipfs-addr-of-a-node-i-know`)

protocol/routing/README.md
```diff
@@ -0,0 +1,74 @@
...
+- Peer-to-peer Structured Overlay Network (DHT): Algorithm for the implicit network organization, based on [Coral](http://iptps03.cs.berkeley.edu/final-papers/coral.pdf) and [mainlineDHT](http://www.bittorrent.org/beps/bep_0005.html)
```
jbenet: How about this:
> - a structured overlay network: algorithms for organizing the network and for finding other nodes and data they may be hosting. (One example of such a routing network is a Kademlia DHT).

many routing protocols will not be DHT. DHT really is what our DRS is, it's just that when we say DHT today, people think this:



and not this:

protocol/routing/discovery.md
```diff
@@ -0,0 +1,41 @@
...
+### bootstrap peer list
+
+List with known and trusted peers shipped with IPFS.
+
+- _How is this list updated?_
+- _Is this list updated periodically_?
+
+### random walk
+
+IPFS issues random Peer lookups periodically to refresh our kbucket if needed. For impl reference, see: https://github.com/ipfs/go-ipfs/blob/master/routing/dht/dht_bootstrap.go#L88-L109.
```
jbenet: - we won't have a kbucket if it's not kademlia.
- random-walk is DHT-specific (or "query-thing-specific"), so make a note of that. maybe even call it "dht-random-walk"

> On nodes running a well-formed network with queries which expose more peers (e.g. DHT), IPFS can issue random queries to discover more nodes along those paths. This is useful for quickly filling a routing table with other nodes.

protocol/routing/discovery.md
```diff
@@ -0,0 +1,41 @@
...
+
+### mDNS
+
+In addition to known peers and random lookups, IPFS also performs Peer Discovery through mDNS ([MultiCast DNS](https://tools.ietf.org/html/rfc6762))
+
+-_How offen do we issue this searches?_
+
```
jbenet: we should add

### PEX -- Peer Exchange

> Peers advertise their connections (or a subset of them that have the PEX protocol enabled) to their peers, that way peers can quickly sample a large swath of the network.

protocol/routing/README.md
diasdavid: Apologies, this `/routing` folder is from very long ago, before we moved Peer Routing to part of the network layer, I forgot to explicit deprecated it as there is lot of incoherences, I'll just delete it all together. The holy grail for all things network is on `/network`

protocol/network/2-state-of-the-art.md
diasdavid: note: I've added, this comment didn't get 'outdated' because it was a couple of lines below

protocol/network/2-state-of-the-art.md
diasdavid: note: Same here. I've added, this comment didn't get 'outdated' because it was a couple of lines below

protocol/network/4-architecture.md
diasdavid: Yep :)

diasdavid:
From @jbenet and @diasdavid discussion

- [x] Agree that Protocol Muxing should stay in Swarm (@jbenet and @diasdavid)
- [x] Agree that transport implementations should be the ones responsible to also offer NAT Traversal such as Hole punching and Relaying (@jbenet and @diasdavid)
- [x] Make Discovery its own standalone subsystem (@diasdavid)
- [ ] Finalize API interface names (@jbenet)

protocol/network/figs/logo.svg
diasdavid: ahaha true :) look at those xml name spaces! + they Sketch prints their brand as a comment, smart !

diasdavid:
ping :)

jbenet:
I'm going to merge this, we can make edits as we go.

diasdavid:
weeeee! :D

PR #37, opened 2015-11-08
WIP: IPLD spec

jbenet:
This PR adds a new IPLD spec.

Some things TODO:
- [x] paths: link to path issue in go-ipfs or go-ipld
- [ ] paths: list path resolving restrictions
- [x] paths: show examples of path resolving
- [ ] examples/bitcoin: make this a real txn
- [ ] more examples

@mildred @diasdavid could you review?

### DIFF
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -286,7 +286,7 @@ Suppose _this_ was the _exact order_ in the _Canonical Format_ (not json, but cb
 
 #### Path Restrictions
 
-There are some important problems that come about with path descriptions in Unix and the web. For a discussion see (TODO link to path issue in go-ipfs or go-ipld). In order to be compatible with the models and expectations of unix and the web, IPLD explicitly disallows paths with certain path components. **Note that the data itself _may_ still contain these properties (someone will do it, and there are legitimate uses for it). So it is only _Path Resolvers_ that MUST NOT resolve through those paths.** The restrictions are the same as typical unix and UTF-8 path systems:
+There are some important problems that come about with path descriptions in Unix and the web. For a discussion see [this discussion](https://github.com/ipfs/go-ipfs/issues/1710). In order to be compatible with the models and expectations of unix and the web, IPLD explicitly disallows paths with certain path components. **Note that the data itself _may_ still contain these properties (someone will do it, and there are legitimate uses for it). So it is only _Path Resolvers_ that MUST NOT resolve through those paths.** The restrictions are the same as typical unix and UTF-8 path systems:
 
 
 TODO:
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -1,0 +1,464 @@
+# IPLD -- the "thin-waist" merkle dag format.
+
+There are a variety of systems that use merkle-tree and hash-chain inspired datastructures (e.g. git, bittorrent, ipfs, tahoe-lafs, sfsro). IPLD defines:
+
+- **_merkle-links_**: the core unit of a merkle-graph
+- **_merkle-dag_**: any graphs whose edges are _merkle-links_.
+- **_merkle-paths_**: unix-style paths for traversing _merkle-dags_ with _named merkle-links**
+- **IPLD Data Model**: a flexible, JSON-based data model for representing merkle-dags.
+- **IPLD Serialized Formats**: a set of formats in which IPLD objects can be represented, for example JSON, CBOR, CSON, YAML, Protobuf, XML, RDF, etc.
+- **IPLD Canonical Format**: a deterministic description on a serialized format that ensures the same _logical_ object is always serialized to _the exact same sequence of bits_. This is critical for merkle-linking, and all cryptographic applications.
+
+In short: JSON documents with named merkle-links that can be traversed.
+
+## Intro
+
+### What is a _merkle-link_?
+
+A _merkle-link_ is a link between two objects which is content-addressed with the _cryptographic hash_ of the target object, and embedded in the source object. Content addressing with merkle-links allows:
+
+- **Cryptographic Integrity Checking**: resolving a link's value can be tested by hashing. In turn, this allows wide, secure, trustless exchanges of data (e.g. git or bittorrent), as others cannot give you any data that does not hash to the link's value.
+- **Immutable Datastructures**: data structures with merkle links cannot mutate, which is a nice property for distributed systems. This is useful for versioning, for representing distributed mutable state (eg CRDTs), and for long term archival.
+
+### What is a _merkle-graph_ or a _merkle-dag_?
+
+Objects with merkle-links form a Graph (merkle-graph), which necessarily is both Directed, and which can be counted on to be Acyclic, iff the properties of the cryptographic hash function hold. I.e. a _merkle-dag_. Hence all graphs which use _merkle-linking_ (_merkle-graph_) are necessarily also Directed Acyclic Graphs (DAGs, hence _merkle-dag_).
+
+### What is a _merkle-path_?
+
+A _merkle-path_ is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
+
+For example, suppose we have this _merkle-path_:
+
+```
+/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d
+```
+Where:
+- `ipfs` is a protocol namespace (to allow the computer to discern what to do)
+- `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` is a cryptographic hash.
+- `a/b/c/d` is a path _traversal_, as in unix.
+- this link traverses five objects.
+
+Resolving it involves looking up each object and attaining a hash value, then traversing to the next.
+
+```
+      +-------------------+
+O_1 = | "a": "QmV76pU..." |  whose hash value is QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_2 = | "b": "QmV76pU..." |  whose hash value is QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_3 = | "c": "QmV76pU..." |  whose hash value is QmQmkZPNPoRkPd7wj2xUJe5v5DsY6MX33MFaGhZKB2pRSE
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_4 = | "d": "QmV76pU..." |  whose hash value is QmWkyYNrN5wnHgX5vfs88q7QUaFKq52TVNTFeTzxm73UbT
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_5 = | "hello": "world"  |  whose hash value is QmR8Bzg59Y4FGWHeu9iTYhwhiP8PHCNFiaGhP1UjywA43j
+      +-------------------+
+```
+
+This entire _merkle-path_ traversal is a unix-style path traversal over a _merkle-dag_ which uses _merkle-links_ with names.
+
+## What is the IPLD Data Model?
+
+The IPLD Data Model defines a simple JSON-based _structure_ for all merkle-dags, and identifies a set of formats to encode the structure into.
+
+### Constraints and Desires
+
+Some Constraints:
+- IPLD paths MUST be unambiguous. A given path string MUST always deterministically traverse to the same object. (e.g. avoid duplicating link names)
+- IPLD paths MUST be universal and avoid opressing non-english societies (e.g. use UTF-8, not ASCII).
+- IPLD paths MUST layer cleanly over UNIX and The Web (use `/`, have deterministic transforms for ASCII systems).
+- Given the wide success of JSON, a huge number of systems present JSON interfaces. IPLD MUST be able to import and export to JSON trivially.
+- The JSON data model is also very simple and easy to use. IPLD MUST be just as easy to use.
+- Definining new datastructures MUST be trivially easy. It should not be cumbersome -- or require much knowledge -- to experiment with new definitions on top of IPLD.
+- IPLD MUST be compatible with RDF and the proper W3C Semantic Web / Linked Data standards. We can achieve this easily through JSON-LD.
+- IPLD Serialized Formats (on disk and on the wire) MUST be fast and space efficient. (should not use JSON as the storage format, and instead use CBOR or similar formats)
+- IPLD cryptographic hashes MUST be upgradeable (use [multihash](https://github.com/jbenet/multihash))
+
+Some nice-to-haves:
+- IPLD SHOULD NOT carry over mistakes, e.g. the lack of integers in JSON.
+- IPLD SHOULD be upgradable, e.g. if a better on-disk format emerges, systems should be able to migrate to it and minimize costs of doing so.
+- IPLD objects SHOULD be able to resolve properties too as paths, not just merkle links.
+- IPLD Canonical Format SHOULD be easy to write a parser for.
+- IPLD Canonical Format SHOULD enable seeking without parsing full objects. (CBOR and Protobuf allow this).
+
+
+### Format Definition
+
+(**NOTE:** Here we will use both JSON and YML to show what formats look like. We explicitly use both to show equivalence of the object across two different formats.)
+
+At its core, IPLD Data Model "is just JSON" in that it (a) is also tree based documents with a few primitive types, (b) maps 1:1 to json, (c) users can use it through JSON itself. It "is not JSON" in that (a) it improves on some mistakes, (b) has an efficient serialized representation, and (c) does not actually specify a single on-wire format, as the world is known to improve.
+
+#### Basic Node
+
+Here is an example IPLD object in JSON:
+
+```json
+{
+  "name": "Vannevar Bush"
+}
+```
+
+Suppose it hashes to the multihash value `QmAAA...AAA`. Note that it has no links at all, just a string name value. But we are still be able to "resolve" the key `name` under it:
+
+```sh
+> ipld cat --json QmAAA...AAA
+{
+  "name": "Vannevar Bush"
+}
+
+> ipld cat --json QmAAA...AAA/name
+"Vannevar Bush"
+```
+
+And -- of course -- we are able to view it in other formats
+
+```sh
+> ipld cat --yml QmAAA...AAA
+---
+name: Vannevar Bush
+
+> ipld cat --xml QmAAA...AAA
+<!xml> <!-- todo -->
+<node>
+  <name>Vannevar Bush</name>
+</node>
+```
+
+#### Linking Between Nodes
+
+Merkle-Linking between nodes is the reason for IPLD to exist. A Link in IPLD is just an embedded node with a special format:
+
+```js
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
+  }
+}
+```
+
+Suppose this hashes to the multihash value `QmBBB...BBB`. This node links the _subpath `author` to `QmAAA...AAA`, the node in the section above. So we can now do:
+
+```sh
+> ipld cat --json QmBBB...BBB
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
+  }
+}
+
+> ipld cat --json QmBBB...BBB/author
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
+  }
+}
+
+> ipld cat --yml QmBBB...BBB/author
+---
+title: As We May Think
+author:
+  mlink: QmAAA...AAA
+
+> ipld cat --json QmBBB...BBB/author/name
+"Vannevar Bush"
+```
+
+#### Link Properties
+
+IPLD allows for links to have other properties themselves. This is useful to encode other invormation into a link, such as the kind of relationship, or ancilliary data required in the link. This is _different from_ "Link Objects", discussed below, which are very useful in their own right. But sometimes, you just want to add a bit of data on the link and not have to make another object. IPLD doesn't get in your way.
+
+For example, supposed you have a file system, and want to assign metadata like permissions, or owners in the link between objects. Suppose you have a `directory` object with hash `QmCCC...CCC` like this:
+
+```js
+{
+  "foo": {
+    "mlink": "QmCCC...111"
+    "mode": "0755",
+    "owner": "jbenet"
+  },
+  "cat.jpg": {
+    "mlink": "QmCCC...222"
+    "mode": "0644",
+    "owner": "jbenet"
+  },
+  "doge.jpg": {
+    "mlink": "QmCCC...333",
+    "mode": "0644",
+    "owner": "jbenet"
+  }
+}
+```
+
+or in YML
+
+```yml
+---
+foo:
+  mlink: QmCCC...111
+  mode: 0755
+  owner: jbenet
+cat.jpg:
+  mlink: QmCCC...222
+  mode: 0644
+  owner: jbenet
+doge.jpg:
+  mlink: QmCCC...333
+  mode: 0644
+  owner: jbenet
+```
+
+Though we have new properties in the links that are _specific to this datastructure_, we can still resolve links just fine:
+
+```js
+> ipld cat --json QmCCC...CCC/cat.jpg
+{
+  "data": "\u0008\u0002\u0012��\u0008����\u0000\u0010JFIF\u0000\u0001\u0001\u0001\u0000H\u0000H..."
+}
+
+> ipld cat --json QmCCC...CCC/doge.jpg
+{
+  "subfiles": [
+    {
+      "mlink": "QmPHPs1P3JaWi53q5qqiNauPhiTqa3S1mbszcVPHKGNWRh"
+    },
+    {
+      "mlink": "QmPCuqUTNb21VDqtp5b8VsNzKEMtUsZCCVsEUBrjhERRSR"
+    },
+    {
+      "mlink": "QmS7zrNSHEt5GpcaKrwdbnv1nckBreUxWnLaV4qivjaNr3"
+    }
+  ]
+}
+
+> ipld cat --yml QmCCC...CCC/doge.jpg
+---
+subfiles:
+  - mlink: QmPHPs1P3JaWi53q5qqiNauPhiTqa3S1mbszcVPHKGNWRh
+  - mlink: QmPCuqUTNb21VDqtp5b8VsNzKEMtUsZCCVsEUBrjhERRSR
+  - mlink: QmS7zrNSHEt5GpcaKrwdbnv1nckBreUxWnLaV4qivjaNr3
+
+> ipld cat --json QmCCC...CCC/doge.jpg/subfiles/1
+{
+  "data": "\u0008\u0002\u0012��\u0008����\u0000\u0010JFIF\u0000\u0001\u0001\u0001\u0000H\u0000H..."
+}
+```
+
+But we can't extract the link as nicely as other properties, as links are meant to _resolve through_.
+
+#### Duplicate property keys
+
+Note that having two properties with _the same_ name IS NOT ALLOWED, but actually impossible to prevent (someone will do it and feed it to parsers), so to be safe, we define the value of the path traversal to be _the first_ entry in the serialized representation. For example, suppose we have the object:
+
+```json
+{
+  "name": "J.C.R. Licklider",
+  "name": "Hans Moravec"
+}
+```
+
+Suppose _this_ was the _exact order_ in the _Canonical Format_ (not json, but cbor), and it hashes to `QmDDD...DDD`. We would _ALWAYS_ get:
+
+```sh
+> ipld cat --json QmDDD...DDD
+{
+  "name": "J.C.R. Licklider",
+  "name": "Hans Moravec"
+}
+> ipld cat --json QmDDD...DDD/name
+"J.C.R. Licklider"
+```
+
+
+#### Path Restrictions
+
+There are some important problems that come about with path descriptions in Unix and the web. For a discussion see (TODO link to path issue in go-ipfs or go-ipld). In order to be compatible with the models and expectations of unix and the web, IPLD explicitly disallows paths with certain path components. **Note that the data itself _may_ still contain these properties (someone will do it, and there are legitimate uses for it). So it is only _Path Resolvers_ that MUST NOT resolve through those paths.** The restrictions are the same as typical unix and UTF-8 path systems:
+
+
+TODO:
+- [ ] list path resolving restrictions
+- [ ] show examples
+
+#### Integers in JSON
+
+IPLD is _directly compatible_ with JSON, to take advantage of JSON's successes, but it need not be _held back_ by JSON's mistakes. This is where we can afford to follow format idiomatic choices, though care MUST be given to ensure there is always a well-defined 1:1 mapping.
+
+On the subject of integers, there exist a variety of formats which represent integers as strings in JSON, for example, [EJSON](https://www.meteor.com/ejson). These can be used and conversion to and from other formats should happen naturally-- that is, when converting JSON to CBOR, an EJSON integer should be transformed naturally to a proper CBOR integer, instead of representing it as a map with string values.
+
+
+## Serialized Data Formats
+
+IPLD supports a variety of serialized data formats trough [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `mlink`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
+
+### Canonical Format
+
+In order to preserve merkle-linking's power, we muste ensure that there is a single **_canonical_** serialized representation of an IPLD document. This ensures that applications arrive at the same cryptographic hashes. It should be noted --though-- that this is a system-wide parameter. Future systems might change it to evolve representations. However we estimate this would need to be done no more than once per decade.
+
+**The IPLD Canonical format is _canonicalized CBOR_.**
+
+
+## Datastructure Examples
+
+It is important that IPLD be a simple, nimble, and flexible format that does not get in the way of users defining new or importing old datastractures. For this purpose, below I will show a few example data structures.
+
+
+### Unix Filesystem
+
+
+#### A small File
+
+```js
+{
+  "data": "hello world",
+  "size": "11"
+}
+```
+
+#### A Chunked File
+
+Split into multiple independent sub-Files.
+
+```js
+{
+  "size": "1424119",
+  "subfiles": [
+    {
+      "mlink": "QmAAA...",
+      "size": "100324"
+    },
+    {
+      "mlink": "QmAA1...",
+      "size": "120345",
+      "repeat": "10"
+    },
+    {
+      "mlink": "QmAA1...",
+      "size": "120345"
+    },
+  ]
+}
+```
+
+#### A Directory
+
+```js
+{
+  "foo": {
+    "mlink": "QmCCC...111"
+    "mode": "0755",
+    "owner": "jbenet"
+  },
+  "cat.jpg": {
+    "mlink": "QmCCC...222"
+    "mode": "0644",
+    "owner": "jbenet"
+  },
+  "doge.jpg": {
+    "mlink": "QmCCC...333",
+    "mode": "0644",
+    "owner": "jbenet"
+  }
+}
+```
+
+### git
+
+#### git blob
+
+```js
+{
+  "data": "hello world"
+}
+```
+
+#### git tree
+
+```js
+{
+  "foo": {
+    "mlink": "QmCCC...111"
+    "mode": "0755"
+  },
+  "cat.jpg": {
+    "mlink": "QmCCC...222"
+    "mode": "0644"
+  },
+  "doge.jpg": {
+    "mlink": "QmCCC...333",
+    "mode": "0644"
+  }
+}
+```
+
+#### git commit
+
+```js
+{
+  "tree": {"mlink": "e4647147e940e2fab134e7f3d8a40c2022cb36f3"},
+  "parents": [
+    {"mlink": "b7d3ead1d80086940409206f5bd1a7a858ab6c95"},
+    {"mlink": "ba8fbf7bc07818fa2892bd1a302081214b452afb"}
+  ],
+  "author": {
+    "name": "Juan Batiz-Benet",
+    "email": "juan@benet.ai",
+    "time": "1435398707 -0700"
+  },
+  "committer": {
+    "name": "Juan Batiz-Benet",
+    "email": "juan@benet.ai",
+    "time": "1435398707 -0700"
+  },
+  "message": "Merge pull request #7 from ipfs/iprs\n\n(WIP) records + merkledag specs"
+}
+```
+
+### Bitcoin
+
+#### Bitcoin Block
+
+```js
+{
+  "parent": {"mlink": "Qm000000002CPGAzmfdYPghgrFtYFB6pf1BqMvqfiPDam8"},
+  "transactions": {"mlink": "QmTgzctfxxE8ZwBNGn744rL5R826EtZWzKvv2TF2dAcd9n"},
+  "nonce": "UJPTFZnR2CPGAzmfdYPghgrFtYFB6pf1BqMvqfiPDam8"
+}
+```
+
+#### Bitcoin Transaction
+
+This time, im YML. TODO: make this a real txn
+
+```yml
+---
+inputs:
+  - input: {mlink: Qmes5e1x9YEku2Y4kDgT6pjf91TPGsE2nJAaAKgwnUqR82}
+    amount: 100
+outputs:
+  - output: {mlink: Qmes5e1x9YEku2Y4kDgT6pjf91TPGsE2nJAaAKgwnUqR82}
+    amount: 50
+  - output: {mlink: QmbcfRVZqMNVRcarRN3JjEJCHhQBcUeqzZfa3zoWMaSrTW}
+    amount: 30
+  - output: {mlink: QmV9PkR2gXcmUgNH7s7zMg9dsk7Hy7bLS18S9SHK96m7zV}
+    amount: 15
+  - output: {mlink: QmP8r8fLUnEywGnRRUrHB28nnBKwmshMLiYeg8udzYg7TK}
+    amount: 5
+script: OP_VERIFY
+```
+
+
+
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -33,11 +33,14 @@ For example, suppose we have this _merkle-path_:
 ```
 /ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d
 ```
+
 Where:
 - `ipfs` is a protocol namespace (to allow the computer to discern what to do)
 - `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` is a cryptographic hash.
 - `a/b/c/d` is a path _traversal_, as in unix.
 - this link traverses five objects.
+
+Suppose also that this path points to the object `{ "hello": "world" }`.
 
 Resolving it involves looking up each object and attaining a hash value, then traversing to the next.
 
@@ -82,7 +85,7 @@ Some Constraints:
 - Given the wide success of JSON, a huge number of systems present JSON interfaces. IPLD MUST be able to import and export to JSON trivially.
 - The JSON data model is also very simple and easy to use. IPLD MUST be just as easy to use.
 - Definining new datastructures MUST be trivially easy. It should not be cumbersome -- or require much knowledge -- to experiment with new definitions on top of IPLD.
-- IPLD MUST be compatible with RDF and the proper W3C Semantic Web / Linked Data standards. We can achieve this easily through JSON-LD.
+- Since IPLD is based on the JSON data model, it is fully compatible with RDF and Linked Data standards through JSON-LD.
 - IPLD Serialized Formats (on disk and on the wire) MUST be fast and space efficient. (should not use JSON as the storage format, and instead use CBOR or similar formats)
 - IPLD cryptographic hashes MUST be upgradeable (use [multihash](https://github.com/jbenet/multihash))
 
@@ -162,17 +165,12 @@ Suppose this hashes to the multihash value `QmBBB...BBB`. This node links the _s
 
 > ipld cat --json QmBBB...BBB/author
 {
-  "title": "As We May Think",
-  "author": {
-    "mlink": "QmAAA...AAA" // links to the node above.
-  }
+  "name": "Vannevar Bush"
 }
 
 > ipld cat --yml QmBBB...BBB/author
 ---
-title: As We May Think
-author:
-  mlink: QmAAA...AAA
+name: "Vannevar Bush"
 
 > ipld cat --json QmBBB...BBB/author/name
 "Vannevar Bush"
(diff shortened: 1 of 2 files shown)


jbenet:
Another TODO:

- [ ] multicodec: attribute to signal the canonical multicodec for a datastructure. I.e. if a datastructure _ought_ to use a canonical serializing format (e.g. different than cbor), we can use an attribute like  `@multicodec` to signal that to implementations. 

It may need to be stored in the serialized format, since decoding it will yield a multicodec. maybe just in the _in memory logical_ representation, so serialization happens correctly. (With the exception for old style `proto/mdagv1`) (thoughts @mildred?)

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+
+
+#### Path Restrictions
+
+There are some important problems that come about with path descriptions in Unix and the web. For a discussion see [this discussion](https://github.com/ipfs/go-ipfs/issues/1710). In order to be compatible with the models and expectations of unix and the web, IPLD explicitly disallows paths with certain path components. **Note that the data itself _may_ still contain these properties (someone will do it, and there are legitimate uses for it). So it is only _Path Resolvers_ that MUST NOT resolve through those paths.** The restrictions are the same as typical unix and UTF-8 path systems:
+
+
+TODO:
+- [ ] list path resolving restrictions
+- [ ] show examples
```
jbenet: @willglynn could you help me fill this out? 

the more concise here the better, but i suspect this section may be a bit large.

see rendered doc here: https://github.com/ipfs/specs/blob/ipld-spec/merkledag/ipld.md

jbenet:
More TODOs:
- [ ] `@attributes`: describe `@` escaping for the potential future `@attributes`. (cc @mildred)
- [ ] Linked Data: add a section describing relationship to _proper_ linked data formats, JSON-LD, etc.
- [ ] Linked Data: add a JSON-LD example
- [ ] Linked Data: add an RDF example

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+```sh
+> ipld cat --json QmBBB...BBB
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
+  }
+}
+
+> ipld cat --json QmBBB...BBB/author
```
noffle: shouldn't this resolve to
```json
"mlink": "QmAAA...AAA" // links to the node above.
```
? Same for the YAML example below.

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_4 = | "d": "QmV76pU..." |  whose hash value is QmWkyYNrN5wnHgX5vfs88q7QUaFKq52TVNTFeTzxm73UbT
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_5 = | "hello": "world"  |  whose hash value is QmR8Bzg59Y4FGWHeu9iTYhwhiP8PHCNFiaGhP1UjywA43j
```
noffle: Might it improve reading comprehensibility if you mention beforehand that `"hello": "world"` is stored at the path `/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d`?

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+
+On the subject of integers, there exist a variety of formats which represent integers as strings in JSON, for example, [EJSON](https://www.meteor.com/ejson). These can be used and conversion to and from other formats should happen naturally-- that is, when converting JSON to CBOR, an EJSON integer should be transformed naturally to a proper CBOR integer, instead of representing it as a map with string values.
+
+
+## Serialized Data Formats
+
+IPLD supports a variety of serialized data formats trough [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `mlink`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
```
noffle: s/trough/through

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+
+### Canonical Format
+
+In order to preserve merkle-linking's power, we muste ensure that there is a single **_canonical_** serialized representation of an IPLD document. This ensures that applications arrive at the same cryptographic hashes. It should be noted --though-- that this is a system-wide parameter. Future systems might change it to evolve representations. However we estimate this would need to be done no more than once per decade.
```
noffle: s/muste/must

noffle:
I left a few nitpicks. On the whole I found this doc hugely helpful in understanding IPLD and its intent.

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+For example, suppose we have this _merkle-path_:
+
+```
+/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d
+```
+Where:
+- `ipfs` is a protocol namespace (to allow the computer to discern what to do)
+- `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` is a cryptographic hash.
+- `a/b/c/d` is a path _traversal_, as in unix.
+- this link traverses five objects.
```
mildred: Not necessarily.

We could have the object `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` having a link named `a/b/c/d` directly pointing to the final object (or any combination in between).

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+
+## What is the IPLD Data Model?
+
+The IPLD Data Model defines a simple JSON-based _structure_ for all merkle-dags, and identifies a set of formats to encode the structure into.
+
+### Constraints and Desires
+
+Some Constraints:
+- IPLD paths MUST be unambiguous. A given path string MUST always deterministically traverse to the same object. (e.g. avoid duplicating link names)
+- IPLD paths MUST be universal and avoid opressing non-english societies (e.g. use UTF-8, not ASCII).
```
mildred: Just nit picking but ASCII is defined on 7 bits only. What we have generally in unix systems as filepath encoding is arbitrary 8bit encoding that cannot be qualified as ASCII. It could be qualified as a binary string or locale dependent encoded string.

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+- IPLD paths MUST layer cleanly over UNIX and The Web (use `/`, have deterministic transforms for ASCII systems).
```
mildred: Just nit picking but ASCII is defined on 7 bits only, and UTF-8 (in which we encode to) is 8 bits. What we have generally in unix systems as filepath encoding is arbitrary 8bit encoding that cannot be qualified as ASCII. It could be qualified as a binary string or locale dependent encoded string.

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+- Given the wide success of JSON, a huge number of systems present JSON interfaces. IPLD MUST be able to import and export to JSON trivially.
+- The JSON data model is also very simple and easy to use. IPLD MUST be just as easy to use.
+- Definining new datastructures MUST be trivially easy. It should not be cumbersome -- or require much knowledge -- to experiment with new definitions on top of IPLD.
+- IPLD MUST be compatible with RDF and the proper W3C Semantic Web / Linked Data standards. We can achieve this easily through JSON-LD.
```
mildred: *compatibility* is a very wide term. If this is a MUST, then we should probably define it more clearly. Or else mode veis elsewhere by saying for example:

> As a result of this design and IPLD being based on the JSON data model, it is fully compatible with RDF and Linked Data standards.

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+```js
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
+  }
+}
+```
+
+Suppose this hashes to the multihash value `QmBBB...BBB`. This node links the _subpath `author` to `QmAAA...AAA`, the node in the section above. So we can now do:
```
mildred: Perhaps we should define this more precisely. What are links ?

In go-ipld, this is defined by a node having a `mlink` key that points to a string. it is assumed that the string content is a valid multihash but that is only checked when trying to access the link.

merkledag/ipld.md
mildred: At least, when IPFS attains a stable release, this should not be changed.

mildred:
> Another TODO:
>
> - [ ] multicodec: attribute to signal the canonical multicodec for a datastructure. I.e. if a datastructure ought to use a canonical serializing format (e.g. different than cbor), we can use an attribute like @multicodec to signal that to implementations.
>
> It may need to be stored in the serialized format, since decoding it will yield a multicodec. maybe just in the in memory logical representation, so serialization happens correctly. (With the exception for old style proto/mdagv1) (thoughts @mildred?)

In that case, we should either provision key escaping to avoid clashing with someone wanting to have `"@multicodec"` as an arbitrary key in his JSON data structure (think a `@multicodec` in a directory object). Or we should transmit this information outside of the IPLD object itself (not encoded as part of the datastructure).

mildred:
:+1: on this whole thing. Still a major TODO for me is the description of the `@` character escaping for introducing IPLD specific directives.

By the way, why choose the `@` character to introduce directives? This might be confusing in the context of JSON-LD where the same character is used (but it will be escaped in the IPLD object). We are free to choose anything (`$%!#&?*+`...)

merkledag/ipld.md
jbenet: @mildred ah that's right we did say we would allow sparse. one question remains re ordering of links there-- do we want to take them by lexicographic order? or in order in the serialized fmt? 

ordering based on the serialized fmt will be needed if links have same name/no name (someone WILL do it so ipld implementations should be written to handle the case even if we say people should not do it)

but ordering lexicographically when links do have names is useful for getting users to expect the same behavior. 

how do we handle this?
```sh
> ipld cat --fmt yml $h1
---
foo: {mlink: $h2}
foo/bar: {mlink: $h3}

> ipld cat --fmt yml $h2
---
bar:
 hello: h2bar1

> ipld cat --fmt yml $h3
---
hello: h3bar2

> ipld cat --fmt yml $h1/foo/bar
# ??? should it be
---
hello: h2bar1

# or should it be
---
hello: h3bar2
```

mildred: I would not expect this to be an issue because what you describe is the compact form of the following object:

    > ipld cat --fmt yml $h1
    ---
    foo: 
      mlink: $h2
      bar:
        mlink: $h3

And in this object, only the "foo" link will be considered valid, not "foo/bar" because :

- keys should not be allowed to have `/` character (so we will never have "foo/bar" verbatim in a key) for the same reasons `/` is not allowed in a unix filename.

- in the previous object, "foo/bar" will not be considered as a link as per the  "Duplicate property keys" section :

    >   Note that having two properties with the same name IS NOT ALLOWED, but actually impossible to prevent (someone will do it and feed it to parsers), so to be safe, we define the value of the path traversal to be the first entry in the serialized representation. For example, suppose we have the object:

mildred: ipld cat --fmt yml $h1/foo/bar

should be `h2bar1`

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+There are a variety of systems that use merkle-tree and hash-chain inspired datastructures (e.g. git, bittorrent, ipfs, tahoe-lafs, sfsro). IPLD defines:
+
+- **_merkle-links_**: the core unit of a merkle-graph
+- **_merkle-dag_**: any graphs whose edges are _merkle-links_.
+- **_merkle-paths_**: unix-style paths for traversing _merkle-dags_ with _named merkle-links**
+- **IPLD Data Model**: a flexible, JSON-based data model for representing merkle-dags.
+- **IPLD Serialized Formats**: a set of formats in which IPLD objects can be represented, for example JSON, CBOR, CSON, YAML, Protobuf, XML, RDF, etc.
+- **IPLD Canonical Format**: a deterministic description on a serialized format that ensures the same _logical_ object is always serialized to _the exact same sequence of bits_. This is critical for merkle-linking, and all cryptographic applications.
+
+In short: JSON documents with named merkle-links that can be traversed.
```
diasdavid: traversed or resolved?

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+
+#### Linking Between Nodes
+
+Merkle-Linking between nodes is the reason for IPLD to exist. A Link in IPLD is just an embedded node with a special format:
+
+```js
+{
+  "title": "As We May Think",
+  "author": {
+    "mlink": "QmAAA...AAA" // links to the node above.
```
diasdavid: Better start calling them `merkle-link` so that we can have a clear distinction with `mazieres-link`

merkledag/ipld.md
```diff
@@ -0,0 +1,464 @@
...
+Resolving it involves looking up each object and attaining a hash value, then traversing to the next.
+
+```
+      +-------------------+
+O_1 = | "a": "QmV76pU..." |  whose hash value is QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k
+      +-------------------+
+                |
+                v
+      +-------------------+
+O_2 = | "b": "QmV76pU..." |  whose hash value is QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT
```
diasdavid: The hashes on the pointers are repeated

diasdavid:
I don't see any specific mention to the data types the links point to, have we decided that isn't IPLD turf (and just json-ld)?

diasdavid:
Also, instead of adding more TODO here, can we before go issue by issue here https://github.com/ipfs/go-ipld/issues and https://github.com/diasdavid/js-ipld/issues/2 and check what is already included in the spec, add what is missing, explicitly add what was accepted and then see what's left, bringing all of that conversation to this PR? So that we all talk over the same page :)

mildred:
> I don't see any specific mention to the data types the links point to, have we decided that isn't IPLD turf (and just json-ld)?

It is possible to describe the merkle link in a JSON-LD context easily, but the context must also describe other parts of the JSON document. So it can't be a single context for all IPLD documents.

So, you'll have different contexts for small files, chunked files, directory, git blob, tree, commit, ..., and each of these contexts will be able to describe the merkle links the same way.

Also, we could imagine allowing any arbitrary key name instead of `"mlink"` and use the provided JSON-LD context to know which key is a merkle link, but this would mean parsing the JSON-LD context for every IPLD operation, which I believe would be way too slow.

candeira:
From IRC:

< kandinski> jbenet: the IPLD spec doesn't define where the "IPLD" initials come from
< kandinski> Linked DAG?
<@jbenet> kandinski: good point. Linked Data. but it's not "legit linked data"
<@jbenet> linked dag could work :)
< kandinski> It's InterPlanetary Linked Data
<@jbenet> but will confuse people
< dignifiedquire> but InterPlanetaryLegitData sounds way cooler

In any case, origin of initials should be referenced at the top. 

Since it's going to be pronounced "Eye Pee Ell Dee" anyway, we could say something like:

"The IPLD (short for InterPlanetary Linked DirectedAcyclicGraph) is...". We do call it a "thin-waist merkle dag" anyway, which is a killer description, by the way.

candeira:
Same as above with the first mention of CDRTs. "Conflict-free replicated data type" is easier to understand for people coming in anew.

merkledag/ipld.md
eminence: I think you're right and this was a copy/paste typo

jbenet:
cc @mekarpeles

jbenet:
https://github.com/ipfs/specs/blob/ipld-spec/merkledag/ipld.md <--

merkledag/ipld.md
jbenet: @mildred i wonder if we should be specifying the canonical encoding in the expanded data itself (or at least allowing it to be overridden). storing the data will include the multicodec header, but when we parse it into logical representations (or even other wire formats), we might want to specify the canonical format.

Aside, for future thought, this is something that a `@type` could do -- similar to `@context` it could define some defaults that hold for all instances. So we _could have_ something like:

- some chosen canonical multicodec default (in our case, cbor)
- can override canonical multicodec in the object's type or context.
- can override canonical multicodec in the object itself

mildred: I don't quite understand what you propose here. In my opinion, there isn't so much value to allow arbitrary wire representation.

merkledag/ipld.md
jbenet: > - keys should not be allowed to have `/` character

i agree, but unlike unix pathnames, there already are datastructs out there that we should be able to store, even if the resolution through them is not perfect. I.e. if we _define_ how the resolution would work even in this case, we avoid the problem of forcing users to change their data.

merkledag/ipld.md
jbenet: ok, want to adjust the language directly? what should it say?

jbenet:
@mildred 

> In that case, we should either provision key escaping to avoid clashing with someone wanting to have "@multicodec" as an arbitrary key in his JSON data structure

Yep, agreed. what's a good place to put that?

jbenet:
> By the way, why choose the `@` character to introduce directives? This might be confusing in the context of JSON-LD where the same character is used (but it will be escaped in the IPLD object). We are free to choose anything `($%!#&?*+...)`

It was chosen because of JSON-LD. We could deviate,. we could use `$`,  `_`, `!` (it should be able to appear in a path / URL.

remaining question: should the text in paths be the raw case, or unescaped? eg, should this be possible or inaccessible?
  
```
> .../foo/bar/@multicodec
```

merkledag/ipld.md
jbenet: both

merkledag/ipld.md
jbenet: that will blow up verbosity and storage size. I was thinking, maybe we should drop `mlink` in favor of just `link` or `@link` (or the special char we go for). the link itself can denote what type of link it is:
- if it has no prefix (not a path, just a hash), it is a merkle-link.  (`"QmAAA...AAA"`)
- it may still have a prefix and be a merkle link (`"/ipfs/QmAAA...AAA/foo/bar"`)
- it may be another type of link ("/ipns/foo.com/a/b/c"`) but is not counted as immutable

jbenet: We _could_ use `href` instead of `link` and be retro + HTML friendly.

jbenet:
@diasdavid 

> I don't see any specific mention to the data types the links point to, have we decided that isn't IPLD turf (and just json-ld)?

I think there is definitely a strong case for `@type` to be defined here-- as something different from JSON-LD's. I don't know fully what it would do just yet, beyond knowing that i would want to have some behavior like `JSON-LD/@context's`, where identifiers can be aliased + shortened, and where defaults for values could be defined. Also definitely useful to have as an identifier for applications to act upon (eg, unixfs/file, unixfs/dir, commit, etc).

jbenet:
@diasdavid 

> Also, instead of adding more TODO here, can we before go issue by issue here https://github.com/ipfs/go-ipld/issues and diasdavid/js-ipld#2 and check what is already included in the spec, add what is missing, explicitly add what was accepted and then see what's left, bringing all of that conversation to this PR? So that we all talk over the same page :)

Agreed, let's do that.

merkledag/ipld.md
jbenet: ah thanks. it should actually be:

```
{
  "name": "Vannevar Bush"
}
```
addressed in 37c662a

merkledag/ipld.md
jbenet: Yep, thanks. addressed in 37c662a

merkledag/ipld.md
jbenet: thanks addressed in 37c662a

merkledag/ipld.md
jbenet: sounds good. addressed in 37c662a

merkledag/ipld.md
jbenet: good catch! addressed in: 37c662a

jbenet:
### Accessing link properties problem (important issue to resolve)

Path resolution for _data within links_ is currently not possible. specifically, i mean:

Suppose we have object which hashes to `QmCCC...000`:

```yml
# 
---
stuff:
  foo:
    mlink: QmCCC...111
    mode: 0755
    owner: jbenet
```
and we have object which hashes to `QmCCC...111` (the foo link):

```yml
---
other:
  cat.jpg:
    mlink: QmCCC...222
    mode: 0644
    owner: jbenet
```

We can traverse so:
```
> ipld cat --yml QmCCC...000/stuff
foo:
  mlink: QmCCC...111
  mode: 0755
  owner: jbenet

> ipld cat --yml QmCCC...000/stuff/foo/other
---
cat.jpg:
  mlink: QmCCC...222
  mode: 0644
  owner: jbenet
```
But we **cannot** access the data in the links themselves, we cannot get the values of `mode` or `owner` properties directly.

### Possible Solutions

#### keep as is, no direct access to link properties

Ok, but possibly a cop-out.

#### introduce another separator to traversal within single objects

suppose we use "." to separate levels within a single object, as in js dot notation. or some other character (`|`).

```
> ipld cat --yml QmCCC...000/stuff.foo.mode
0755
> ipld cat --yml QmCCC...000/stuff/foo
error: no link named "stuff" in object QmCCC...000
```

#### introduce another separator, to use _ALSO_ within single objects

suppose we use "." to separate levels within a single object, as in js dot notation. or some other character (`|`). this separator would always be to disambiguate, and not _replace_ the use of `/` within objects. It would also need to be escaped if found normally.

```
> ipld cat --yml QmCCC...000/stuff/foo.mode
0755
> ipld cat --yml QmCCC...000/stuff.foo.mode
0755
> ipld cat --yml QmCCC...000/stuff/foo/mode
error: no key "mode" in object QmCCC...000/stuff/foo/
```

#### Use a special key to access link properties specifically (not traverse all object parts)

suppose we use `.link`

```
> ipld cat --yml QmCCC...000/stuff/foo.link/mode
0755
```

merkledag/ipld.md
mildred: I'd say, don't resolve links for keys that have `/` in them. We can still store data structure which have those keys, we just can't resolve them through paths. I don't see a problem in that. We would have a separate API to parse the local data structure without following links.

merkledag/ipld.md
mildred: > * IPLD paths MUST layer cleanly over UNIX and The Web (use `/`, have deterministic transforms for paths not having specified character encoding).

mildred:
> >  In that case, we should either provision key escaping to avoid clashing with someone wanting to have "@multicodec" as an arbitrary key in his JSON data structure
>
> Yep, agreed. what's a good place to put that?

I would say, here in this spec. Something like:

> Some keys are reserved in the maps of the datastructure for internal IPLD use. Every key starting with `@` is considered an IPLD directive. Unknown directives should be ignored. Legitimate keys containing the `@` character should be escaped using `\@` instead of `@` and `\\` instead of `\`.
>
> When converting from JSON to an IPLD document, the keys must be escaped appropriately. When decoding an IPLD document back to JSON, the keys must be decoded back. When manipulating an IPLD document, the IPLD directives must be evaluated. The system may also allow to represent an IPLD document with its directives, but it should be clear that the document must follow specific IPLD notation (special directives and escaping of keys)
>
> IPLD directives:
>
> - `@codec`: at the root of the IPLD document, specifies a codec in which the IPLD document was encoded.
> - ...

mildred:
> remaining question: should the text in paths be the raw case, or unescaped? eg, should this be possible or inaccessible?
>
>> .../foo/bar/@multicodec

I say it shouldn't be possible. I see no use for that, and it will prevent paths like `/ipfs/HASH/dir/@multicodec/filename`. The whole point of escaping is to allow that.

mildred:
> ### Accessing link properties problem (important issue to resolve)

I say, take the same approach as unix symbolic links:

* Have the ability to specify when resolving the link if we want to follow the link or not. POSIX has this with `stat()` vs `lstat()` for instance.

        > ipld cat --no-resolve --yml QmCCC...000/stuff/foo/
        ---
        mlink: QmCCC...111
        mode: 0755
        owner: jbenet

        > ipld cat --resolve --yml QmCCC...000/stuff/foo/
        ------
        other:
          cat.jpg:
            mlink: QmCCC...222
            mode: 0644
            owner: jbenet

* Or, change the behaviour depending on if the trailing `/` is present.

        > ipld cat --yml QmCCC...000/stuff/foo
        ---
        mlink: QmCCC...111
        mode: 0755
        owner: jbenet

        > ipld cat --yml QmCCC...000/stuff/foo/
        ------
        other:
          cat.jpg:
            mlink: QmCCC...222
            mode: 0644
            owner: jbenet

    You have the same on unix as well:

        > ln -s / foo
        > stat ./foo
          File: ‘foo’ -> ‘/’
          Size: 1         	Blocks: 0          IO Block: 4096   symbolic link
        > stat ./foo/
          Size: 4096      	Blocks: 8          IO Block: 4096   directory
        Device: fe02h/65026d	Inode: 2           Links: 25

    The problem with this solution is that it makes easier to access the link properties than the link content in itself. Generally we want the opposite (I would prefer to `cat .../toto.txt` instead of `cat .../toto.txt/`). It can be solved by:

* Access link properties when the separator is `//` instead of `/`.

        > ipld cat --yml QmCCC...000/stuff/foo//
        ---
        mlink: QmCCC...111
        mode: 0755
        owner: jbenet

        > ipld cat --yml QmCCC...000/stuff/foo
        ------
        other:
          cat.jpg:
            mlink: QmCCC...222
            mode: 0644
            owner: jbenet

    This is close to your *introduce another separator to traversal within single objects* solution but does not introduces a new separator. This makes it incompatible with unix paths in a slight way because on unix `foo/bar` is strictly equivalent to `foo//bar` or `foo///bar`. This won't be the case here. This also means that fuse interfaces and unix applications might simplify `//` to `/` and this feature may not be available there.

As for the other solutions you proposed:

- **keep as is, no direct access to link properties:** this is nice because it makes it compatible with everything we have elsewhere (unix paths, URLs, ...) for free

- **introduce another separator to traversal within single objects:** Imagine the problem if I `touch dir/foo.jpeg && rm -f dir/foo.jpeg.mode`. What do I get ?

    We can also forbid ever using the `.` or `|` character (or the special selected character) in paths. We loose compatibility with unix paths.

- **introduce another separator, to use ALSO within single objects:** same compatibility/escaping problem here.

- **Use a special key to access link properties specifically (not traverse all object parts):** I prefer this one. It is conceptually closer to what we understand how a filesystem is. This has the same problem of clashing with legitimate file names: `touch toto && rm -f toto.link/mode` or `rm -f test && mkdir test.link && cat test`.

merkledag/ipld.md
```diff
@@ -0,0 +1,462 @@
...
+
+On the subject of integers, there exist a variety of formats which represent integers as strings in JSON, for example, [EJSON](https://www.meteor.com/ejson). These can be used and conversion to and from other formats should happen naturally-- that is, when converting JSON to CBOR, an EJSON integer should be transformed naturally to a proper CBOR integer, instead of representing it as a map with string values.
+
+
+## Serialized Data Formats
+
+IPLD supports a variety of serialized data formats through [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `mlink`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
```
davidar: :+1:

jbenet:
> Have the ability to specify when resolving the link if we want to follow the link or not. POSIX has this with stat() vs lstat() for instance.

yep, could do this. would love to be able to do it directly in the path expression though.

> Or, change the behaviour depending on if the trailing / is present.

this doesn't allow grabbing links themselves.

> Access link properties when the separator is // instead of /.

i had thought of this as the reverse, ie. _`//` is the resolve through operator_ -- but yes, problems in unix + web both. may be too subtle for users.

jbenet:
Ok, update here: 

- we really need the "Accessing link properties problem (important issue to resolve)" thing solved. i still haven't reached a solid perspective. its important to solve now as doing it right may involve changing the format.
- people have brought up that we should use `@link` or just `link` instead of `mlink`, since `/ipfs/...` indicates a merkle link, and we may want other kinds of links. now is the last time we can change this without much pain, so we should decide that asap.

I **think** these are the last two decisions before we can push this into go-ipfs.

merkledag/ipld.md
davidar: Why not escape slashes, so:

```
> ipld cat --fmt yml $h1/foo/bar
---
hello: h2bar1

> ipld cat --fmt yml $h1/foo\/bar
---
hello: h3bar2
```

merkledag/ipld.md
davidar: This is why I like the data-constructor notation better

```yaml
title: As We May Think
author: !mlink QmAAA...AAA
```

It makes it clear that we're *constructing* a link with the given hash as the argument to the constructor. The magic "mlink" key pointing to a random string is not so clear.

I know that JSON doesn't have that luxury, so it'll have to go with magic keys, but I think it's worth making the distinction at the protocol level.

merkledag/ipld.md
davidar: > that will blow up verbosity and storage size.

If we're storing data as CBOR, and encoding mlink as a tag, that shouldn't be an issue? It would increase verbosity when displaying as JSON though, yes.

davidar:
>  Or we should transmit this information outside of the IPLD object itself (not encoded as part of the datastructure).

:+1: for out-of-band signalling

> Path resolution for data within links is currently not possible.

Using a single character (especially a common one like `.`) means more escaping. XPath does something like

```
> ipld cat --yml QmCCC...000/stuff/foo/@mode
0755
```

which uses only the existing special characters, but I'm not sure if that would conflict with the other `@directive` stuff? The less concise XPath version would be

```
> ipld cat --yml QmCCC...000/stuff/foo/attribute::mode
0755
```

> people have brought up that we should use @link or just link instead of mlink, since /ipfs/... indicates a merkle link, and we may want other kinds of links. now is the last time we can change this without much pain, so we should decide that asap.

If we use CBOR tags for links on the wire, it shouldn't matter too much what we name them (i.e. it's an API-level issue rather than a protocol-level one)?

mildred:
> Given that we're already using @ as a special character, why not do something like
>
>     > ipld cat --yml QmCCC...000/stuff/foo@mode
>     0755
>
>Keys containing a literal @ would be escaped as usual
>
>     > ipld cat --yml QmCCC...000/stuff/foo\@mode
>     Some content

We can't, because the whole point of escaping the `@` character in JSON keys is to allow the character to appear in paths. The actual directive (with `@` unescaped) is not accessible directly using the path notation.

The whole point is to allow using IPFS as a regular unix filesystem. Say I want to mount ipfs somewhere on my system, and I them `touch  toto@mode` somewhere. I can also `touch 'toto\@mode'` because on unix, I can have `\` in filenames, try it.

Perhaps we need two different kind of paths. Paths meant for the filesystem or the web that allow any character to be present in the filesystem. And paths meant for IPLD -aware users that allow access to more things, but which won't be compatible with usage in a unix filesystem. It has to be clearly specified so people do not mistake one for the other.

    > ipld cat --yml --ipld-path=QmCCC...000/stuff/foo@mode
    0755

    > ipld cat --yml --file-path=QmCCC...000/stuff/foo@mode
    Some content

:trollface: we could even prefix these paths with either `file:` for the unixfs compatible paths, or with `ipld:` so we can't mistake one for the other.

    > ipld cat --yml ipld:/ipfsQmCCC...000/stuff/foo@mode
    0755

    > ipld cat --yml file:/ipfs/QmCCC...000/stuff/foo@mode
    Some content

</ :trollface: >

mildred:
**Accessing link properties problem (important issue to resolve)**

I see few solutions, none is perfect. Could we decide which one we want ?

- Use a special character(s) in paths to distinguish between following links and not following links. This break unix compatibility because this special character is no longer usable for regular files, or only in escaped form, which break compatibility anyway.

    - `QmCCC...000/stuff.foo.mode`
    - `QmCCC...000/stuff/foo.link/mode`
    - `QmCCC...000/stuff/foo@mode`

  Note: we can use a character sequence that is most likely not found anywhere else so we can avoid *most* of the conflicts (not all unfortunately). For example by replacing `.link` with something else we could have `QmCCC...000/stuff/foo@ipld-link-properties/mode`. Let's hope nobody wants to create a file containing `@ipld-link-properties`.

- Use the `//` character sequence as special character in the above solution. This is a minor breaking of unix compatibility which states that `foo//bar` == `foo/bar`. This might be mitigated by the fact that  some API (fuse?) might sanitize such path before passing them to IPFS.

    - `QmCCC...000/stuff/foo//mode`

- Do not allow path representation of link properties. Only allow access using a specific API (think `lstat()` vs `stat()`)

- Any combination of the above: depending on the context or the API function we use, we can choose a path format or another. Problem: how do we know which path representation to use? Paths are no longer self describing. (hence prefixing it with `file:` or `ipld:` :trollface:)

Basically, it all comes down to the fact that if the extra keys are exposed to the fuse filesystem, you must use some notation that is not fully transparent. You don't have a generic filesystem anymore, you have a filesystem with extra features.

merkledag/ipld.md
mildred: I suppose that you meant to put the path in simple quotes, else your bourne compatible shell will replace `\/` by `/` and both commands are identical.

    > ipld cat --fmt yml '$h1/foo\/bar'
    ---
    hello: h3bar2

The kernel doesn't know escaping so what the linux kernel will understand when presented with the path `foo\/bar` is the entry `bar` enclosed in a directory called `foo\`. Backslashes in file names are valid and this notation would prevent using them.

mildred:
> If we use CBOR tags for links on the wire, it shouldn't matter too much what we name them (i.e. it's an API-level issue rather than a protocol-level one)?

Using CBOR tags for the wire format would be nice (@jbenet, an opinion?).

I think the solution to use CBOR tags would be to transform every `mlink` property containing a valid multihash (and only those) to a CBOR binary property containing a link tag. This should be done in all cases to have a unique representation (and a unique hash for a given data).

Note: if the transformation from ascii multihash to binary is not a [bijection](https://en.wikipedia.org/wiki/Bijection) function, we have a problem.

mildred:
> Given that we're already using @ as a special character

Note that the spec in this PR doesn't talk about this specifically. For now, I assume there is no special characters in JSON and no escaping needed in JSON. I prefer it this way and gladly remove every escaping function in go-ipld.

merkledag/ipld.md
davidar: Sure, you wouldn't be able to access keys like that with the fuse mount, but I don't see why that doesn't mean the ipld command couldn't support escaping? If you wanted a literal backslash, you'd write `\\`

davidar:
> Perhaps we need two different kind of paths. Paths meant for the filesystem or the web that allow any character to be present in the filesystem. And paths meant for IPLD -aware users that allow access to more things, but which won't be compatible with usage in a unix filesystem. It has to be clearly specified so people do not mistake one for the other.

Sounds reasonable to me.

> Use the // character sequence as special character in the above solution. This is a minor breaking of unix compatibility which states that foo//bar == foo/bar.

I feel like that's just asking for trouble. Someone uses "$DIR/foo.txt", and if $DIR has a trailing slash, the path suddenly has a completely different meaning.

> Note: if the transformation from ascii multihash to binary is not a bijection function, we have a problem.

Given that the canonical ASCII representation of a multihash is just base58 encoded binary data, I think that should be fine.

> Note that the spec in this PR doesn't talk about this specifically. For now, I assume there is no special characters in JSON and no escaping needed in JSON. I prefer it this way and gladly remove every escaping function in go-ipld.

I'd still like some way of specifying datatypes in IPLD, and it seems like there will need to be escaping of some description in JSON to do that. @jbenet ?

mildred:
> I feel like that's just asking for trouble. Someone uses "$DIR/foo.txt", and if $DIR has a trailing slash, the path suddenly has a completely different meaning.

That's exactly why POSIX says `foo//bar` == `foo/bar`. If we are clear about compatibility with unix filesystem, I'd also prefer another solution.

> I'd still like some way of specifying datatypes in IPLD, and it seems like there will need to be escaping of some description in JSON to do that. @jbenet ?

Just like [encoding](https://github.com/ipfs/specs/pull/37#issuecomment-162139794) ([and this](https://github.com/ipfs/specs/pull/37#issuecomment-154996715)) could be and out-of-band parameter, the type could just be the same. It's similar to HTTP's Content-Type header anyway.

Or if we prefer self describing objects, we'll need to specify that escape mechanism, and how that translates to CBOR tags. And we include in that the object type and its codec (makes sense to include it as well).

almereyda:
Just to note: RDF *is* no serialization, but it *has* serializations.

merkledag/ipld.md
diasdavid: got it, my initial thought was that merkle links can be resolved and something like a smart bitswap can traverse

parkan:
Are we at a point where I can start writing out more or less IPLD-compatible objects if I don't have ambiguous links? Even something like raw JSON-LD inside current protobuf format objects that can be later promoted to the new format? It's a bit hard to parse out where things are at between this PR and the  open issues on go-ipld

mildred:
I just noticed we left out an important part of the spec aside: the relationship with protocol buffer serialization. I described it in PR #59. Compatibility with protocol buffer is tricky enough that it can lead to modifications in other parts of the spec (like IPLD paths).

Here is what I added, in full Perhaps, it would be easier to discuss about it in PR #59 to avoid cluttering this thread.

## Relationship with Protocol Buffers legacy IPFS node format

IPLD has a known conversion with the legacy Protocol Buffers format. This format is defined with the Protocol Buffers syntax as:

    message PBLink {
        optional bytes  Hash = 1;
        optional string Name = 2;
        optional uint64 Tsize = 3;
    }
    
    message PBNode {
        repeated PBLink Links = 2;
        optional bytes  Data = 1;
    }

The conversion to the IPLD data model must have the following properties:

- It should be convertible back to protocol buffers, resulting in an identical byte stream (so the hash corresponds). This implies that ordering and duplicate links must be preserved in some way.
- When using paths as defined earlier in this document, links should be accessible without further indirection. This requires the top node object to have keys corresponding to link names.
- Link names should not conflict with other keys.

There are multiple ways to do that that will be described next.

### Current encoding in go-ipld

go-ipld implements the following conversion:

    {
      "<Links[0].Name.(escaped)>": {
        "hash": "<Links[0].Hash>",
        "name": "<Links[0].Name>",
        "size": <Links[0].Tsize>
      },
      "<Links[1].Name.(escaped)>": {
        "hash": "<Links[1].Hash>",
        "name": "<Links[1].Name>",
        "size": <Links[1].Tsize>
      },
      ...
      "@attrs": {
        "data": "<Data>",
        "links": [
          {
            "hash": "<Links[0].Hash>",
            "name": "<Links[0].Name>",
            "size": <Links[0].Tsize>
          },
          {
            "hash": "<Links[1].Hash>",
            "name": "<Links[1].Name>",
            "size": <Links[1].Tsize>
          }
        ]
      }
    }

Notes :

- The `links` array in the `@attrs` section is there to preserve order and duplicate links to hash back to the exact same protocol buffer object.

- The link names are escaped to prevent clashing with the `@attr` key.

- The escaping mechanism transforms the `@` character into `\@`. This mechanism also implies a modification of the path algorithm. When a path component contains the `@` character, it must be escaped to look it up in the IPLD Node object.

    For example, a path a path `/root/first@component/second@component/third_component` would look for object `root["first\@component"]["second\@component"]["third_component"]` (following mlinks when necessary).

- Links are represented using the `hash` key instead of `mlink` as used in this specification. This must be changed.

### Other proposition that does away with escaping

We can imagine another transformation where the link names are not escaped. For example:

    {
      "<Links[0].Name>": {
        "mlink": "<Links[0].Hash>",
        "tsize": <Links[0].Tsize>
      },
      "<Links[1].Name>": {
        "mlink": "<Links[2].Hash>",
        "tsize": <Links[2].Tsize>
      },
      ...
      ".": {
        "data": "<Data>",
        "links": [
          "<Links[0].Name>",
          {
            "name": "<Link[1].Name>",
            "mlink": "<Links[1].Hash>",
            "tsize": <Links[1].Tsize>
          }
          "<Links[2].Name>",
          ...
        ]
      }
    }

Notes:

- Very conveniently, we use the key `.` to represent data for the current node, and any other key can represent a link. This means that we forbid link to be named `.`. This is in any case a good thing to do as the `.` element in paths can always be removed (the same way `..` can be replaced by the parent directory)

- No escaping is needed, and no modification to the path algorithm is needed.

- Link order is kept by using the link names for links present in the top node. This avoid repeating identical data (even though it is probably generated on the fly).

    Links that cannot be present in the top node (the case for the link named `.`, which is forbidden, or for links that are repeated with the same name) are present in full to allow reconstructing the exact protocol buffer object.

### Other encodings

Speak up in comments if you have other encoding suggestions, or you have another implementation with another encoding.

mildred:
See my PR #60 which address the "Accessing link properties problem".

> people have brought up that we should use `@link` or just `link` instead of `mlink`

Using `@link` makes no sense if we don't escape elements in IPLD (which I would prefer, as it is simpler). `mlink` or `merkle-link` seems preferable to me than just `link` as it is more explicit.

If we use `link` only, I would be in facor of prefixing the hash with `/ipfs/` on the associated value.

Also, if we want to encode links in a special way in the CBOR format, this should be included in the spec. Does anyone that understands CBOR wants to do that?

mildred:
Pending issues:

- [ ] Protocol Buffer conversion (see #59)
- [ ] Accessing link properties problem (see #60)
- [ ] [people have brought up that we should use `@link` or just `link` instead of `mlink`, since /ipfs/... indicates a merkle link, and we may want other kinds of links. now is the last time we can change this without much pain, so we should decide that asap.](/ipfs/specs/pull/37#issuecomment-161863743)
- [ ] Encode merkle-links in a special way in CBOR format

jbenet:
> If we use `link` only, I would be in facor of prefixing the hash with `/ipfs/` on the associated value.

Agreed. we could have a "default link type" which is an `/ipfs/<link>` the same way the ipfs cli works, to avoid unnecessary bloat in objects (the `/ipfs/`s will add up). or we may investigate using cbor to define some link types as tags. im not sure how this would be done. @davidar have you delved into the cbor spec enough to know?

jbenet:
> Also, if we want to encode links in a special way in the CBOR format, this should be included in the spec. Does anyone that understands CBOR wants to do that?

yes. the more compressed we can make the representation, the better.

jbenet:
@parkan 

> Are we at a point where I can start writing out more or less IPLD-compatible objects if I don't have ambiguous links?

If you need to get rolling with IPLD, what we could do is this:

- take the latest state of this spec and make sure go-ipld matches it.
- create a ipld branch in go-ipfs for the merge
- have [gobuilder](https://gobuilder.me/github.com/ipfs/go-ipfs/cmd/ipfs) build that branch for people wanting to try it.
- we'll still have migration and interop stuff to work on, but will be useful to those wanting the new format without having to wait for it to merge into master.
- we need to do this anyway to merge go-ipld :) 

> Even something like raw JSON-LD inside current protobuf format objects that can be later promoted to the new format? 

Yes you can definitely use JSON (and JSON-LD) inside the data segment of today's protobuf stuff. it will likely be a bit annoying, but can do it fine. you can see how unix-fs in go-ipfs does things: https://github.com/ipfs/go-ipfs/tree/master/unixfs this is "unixfs protobuf" inside "ipfs object protobuf". and how https://github.com/ipfs/ipfs-geoip works (it uses "geoip btree json" inside "ipfs object protobuf"). the thing to remember is to make sure the protobuf links properly link to the child objects (careful wrapping).

> It's a bit hard to parse out where things are at between this PR and the open issues on go-ipld

This spec surfaced some issues in the pathing + conversion that we're working out. the expressibility of this pathing is really, really nice, so making sure it's right ill make us all very happy in the long term. (it's one of those "small but important" things, like "comments in json")

mildred:
In PR #59 we are deciding which representation a protocol buffer object should have in the IPLD object model. Please give your opinion on which conversion you prefer (or if you prefer another conversion entirely, please describe it)

mildred:
> Agreed. we could have a "default link type" which is an /ipfs/<link> the same way the ipfs cli works, to avoid unnecessary bloat in objects (the /ipfs/s will add up). or we may investigate using cbor to define some link types as tags. im not sure how this would be done. @davidar have you delved into the cbor spec enough to know?

Are you saying that ... ?

- we should use the `link` key in IPLD object instead of `mlink`
- if the corresponding value corresponds to a base58 hash, we should add `/ipfd/` at the beginning and follow the link
- else we follow the link which can be anything else (`/ipns/...` for example)

If so, do we take that decision and replace all occurrences of `mlink` by `link` ?

mildred:
I added a PR (#61) to describe a possible way to add tags to CBOR encoding. The pending issue list is now:

- [ ] Protocol Buffer conversion (see #59)
- [ ] Accessing link properties problem (see #60)
- [ ] Decide if we should replace `mlink` with `link`.
- [ ] Encode merkle-links in a special way in CBOR format (see #61)

mildred:
New TODO item:

- [ ] Know how to deal with IPLD objects that have different encoding when we need to compute the hash (PR  #63)

mildred:
#59 (protobuf conversion to ipld), #61 (CBOR tagging), #62 (merkle-path modifications to access link properties and #64 (rename mlink to link) are all in a coherent state. We no longer need key escaping. Do all parties agree on these? Please take a look.

jbenet:
For those following along, see https://github.com/ipfs/specs/pull/62#issuecomment-173820870 

> @mildred I've made https://github.com/ipfs/ipld-examples to try and resolve this question. So far i've only made two examples `unixfs` and `post`. take a look, i changed things up somewhat after giving more thought. I'm still not decided, but i'm seeing how horrible some of those options are.
>
> could you please double check my work? im not sure i got everything right-- I may have messed up some of them, given that there are things you mentioned were problems but i didnt run head into them. (escaping `.` when using it as a delimiter, for example. may be that this _could_ be a non issue, but anyway, its separate).
> 
> I think so far, my favorites are (8), (4), and (5). -- the other feel odd or are very, very confusing.
>
> Also, see the `unixfs` pathing-- i made it a separate thing there too, and i am more convinced this is the right thing to do.

monokles:
Just a quick heads up: The document does not currently mention what the abbreviation IPLD stands for.

jbenet:
Thanks! Will add it in.

@mildred what do you think of pathing option 8?

mildred:
I don't have much opinions on all these solutions. Al long as it is possible to address everything, I don't have any preferences. (1) is good but does not address every data. (2) is equivalent of (1) so I would not consider it differently. (3) is confusing. (4) is good and (5) is only a variation of it, probably more confusing. (6) is fine.

I see an issue with solutions (7) and (8). What if we want to address a data structure with a key named "`link`"? It is possible, but confusing. See https://github.com/mildred/ipld-examples/blob/d6dadbd4d73aa57c3fe29c2093570d42bb5680e1/unixfs/example.yml and the end of https://github.com/mildred/ipld-examples/blob/d6dadbd4d73aa57c3fe29c2093570d42bb5680e1/post/paths-7.md

mildred:
Perhaps we can work in steps:

- try to finish the issues where there is not much debate: #59 (legacy protocol buffer serialisation) and #64 (rename `mlink` to `@link`) and merge them here
- try to finish up CBOR tagging quickly #61
- try to find a subset of IPLD that can be merged in master now
- leave the discussion about the path #62 in a separate issue to be resolved before we ship IPLD for real. Core IPLD can work without any path mechanism (I didn't implement one in go-ipld and go-ipfs/merkledag don't have a path mechanism) and implementation can benefit from having a solid spec for core features.
- don't shy away from opening other issues if some are raised during IPLD implementation.

This way, we can perhaps avoid having a gigantic issue where everything in IPLD is discussed at once.

mildred:
#59 #61 and #64 are LGTM. Waiting for other's opinions. #62 still needs more thoughts, but I don't really mind about the path format if we agree that it won't be used for unixfs if it is incompatible with most unix filenames.

jbenet:
@mildred thanks for pushing on this. agreed to proceeding on the other fronts will review shortly.

jbenet:
> I see an issue with solutions (7) and (8). What if we want to address a data structure with a key named "link"? It is possible, but confusing.

well, the key would be taken, the benefit of `@link` is that we dont _have to_ escape it (i think we've discussed this version before). i.e. if the user needs `@link` exactly for something else, user can escape it when constructing / importing to IPLD. (`@link` is very unlikely to clash with anything else, and importers libs could well escape for the user, but the core implementation doesnt have to)

---

@mildred 
- i think i like solution (8) the most, with `/` and `//` and `/@link/` -- the reason i want (8) over (4) is that i do want ipld paths to be expressible in a unix path, web URI/URL, and having `/@link/` make it possible to do. I would prefer `//` the rest of the time.
- i like solution (4) with `.` and `/` second. 
- I like solution (1) as the last option

mildred:
> i think i like solution (8) the most, with / and // and /@link/ -- the reason i want (8) over (4) is that i do want ipld paths to be expressible in a unix path, web URI/URL, and having /@link/ make it possible to do. I would prefer // the rest of the time.

I like these options as well. If the filesystem layer contracts `//` to `/` it is always possible to use `/@link/`, and this can be nicely represented on a filesystem.

jbenet:
OK! #59 #61 #62 #64 are all merged! :+1: MASSIVE thanks to @mildred for pushing it through. We all are very thankful :) 

What issues remain here? I think I will merge this (FINALLY!) and let's continue to iron it out with future PRs against master. I think we have a solid spec, and now https://github.com/ipfs/go-ipld/ and https://github.com/diasdavid/js-ipld/ can match it.

jbenet:
Note: let's keep the branch, as there's many links to that specifically.

noffle:
Great work, @mildred, @jbenet! :clap:

parkan:
:fireworks:

PR #40, opened 2015-11-16
fix examples in ipld spec/'Linking Between Nodes'

candeira:
Either I misunderstood the point of named Unix paths, or the examples were wrong.


jbenet:
Ah!! thanks! sorry didnt see this PR till now. -- i think i just fixed it. lmk if not.

PR #43, opened 2015-11-23
IPFS Repo Spec update

jbenet:
The repo spec has some discrepancies to fix. I've added some commits that fix some, and noted others. This PR is a discussion to fix it.

Thanks to @masylum for prompting this fix, and sorry you had to find it by hand.

Discrepancies to address:
- [ ] fs-repo: version format
- [ ] fs-repo: pid in repo.lock
- [ ] fs-repo: `ipfs repo fsck` to fix lock issues.
- [ ] fs-repo: blocks dir layout (fanout/depth)
- [ ] fs-repo: come up with a name for api-file-only repo

cc @ion1 @whyrusleeping @diasdavid

### DIFF
--- a/repo/README.md
+++ b/repo/README.md
@@ -17,7 +17,6 @@ A `repo` is the storage repository of an IPFS node. It is the subsystem that
 actually stores the data ipfs nodes use. All IPFS objects are stored in
 in a repo (similar to git).
 
-
 There are many possible repo implementations, depending on the storage media
 used. Most commonly, ipfs nodes use an [fs-repo](fs-repo).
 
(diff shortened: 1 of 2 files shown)


repo/README.md
```diff
@@ -86,26 +74,53 @@ feature swappable datastores, for example:
...
+
+### config (state)
+
+The node's `config` (configuration) is a tree of variables, used to configure various aspects of operation. For example:
+- the set of bootstrap peers IPFS uses to connect to the network
+- the Swarm, API, and Gateway network listen addresses
+
+It is recommended that `config` files avoid identifying information, so that they may be re-shared across multiple nodes.
+
+**CHANGES**: today, implementations like go-ipfs store the peer-id and private key directly in the config. These will be removed and moved out.
```
masylum: Would be good to specify the json schema? Ideally any implementation should be compatible.

repo/README.md
```diff
@@ -86,26 +74,53 @@ feature swappable datastores, for example:
...
+
 ### logs
 
-A full IPFS node is complex. Many events can happen, and thus ipfs
+A full IPFS node is complex. Many events can happen, and thus some ipfs
 implementations capture event logs and (optionally) store them for user review
 or debugging.
 
+Logs MAY be stored directly as IPLD objects along with everything else, but this may be a problem if the logs
```
masylum: we will never know how this sentence ends :) Suspense !

diasdavid: choose your own adventure! :D

masylum:
Good to see specs moving forward little by little.

My two cents: I see some overlap between `fs-repo` and the `repo` specs. For instance, I'm not sure `repo.lock` is something a non-fs repo should be worried about. I'm thinking, for instance of an `IndexedDB/Localstorage ` implementation for the browser, it probably does not care about locks at that level.This week I will try to move forward with the js implementation. Those differences will become more obvious when trying to get a non fs based repository.

jbenet:
>  I'm not sure repo.lock is something a non-fs repo should be worried about. I'm thinking, for instance of an IndexedDB/Localstorage implementation for the browser, it probably does not care about locks at that level.

makes sense. I only included that in `repo` spec to make it clear people need to deal with concurrent access. we could just make it a clear warning and remove the `repo.lock` requirement there altogether.

any other observations like that?

> Those differences will become more obvious when trying to get a non fs based repository.

sounds great!

thanks for the feedback-- a solid spec here will go a long way

repo/README.md
jbenet: that sounds good-- though i do want to keep config files flexible, in that people may want to add more values into it than we give -- similar to how people use `.git/config`

repo/README.md
jbenet: it was never logged! :astonished:

repo/README.md
```diff
@@ -32,46 +31,35 @@ Repo Implementations:
```
diasdavid: It would be pretty sweet to upgrade https://github.com/maxogden/abstract-blob-store to have a parallel interface for Go, so that we can have all of those s3/fs/mem repos for 'free'.

jbenet: In go, we use https://github.com/jbenet/go-datastore/ -- which is similar and alraedy has a bunch. 

i've been making these abstractions for a long time: https://github.com/datastore/datastore

PR #47, opened 2015-11-25
bubble up and add logo

diasdavid:
bring it up, so it is not hidden behind protocol (easier to find)


diasdavid:
Going to move ahead and merge this (as it is a very simple change)

jbenet:
lgtm :+1:

PR #49, opened 2015-11-25
libp2p - Update Implementations Chapter

diasdavid:


libp2p/8-implementations.md
```diff
@@ -1,40 +1,57 @@
 8 Implementations
 =================
 
-This is a list of known libp2p module implementations. They are components that respect the interfaces and expectations defined in the "Interfaces" chapter, and can be composed to make a working libp2p library.
+An libp2p implementation should(recommended) follow a certain level of granulatiry when implementing different modules and functionalities, so that common interfaces are easy to expose, test and check for interoperability with other implementations.
+
+This is the list of current modules available for libp2p:
+
+- libp2p (entry point)
```
Dignifiedquire: It would be nice if each of these parts had a list of implementations in the different languages attached to it, like this:

```markdown
- abstract-transport [spec](link/to/spec) [go](link/to/go) [JS](link/to/js)
```

jbenet: if we're going to use the "abstract-X" modules for more than just javascript, can we:
- rename the convention to something like "X-interface" or "interface-X" or "libp2p-X" or "X-protocol" or "X-proto"
- use a programming language with _types_ to define the interfaces (Go or Haskell ok for me)

---

- bikeshed discussion: "interface" is a well-understood concept across programming languages. the use of "abstract-X" in node/js makes up for javascript's lack of a type system, but will not be understood as readily by people in other languages. the right word for this in PL vocabulary is "protocol" or "interface". "protocol" is too general of a word and has so many meanings-- "interface" tends to be preferred these days. 
- bikeshed link: https://en.wikipedia.org/wiki/Protocol_(object-oriented_programming) 
- bikeshed trivia: "concepts" and "abstract base class" are two holdovers from C++ and friends (probably in typical disregard of "the type people")

libp2p/8-implementations.md
```diff
@@ -1,40 +1,57 @@
...
+  - libp2p-kad-record-store
+- **Generic**
+  - PeerInfo
+  - PeerId
+  - multihash
+  - multiaddr
+  - multistream
+  - multicodec
+  - ipld
+  - repo
```
jbenet: :heart: love this list

libp2p/8-implementations.md
```diff
@@ -1,40 +1,57 @@
...
+  - libp2p-random-walk
+  - libp2p-railing
+- **Distributed Record Store**
+  - libp2p-record
+  - [abstract-record-store](https://github.com/diasdavid/abstract-record-store)
+  - libp2p-distributed-record-store
+  - libp2p-kad-record-store
+- **Generic**
+  - PeerInfo
+  - PeerId
```
jbenet: maybe these could be part of a `peer` package?

jbenet:
LGTM

libp2p/8-implementations.md
Dignifiedquire: :+1: for defining types in Haskell from my side

---

bikeshed: :+1: for `interface`, :-1: for `protocol` (too generic, too much overlap with network protocols)

libp2p/8-implementations.md
diasdavid: What is `Peer` for you, I believe `PeerInfo` === `Peer`

libp2p/8-implementations.md
diasdavid: I remember having this conversation and agreeing with you, should have added it as a TODO right away.

I like `interface-X`.

:+1: for Haskell too, however, we will have a mapping for each language as implementations appear :)

davidar: :+1: Haskell :)

diasdavid:
interfaces updated, merging :)

PR #50, opened 2015-11-26
libp2p - add more data structures, so they can be used in other parts of the spec, referencing to this section

diasdavid:


Dignifiedquire:
More docs and understandable with links, so :+1: from me

jbenet:
LGTM

PR #51, opened 2015-12-11
Added the API Specs from ipfs/api

RichardLitt:
This is in reference to ipfs/api#11. Thanks @diasdavid for the suggestion.


jbenet:
good start.

jbenet:
I have time to work on this next week, and following. lt's set a goal:

- Completed IPFS API Spec (core + transports) by Jan 1

diasdavid:
thank you :):)

PR #53, opened 2015-12-19
WIP: exchange spec

diasdavid:


exchange/README.md
```diff
@@ -0,0 +1,108 @@
...
+##### auction house
+
+`auction-house` keeps track of what blocks my peer wants, what others peer wants list are, peers bandwith, reputation and other important metricts that will enable bitswap to become smarter.
+
+### interface
+
+`fn get(Multihash: key) -> (Vec<u8>: block, Err: err)`
+`fn put(Multihash: key, Vec<u8>: block) -> Err: err`
+
+Q: bitswap should not expect an already serialized block so that we can pick the serialization format inside when passing it around.
```
diasdavid: @whyrusleeping I'm trying to come with a best way to enable cbor and protobuf to coexist. We will need to make it very explicit or just let bitswap figure out how to encode  something.

Also, right now blocks in datastore are stored as `.data`, we can have `.data` for serialized protobufs and `.cbor` for serialized IPLD objects.

exchange/README.md
```diff
@@ -0,0 +1,108 @@
...
+### wire messages
+
+bitswap wire format can vary between protobuf (legacy) and cbor (IPLD).
+
+Types of messages:
+
+- here is my want list
+- I have a block
+- I no longer want a block
+
```
diasdavid: We will also need to do some multicodec handshake on how we pass bitswap messages or we can encapsulate everything in a protobuf always and just let the protobuf tell which type the block being passed on the wire is. // @whyrusleeping

exchange/README.md
```diff
@@ -0,0 +1,108 @@
...
+└─────────┘  └────────────┘  └───────┘ └───────────────┘
+     │              ▲            │
+
+     │              │            │
+
+     ▼              │            │
+┌─────────┐ ┌───────────────┐
+│ipfs-repo│ │ libp2p        │◀ ─ ┘
+└─────────┘ └───────────────┘
+```
```
diasdavid: This is how I'm organising the pieces of bitswap so far :)

exchange/README.md
whyrusleeping: 'I have a block' isn't a bitswap message, its on the dht.

diasdavid: oh, I meant, "Here is a block I have that you wanted"

diasdavid:
going to open a new one :)

diasdavid:
See https://github.com/ipfs/specs/pull/95

exchange/README.md
bpcamac: .data is very generic.  How about .proto for Protobufs, and .cbor for serialized IPLD.

PR #59, opened 2016-01-07
Relationship with Protocol Buffers legacy IPFS node format

mildred:
In PR #37, we left out an important part of the spec aside: the relationship with protocol buffer serialization. This ought to be described as it has effects that may be far reaching.

TODO items:

- [X] Decide if we choose a format that requires path component escaping: no escaping needed
- [X] Decide which special key to use to avoid conflict with path component (`@attrs` in one proposition, with `@` escaping, `.` in the other proposition): not needed


mildred:
What is added:

## Relationship with Protocol Buffers legacy IPFS node format

IPLD has a known conversion with the legacy Protocol Buffers format. This format is defined with the Protocol Buffers syntax as:

    message PBLink {
        optional bytes  Hash = 1;
        optional string Name = 2;
        optional uint64 Tsize = 3;
    }
    
    message PBNode {
        repeated PBLink Links = 2;
        optional bytes  Data = 1;
    }

The conversion to the IPLD data model must have the following properties:

- It should be convertible back to protocol buffers, resulting in an identical byte stream (so the hash corresponds). This implies that ordering and duplicate links must be preserved in some way.
- When using paths as defined earlier in this document, links should be accessible without further indirection. This requires the top node object to have keys corresponding to link names.
- Link names should not conflict with other keys.

There are multiple ways to do that that will be described next.

### Current encoding in go-ipld

go-ipld implements the following conversion:

    {
      "<Links[0].Name.(escaped)>": {
        "hash": "<Links[0].Hash>",
        "name": "<Links[0].Name>",
        "size": <Links[0].Tsize>
      },
      "<Links[1].Name.(escaped)>": {
        "hash": "<Links[1].Hash>",
        "name": "<Links[1].Name>",
        "size": <Links[1].Tsize>
      },
      ...
      "@attrs": {
        "data": "<Data>",
        "links": [
          {
            "hash": "<Links[0].Hash>",
            "name": "<Links[0].Name>",
            "size": <Links[0].Tsize>
          },
          {
            "hash": "<Links[1].Hash>",
            "name": "<Links[1].Name>",
            "size": <Links[1].Tsize>
          }
        ]
      }
    }

Notes :

- The `links` array in the `@attrs` section is there to preserve order and duplicate links to hash back to the exact same protocol buffer object.

- The link names are escaped to prevent clashing with the `@attr` key.

- The escaping mechanism transforms the `@` character into `\@`. This mechanism also implies a modification of the path algorithm. When a path component contains the `@` character, it must be escaped to look it up in the IPLD Node object.

    For example, a path a path `/root/first@component/second@component/third_component` would look for object `root["first\@component"]["second\@component"]["third_component"]` (following mlinks when necessary).

- Links are represented using the `hash` key instead of `mlink` as used in this specification. This must be changed.

### Other proposition that does away with escaping

We can imagine another transformation where the link names are not escaped. For example:

    {
      "<Links[0].Name>": {
        "mlink": "<Links[0].Hash>",
        "tsize": <Links[0].Tsize>
      },
      "<Links[1].Name>": {
        "mlink": "<Links[2].Hash>",
        "tsize": <Links[2].Tsize>
      },
      ...
      ".": {
        "data": "<Data>",
        "links": [
          "<Links[0].Name>",
          {
            "name": "<Link[1].Name>",
            "mlink": "<Links[1].Hash>",
            "tsize": <Links[1].Tsize>
          }
          "<Links[2].Name>",
          ...
        ]
      }
    }

Notes:

- Very conveniently, we use the key `.` to represent data for the current node, and any other key can represent a link. This means that we forbid link to be named `.`. This is in any case a good thing to do as the `.` element in paths can always be removed (the same way `..` can be replaced by the parent directory)

- No escaping is needed, and no modification to the path algorithm is needed.

- Link order is kept by using the link names for links present in the top node. This avoid repeating identical data (even though it is probably generated on the fly).

    Links that cannot be present in the top node (the case for the link named `.`, which is forbidden, or for links that are repeated with the same name) are present in full to allow reconstructing the exact protocol buffer object.

### Other encodings

Speak up in comments if you have other encoding suggestions, or you have another implementation with another encoding.

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+
+- The escaping mechanism transforms the `@` character into `\@`. This mechanism also implies a modification of the path algorithm. When a path component contains the `@` character, it must be escaped to look it up in the IPLD Node object.
+
+    For example, a path a path `/root/first@component/second@component/third_component` would look for object `root["first\@component"]["second\@component"]["third_component"]` (following mlinks when necessary).
+
+- Links are represented using the `hash` key instead of `mlink` as used in this specification. This must be changed.
+
+### Other proposition that does away with escaping
+
+We can imagine another transformation where the link names are not escaped. For example:
```
jbenet: i dont like this one (using `.`) as much. i prefer the `@attrs` with escaping

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+    message PBNode {
+        repeated PBLink Links = 2;
+        optional bytes  Data = 1;
+    }
+
+The conversion to the IPLD data model must have the following properties:
+
+- It should be convertible back to protocol buffers, resulting in an identical byte stream (so the hash corresponds). This implies that ordering and duplicate links must be preserved in some way.
+- When using paths as defined earlier in this document, links should be accessible without further indirection. This requires the top node object to have keys corresponding to link names.
+- Link names should not conflict with other keys.
```
jbenet: maybe use caps for "SHOULD NOT", but note that many objects already do. so we define an access order: "in case of a conflict, use the first one to appear in the protobuf."

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+
+### Current encoding in go-ipld
+
+go-ipld implements the following conversion:
+
+    {
+      "<Links[0].Name.(escaped)>": {
+        "hash": "<Links[0].Hash>",
+        "name": "<Links[0].Name>",
+        "size": <Links[0].Tsize>
```
jbenet: maybe these should be proper `mlinks`?  would this work?

```js
{
  "<Links[0].Name.(escaped)>": {
    "mlink": "<Links[0].Hash>",
    "size": <Links[0].Tsize>
  },
  ...
  "@attrs": {
    "data": "<Data>",
    "links": [
      {
        "hash": "<Links[0].Hash>",
        "name": "<Links[0].Name>",
        "size": <Links[0].Tsize>
      },
      ...
    ]
  }
}

```

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+      }
+    }
+
+Notes :
+
+- The `links` array in the `@attrs` section is there to preserve order and duplicate links to hash back to the exact same protocol buffer object.
+
+- The link names are escaped to prevent clashing with the `@attr` key.
+
+- The escaping mechanism transforms the `@` character into `\@`. This mechanism also implies a modification of the path algorithm. When a path component contains the `@` character, it must be escaped to look it up in the IPLD Node object.
```

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+          ...
+        ]
+      }
+    }
+
+Notes:
+
+- Very conveniently, we use the key `.` to represent data for the current node, and any other key can represent a link. This means that we forbid link to be named `.`. This is in any case a good thing to do as the `.` element in paths can always be removed (the same way `..` can be replaced by the parent directory)
+
+- No escaping is needed, and no modification to the path algorithm is needed.
```
jbenet: this is a nice property

merkledag/ipld.md
```diff
@@ -308,6 +308,116 @@ In order to preserve merkle-linking's power, we must ensure that there is a sing
...
+          }
+          "<Links[2].Name>",
+          ...
+        ]
+      }
+    }
+
+Notes:
+
+- Very conveniently, we use the key `.` to represent data for the current node, and any other key can represent a link. This means that we forbid link to be named `.`. This is in any case a good thing to do as the `.` element in paths can always be removed (the same way `..` can be replaced by the parent directory)
```
jbenet: but we cant access the `data` field here either. (we cant in the above one either). 

Ok, here's an idea that i haven't fully thought through. it may be broken:
- escape user given `@`'s
- but no escaping on paths
- user must escape paths to use their keys with `@`.
- user _can_ access most of the `@directives`. (not sure if this is a good idea)
- this makes it possible to grab `$root/@attrs/data

eg these would work:

- path `/root/@attrs/data` works
- path `/root/\@attrs/data` does not work
- path `/root/first@component/second@component/third_component` does not work
- path `/root/first\@component/second\@component/third_component` works

jbenet:
i think this should go in a separate doc that extends the spec, or as an example of a special case format.

we can have it inform (and modify) the proper spec, but i think including this format in the main document will confuse people? we could link to another doc.

jbenet:
@mildred thanks for this!! :+1:

merkledag/ipld.md
mildred: I was talking about the fact that links names (for example "test.txt", "data", "@attrs.txt" or "@attrs") should not conflict with other keys used for other purposes (including data). This is the reason why we need escaping, to avoid conflicts.

merkledag/ipld.md
mildred: That's right, but they currently aren't. This is something that needs to be modified in go-ipld (and in this spec as well).

merkledag/ipld.md
mildred: That's why that's the solution (in the two exposed) that I prefer. Escaping is not very nice at this stage as it introduce two JSON representation for one node (escaped JSON, and unescaped JSON that has missing information but that is easier to work with links).

merkledag/ipld.md
mildred: See PR #60 where I make a distinction between filesystem paths (for use in filesystems) and IPLD paths (used to access every corner of the object model).

This would be ok for me for the IPLD paths. But for filesystem paths, I still want to have a file named "@attrs" in a directory somewhere (or in a HTTP url) without intending a special meaning to it.

mildred:
> I think this should go in a separate doc that extends the spec, or as an example of a special case format.

In case we go the route of escaping, this needs to be linked in other parts of the spec as it has impact on it.

merkledag/ipld.md
mildred: Even considering there is no escaping?

merkledag/ipld.md
mildred: > but we cant access the data field here either. (we cant in the above one either).

See PR #60, we could use IPLD paths so we can access this "." property:

- IPLD paths solution 1: `/root/..object/data`: is an access to `root["."]["data"]` because `..data` could be interpreted as a disambiguation form to access the `.` key
- IPLD paths solution 2: `/root/\..data`: is an access to `root["."]["data"]` assuming that `\.` is an escape sequence for the `.` character

As I said just above, it makes no sense to access these properties on the filesystem as files.

mildred:
Added considerations about escaping in the path section. See also PR #60 for similar changes.

the8472:
Based on IRC discussion and reading #60 it seems to me that traversing the IPLD structure should not be shoehorned into IP*S paths. It's more of a reflection / metadata thing for the paths themselves. Equivalent to `stat`ing a file.

Putting arbitrary keys at the top level of an associative array (hash) also scratches me the wrong way.

The metadata could look like this `{  attrs: {...}, namedlinks: {foo: ..., bar: ...}}` while the file path still only contains "foo" without "namedlinks". If you want to access the metadata you could go through a separate namespace, e.g. `/ipld/ipfs/<hash>/namedLinks/foo/attrs/`

mildred:
> Putting arbitrary keys at the top level of an associative array (hash) also scratches me the wrong way.

Especially considering that unixfs paths are separate from IPLD paths according to last state of PR #62. There is nothing preventing the data to look like this as @the8472 told us:

    {
      "data": "<Data>",
      "named-links": {
        "<Links[0].Name>": {
          "mlink": "<Links[0].Hash.(base58)>",
          "name": "<Links[0].Name>",
          "size": <Links[0].Tsize>
        },
        "<Links[1].Name>": {
          "mlink": "<Links[1].Hash.(base58)>",
          "name": "<Links[1].Name>",
          "size": <Links[1].Tsize>
        },
        ...
      }
      "ordered-links": [
        {
          "mlink": "<Links[0].Hash.(base58)>",
          "name": "<Links[0].Name>",
          "size": <Links[0].Tsize>
        },
        {
          "mlink": "<Links[1].Hash.(base58)>",
          "name": "<Links[1].Name>",
          "size": <Links[1].Tsize>
        }
      ]
    }

Is there a reason why we couldn't do this way. @jbenet?

mildred:
Updated the conversion format.

merkledag/ipld-compat-protobuf.md
```diff
@@ -0,0 +1,205 @@
+# IPLD conversion with Protocol Buffer legacy IPFS node format
+
+IPLD has a known conversion with the legacy Protocol Buffers format in order for new IPLD objects to interact with older protocol buffer objects.
+
+## Detecting if the legacy format is in use
+
+The format is encapsulated after a multicodec header that tells which codec to use. In addition, older applications that do not yet use the multicodec header will transmit a protocol buffer stream. This can be detected by looking at the first byte:
+
+- if the first byte is between 0 and 127, it is a multicodec header
+- if the first byte if between 128 and 255, it is a protocol buffer stream
```
mildred: I believe I am wrong here. The multicodec header length is not limited to one byte, but can be encoded in multiple bytes if it is above 127. Setting the MSB to 1 is just the way varint works.

I also assumed that protocol buffers always started with a byte with MSB set to 1, but I don't know if that true. Probably not.

So, it's probably not possible to detect in a such eaasy way if we are transmitting a multicodec header or a protocol buffer message. I'm currently rewriting this part (as I am implementing it in go-ipld).

merkledag/ipld-compat-protobuf.md
```diff
@@ -0,0 +1,79 @@
...
+    message PBNode {
+        repeated PBLink Links = 2;
+        optional bytes  Data = 1;
+    }
+
+## Conversion to IPLD model
+
+The conversion to the IPLD data model must have the following properties:
+
+- It should be convertible back to protocol buffers, resulting in an identical byte stream (so the hash corresponds). This implies that ordering and duplicate links must be preserved in some way.
```
jbenet: "- It MUST be ..."

merkledag/ipld-compat-protobuf.md
```diff
@@ -0,0 +1,79 @@
...
+- When using paths as defined earlier in this document, links should be accessible without further indirection. This requires the top node object to have keys corresponding to link names.
```
jbenet: "When using paths as defined in the IPLD spec..."

merkledag/ipld-compat-protobuf.md
```diff
@@ -0,0 +1,79 @@
...
+There is a canonical form which is described below:
+
+    {
+      "data": "<Data>",
+      "named-links": {
+        "<Links[0].Name>": {
+          "@link": "<Links[0].Hash.(base58)>",
+          "name": "<Links[0].Name>",
+          "size": <Links[0].Tsize>
+        },
```
jbenet: these could be offsets into the array, to decrease memory concerns. otherwise we blow up by ~2x

jbenet: (im ok with either, can maybe state implementations can just dedup?)

jbenet:
minor things, otherwise LGTM

merkledag/ipld-compat-protobuf.md
mildred: We don't duplicate links here. A link is represented only once. The rule is that:

- if the link can be represented in named-links, it is represented just there. The ordered-links section contain just the string key referencing the named link.
- if the link cannot be represented in the named-links (name conflict, malformed protocol buffer document), it only appear, this time in full, in the ordered-links section.

jbenet: Ah I see. i think we should reverse to simplify, i.e.
- always keep the full link object in `ordered-links`
- always point (with an offset number) from the `named-links` section.

mildred: I don't know. It depends on which information you want to have most accessible. The named links section was initially the only section, but the ordered links was necessary to be able to convert objects with duplicate links back to protocol buffers.

The idea is that all named links are available in the named-links section by their name, and the key is the name of the link. It's easy to make a mapping between the link and its name.

If the named links contain the index to the ordered links, making the mapping between each link and its name is a little more difficult if you are not parsing both sections at the same time. Also, the benefit of having access to the link with a path containing the link name is gone.

If we want to simplify, I'd be in favor of removing the named links section entirely instead and adding a name attribute inside the link object. There is no benefit of having a named links section being just a mapping between name and link index.

mildred:
@jbenet I pushed some fixes, also, what do you think of the ordered-links vs named-links issue ?

merkledag/ipld-compat-protobuf.md
mildred: The main argument here is that link order was never meant to be significant in the protocol buffer serialization format. We keep the order only to allow conversion back to the original format and be guaranteed we have the same hash. If not for that (and for the possibility of link conflicts), we would have removed the ordered links section entirely

jbenet: Im ok removing the named-links section, but it may help-- makes it easier to find stuff. In any case, we can remove it for now, and add it later, as the ordered-links section is the important part anyway.

Link conflicts require the ordering, so ordering is very significant. Most protobuf objects out there have lots of links with the same link name (`""`), this is current unixfs files' sub-files.

jbenet:
https://github.com/ipfs/specs/pull/59#discussion_r52722048

mildred:
@jbenet I removed the named links section and made it an option for the implementations as it is not an important part of the spec.

jbenet:
SGTM! :+1: thanks

PR #60, opened 2016-01-07
Separate filesystem merkle-path from IPLD merkle-path

mildred:
This is an attempt at solving "**Accessing link properties problem (important issue to resolve)**" from PR #37

TODO items:

- [ ] Pending decision in PR #59 about path component escaping
- [ ] Decide which format to use for _IPLD merkle-paths_ (suffix, special character or other)
- [ ] Decide which suffix (`.object` vs `?object` vs `:object`) or which special character to use (`.` vs `|`)

I proposed two solutions, and one with a variant, corresponding to [these three ideas](/ipfs/specs/pull/37#issuecomment-158766176)

Introductions to the changes:

### What is a _merkle-path_?

A merkle-path is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and allows access of elements of the referenced node and other nodes transitively.

There is no single merkle-path, but there are two:

- merkle-path for filesystems: this is a merkle-path that is designed to be used in the context of filesystems (that also includes network protocols such as HTTP or FTP). Their idea is to be as close as possible to the traditional filesystem semantic
- merkle-path for IPLD: this is a merkle-path that can be used to access more elements of the IPLD data model (specifically: link properties) but that doesn't fit within the traditional filesystem model.

When you use a merkle path, make sure of which one you use. Command line tools are encouraged to allow switching between the two flavors using a switch.

### DIFF
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -104,7 +104,8 @@ To unescape IPLD object keys that are not reserved and get the corresponding pat
 
 ### IPLD merkle-path (best solution)
 
-An _IPLD merkle-path_ is an extension of a _filesystem merkle-path_ which uses a special syntax to access link properties.
+An _IPLD merkle-path_ is an extension of a _filesystem merkle-path_ which uses a special syntax to access link properties. **[In case we use escaping in protobuf IPLD format** Except that key escaping is not performed when looking up items in the IPLD objects. This allow accessing reserved keys using _IPLD merkle-paths_ that are not accessible in filesystems.**]**
+
 
 Path elements are suffixed by either `.link` to access the link properties or by `.object` to dereference the _merkle-link_. if no suffix is present, the _merkle-link_ is dereferenced (to be compatible with _filesystem merkle-paths_ in most cases)
 
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -26,7 +26,18 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
 
 ### What is a _merkle-path_?
 
-A _merkle-path_ is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
+A merkle-path is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and allows access of elements of the referenced node and other nodes transitively.
+
+There is no single merkle-path, but there are two:
+
+- merkle-path for filesystems: this is a merkle-path that is designed to be used in the context of filesystems (that also includes network protocols such as HTTP or FTP). Their idea is to be as close as possible to the traditional filesystem semantic
+- merkle-path for IPLD: this is a merkle-path that can be used to access more elements of the IPLD data model (specifically: link properties) but that doesn't fit within the traditional filesystem model.
+
+When you use a merkle path, make sure of which one you use. Command line tools are encouraged to allow switching between the two flavors using a switch.
+
+### Filesystem merkle-path
+
+A _filesystem merkle-path_ is a unix-style path which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
 
 For example, suppose we have this _merkle-path_:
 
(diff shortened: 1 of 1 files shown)


merkledag/ipld.md
```diff
@@ -72,6 +83,81 @@ O_5 = | "hello": "world"  |  whose hash value is QmR8Bzg59Y4FGWHeu9iTYhwhiP8PHCN
 
 This entire _merkle-path_ traversal is a unix-style path traversal over a _merkle-dag_ which uses _merkle-links_ with names.
 
+### IPLD merkle-path (best solution)
+
+An _IPLD merkle-path_ is an extension of a _filesystem merkle-path_ which uses a special syntax to access link properties.
+
+Path elements are suffixed by either `.link` to access the link properties or by `.object` to dereference the _merkle-link_. if no suffix is present, the _merkle-link_ is dereferenced (to be compatible with _filesystem merkle-paths_ in most cases)
+
+**FIXME**: perhaps use different suffixes so we are less likely to have ambiguities. Using a character that is denied by Windows would be a good idea since those are less likely to be present in filenames. For most cases, this would make _IPLD merkle-paths_ a superset of _filesystem merkle-paths_. For example we could use `?link` and `?object`
```
jbenet: being able to access link properties includes making it possible to do so via HTTP

merkledag/ipld.md
```diff
@@ -72,6 +83,81 @@ O_5 = | "hello": "world"  |  whose hash value is QmR8Bzg59Y4FGWHeu9iTYhwhiP8PHCN
...
+      cat.jpg:
+        mlink: QmCCC...222
+        mode: 0644
+        owner: jbenet
+
+Now:
+
+- the path `/ipfs/QmCCC...000/stuff.foo.mode` yields `0755`
+- the path `/ipfs/QmCCC...000/stuff/foo` does not exists because in the object `QmCCC...000`, the `stuff` object cannot is not a _merkle-link_ (it doesn't have the `mlink` key)
+- the path `/ipfs/QmCCC...000/stuff.foo/other.cat\.jpg` yields object `QmCCC...222`
```
jbenet: - using `.` requiring escape is liable to confuse a lot of people since dots are used for extensions :/

(i do agree that we should avoid ambiguous meanings, like `orther.car.jpg` is totally weird)

jbenet: (maybe this is fixed by using a different separator, not sure)

merkledag/ipld.md
mildred: I did not intend to use the `?` character to enable a special HTTP feature, but we could imagine it without problem. That's a separate issue for me.

merkledag/ipld.md
mildred: That's why I prefer the first solution presented. only one needs to be chosen. And I agree if we use the second solution, another character is probably better (you suggested `|` in [your comment](https://github.com/ipfs/specs/pull/37#issuecomment-158766176))

mildred:
Commit _Talk about escaping keys in merkle-paths_ is taken from PR #59 and commit _Consideration about key escaping_ is additional remarks for _IPLD merkle-paths_ in case we use escaping.

mildred:
What path syntax do you prefer?

- fist solution: `/ipfs/QmCCC...000/stuff/foo.object/other/cat.link`
- fist solution with another character: `/ipfs/QmCCC...000/stuff/foo?object/other/cat?link`
- fist solution with yet another character: `/ipfs/QmCCC...000/stuff/foo:object/other/cat:link`
- second solution: `/ipfs/QmCCC...000/stuff.foo/other.cat\.jpg`
- second solution with another character: `/ipfs/QmCCC...000/stuff|foo/other|cat.jpg`
- second solution and allowing either `/` or `.` (or `|`) if there is no ambiguity

Perhaps more people should express their opinions on those. As for me, I prefer the first solution (that's why I put it first). I generally use prefixes and suffixes when I need to encode things in file names.

mildred:
Also, is there any objection to having two different path syntaxes? One specific for filesystems and one for accessing properties.

merkledag/ipld.md
```diff
@@ -72,6 +85,99 @@ O_5 = | "hello": "world"  |  whose hash value is QmR8Bzg59Y4FGWHeu9iTYhwhiP8PHCN
...
+- every `\@` sequence in the key must be replaced by `@`
+- every `\\` sequence in the key must be replaced by `\`
+
+
+### IPLD merkle-path (best solution)
+
+An _IPLD merkle-path_ is an extension of a _filesystem merkle-path_ which uses a special syntax to access link properties. **[In case we use escaping in protobuf IPLD format** Except that key escaping is not performed when looking up items in the IPLD objects. This allow accessing reserved keys using _IPLD merkle-paths_ that are not accessible in filesystems.**]**
+
+
+Path elements are suffixed by either `.link` to access the link properties or by `.object` to dereference the _merkle-link_. if no suffix is present, the _merkle-link_ is dereferenced (to be compatible with _filesystem merkle-paths_ in most cases)
```
Dignifiedquire: **Disclaimer:** I'm late to this party so if I missed something please excuse me

I do very much like the notion of this solution, but `.link` and `.object` seem quite verbose to me, why not use two separators like `#` and `.` instead, so the example paths would be transformed to the ones you can see below

- the path `/ipfs/QmCCC...000/stuff/foo#mode` yields `0755`
- the path `/ipfs/QmCCC...000/stuff/foo/other/cat#mode` does not exists because `other` does not have a `cat` object, only a `cat.link`
- the path `/ipfs/QmCCC...000/stuff/foo/other/cat.link.mode` yields `0644`
- the path `/ipfs/QmCCC...000/stuff/foo.other/cat#` yields object `QmCCC...222`

This would make the paths a little less verbose for me. I do see that conflicts with things like `cat.link` in the second object are still an issue, so it might make sense to use different separators, though `.` feels quite natural for the object selection to me.

Dignifiedquire:
I prefer the second solution (`/ipfs/QmCCC...000/stuff.foo/other.cat\.jpg`) because it is a) less verbose than the first one, and b) feels more "natural" as I'm used to `/` being a separator in e.g. http links and `.` being a separator for nested object notation. So something like `/ipfs/Qm...000/my.object/other.cat` is very easy to parse and write mentally for me.

mildred:
I have no strong opinions on the matter, but I just feel that `.` would be confusing with extensions (as [jbenet also notices](https://github.com/ipfs/specs/pull/60#discussion_r49112292)). Also the `\` escape character is difficult to use in some circumstances (when there are multiple layers of those).

@jbenet, do you have a preference for one syntax or the other ?

jbenet:
@mildred still mulling over. I'm re-confronting this once more.

quick note:

> Also, is there any objection to having two different path syntaxes? One specific for filesystems and one for accessing properties.

We could do this. my intuition is that it _will_ be quite confusing for people (when to use each), so if we do go down this path we need to explain the clear need for them, have them be distinct enough. looking at the simplicity of JSON is a good reminder of how simple formats win.

It is also possible to allocate a prefix, like with `/ipns` -- we could have `/unixfs/...` paths be proper files and `/ipfs/...` be raw IPLD objects. or `/ipfs/...` and `/ipld/...` or whatever. there is already a dichotomy/tension between "unixfs" + "ipfs objects".

mildred:
> It is also possible to allocate a prefix, like with /ipns -- we could have /unixfs/... paths be proper files and /ipfs/... be raw IPLD objects. or /ipfs/... and /ipld/... or whatever. there is already a dichotomy/tension between "unixfs" + "ipfs objects".

You're right, we don't need to have different kind of merkle-paths if we do not intend to use them for filesystems for HTTP gateways. If as you say (and I have forgotten) the unixfs paths is already different from merkle-paths, we can decide for just one kind of merkle-paths: what I called IPLD merkle-paths.

mildred:
Probably obsolete in favor of #62

PR #61, opened 2016-01-08
IPLD CBOR tagging

mildred:
See PR #37 

TODO items:

- [ ] register CBOR tags with IANA

### DIFF
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -312,25 +312,38 @@ IPLD supports a variety of serialized data formats through [multicodec](https://
 
 IPLD objects can be represented using cbor using the tags described below when possible. Tags are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4):
 
-- `<tag-escaped-key>`: **[If key escaping is necessary]** The string that follows (major type 2 or 3) is interpreted as an escaped string (of the same major type). Every occurrences of `\` are considered to be `\\`, and every occurrences of `@` are considered to be `\@`.
+- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string (using the IPFS alphabet).
 
-- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string.
+- `<tag-link-object>`:  an array (major type 4) of two or three elements (link prefix, link hash (optional) and map) must follow:
 
-- `<tag-ipfs-path>`: the text string (major type 2) that follows (or the byte string tagged with `<tag-base58>`) is to be interpreted with "`/ipfs/`" added in front of the string.
+    - The link prefix must be an integer representing the first path of the link, or a text string appended at the beginning of the link. Available integers are:
+        - `1`: represents the prefix `/ipfs/`
 
-- `<tag-link-object>`:  an array (major type 4) must follow. The array must have two elements: a text string (or a byte string tagged using `<tag-ipfs-path>`) followed by a map (major type 5). This whole must be interpreted as a map identical to the map of the array, but with an additional entry. The additional entry would have a text string containing `link` as a key, and the text string contained in the array as value.
+    - The link hash, either a text string to be appended after the link prefix, or a tag `<tag-base58>` followed by the binary string representing the hash digest.
 
-**FIXME:** register tags with IANA.
+    - a map
 
-When encoding an IPLD node to CBOR with tags, these tags must be included whenever possible, and avoided if not necessary. This will ensure a unique encoding across implementations. More specifically (and in this order):
+    The link value is constructed by concatenating the link prefix and the link hash (if present) in its text form.
 
-- If map key is a text string `s` and `escape(unescape(s)) == s`, then this string is transformed to `unescape(s)` and tagged with `<tag-escaped-key>`
+    This must be interpreted as a map identical to the map of the array, but with an additional entry. The additional entry would have a text string containing `link` as a key, and the text string representing the link formed by the first two elements of the array. When iterating over the map, this entry must appear first.
 
-- if a text string starts with "`/ipfs/`", this prefix is removed and the string is tagged with `<tag-ipfs-path>`.
+**TODO:**
 
-- If a text string contains a valid base58 encoded value, it is converted to a binary string and tagged with `<tag-base58>`
+- [ ] register tags with IANA.
+- [ ] specify tags we use for escaping (if we want to store escaped string in unescaped form)
 
-- If a map contains an entry which key is the text string "`link`", this entry is removed from the map, an array is created containing the entry value and the map, and this array is prefixed by the tag `<tag-link-object>`. The result is used in place of the map.
+When encoding an IPLD node to CBOR with tags, some conversion steps are necessary (in this order):
+
+- if a map contains an entry which key is the text string `link` and the value is a text string, the map is converted to a link object:
+
+    - the `link` entry is removed from the map
+    - if the link value cannot be split in a prefix and a base58 suffix, an array is created with the link value (a text string) and the transformed map.
+    - else, the link is split in a textual prefix and a base58 binary digest (the base58 value is decoded) and an array with the prefix, the `<tag-base58>` followed by the binary hash, and the transformed map is created
+    - the original map is transformed to a `<tag-link-object>` followed by the array just created
+
+- if a text string is a canonical base58 representation of a binary string, the text string is converted to binary and `<tag-base58>` is added at the beginning
+
+- if a text string is a canonical base64 representation (with no stray characters) of a binary string, the text string is converted to binary and the tag `22` (defined in RFC7049) is added at the beginning
 
 When an IPLD object contains these tags in the way explained here, the multicodec header used to represent the object codec must be `/cbor/ipld-tagsv1` instead of just `/cbor`. Readers will be able to use an optimized reading process to detect links using these tags.
 
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -310,48 +310,36 @@ IPLD supports a variety of serialized data formats through [multicodec](https://
 
 ## Serialised CBOR with tags
 
-IPLD objects can be represented using cbor using the tags described below when possible. Tags are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4):
+IPLD objects can be represented using cbor using tags which are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4).
 
-- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string (using the IPFS alphabet).
+A tag `<tag-link-object>` is defined. This tag must be followed by an array (major type 4) containing two elements. The first being either a text string (major type 3) or a byte string (major type 2). The second element is defined to be a map (major type 5).
 
-- `<tag-link-object>`:  an array (major type 4) of two or three elements (link prefix, link hash (optional) and map) must follow:
+When encoding an IPLD object to CBOR, every map that contain a link key is transformed to a `<tag-link-object>` followed by the array containing the link and then containing the CBOR version of the map without the link key.
 
-    - The link prefix must be an integer representing the first path of the link, or a text string appended at the beginning of the link. Available integers are:
-        - `1`: represents the prefix `/ipfs/`
+- if the link key is a valid [multiaddress](https://github.com/jbenet/multiaddr) and converting that link text to the multiaddress binary string and back to text is guaranteed to result to the exact same text, the link is stored as a binary multiaddress as the first array item.
+- else, the link is stored as text as the first array item.
 
-    - The link hash, either a text string to be appended after the link prefix, or a tag `<tag-base58>` followed by the binary string representing the hash digest.
+When decoding CBOR and converting it to IPLD, each occurences of `<tag-link-object>` with its following array is transformed.
 
-    - a map
+- If the first array item is a binary string, it is interpreted as a multiaddress and converted to a textual format. Else, the text string is used directly.
+- The map that follows is augmented with a new pair. The key is the standard IPLD link property, the value is the link in its textual format.
+- When iterating over this augmented map, the link property must come first and not in any other order. This guarantee a consistent ordering.
+- This augmented map is used instead of the `<tag-link-object>` in the IPLD output.
 
-    The link value is constructed by concatenating the link prefix and the link hash (if present) in its text form.
-
-    This must be interpreted as a map identical to the map of the array, but with an additional entry. The additional entry would have a text string containing `link` as a key, and the text string representing the link formed by the first two elements of the array. When iterating over the map, this entry must appear first.
+When an IPLD object contains these tags in the way explained here, the multicodec header used to represent the object codec must be `/cbor/ipld-tagsv1` instead of just `/cbor`. Readers will be able to use an optimized reading process to detect links using these tags.
 
 **TODO:**
 
 - [ ] register tags with IANA.
-- [ ] specify tags we use for escaping (if we want to store escaped string in unescaped form)
 
-When encoding an IPLD node to CBOR with tags, some conversion steps are necessary (in this order):
-
-- if a map contains an entry which key is the text string `link` and the value is a text string, the map is converted to a link object:
-
-    - the `link` entry is removed from the map
-    - if the link value cannot be split in a prefix and a base58 suffix, an array is created with the link value (a text string) and the transformed map.
-    - else, the link is split in a textual prefix and a base58 binary digest (the base58 value is decoded) and an array with the prefix, the `<tag-base58>` followed by the binary hash, and the transformed map is created
-    - the original map is transformed to a `<tag-link-object>` followed by the array just created
-
-- if a text string is a canonical base58 representation of a binary string, the text string is converted to binary and `<tag-base58>` is added at the beginning
-
-- if a text string is a canonical base64 representation (with no stray characters) of a binary string, the text string is converted to binary and the tag `22` (defined in RFC7049) is added at the beginning
-
-When an IPLD object contains these tags in the way explained here, the multicodec header used to represent the object codec must be `/cbor/ipld-tagsv1` instead of just `/cbor`. Readers will be able to use an optimized reading process to detect links using these tags.
 
 ### Canonical Format
 
 In order to preserve merkle-linking's power, we must ensure that there is a single **_canonical_** serialized representation of an IPLD document. This ensures that applications arrive at the same cryptographic hashes. It should be noted --though-- that this is a system-wide parameter. Future systems might change it to evolve representations. However we estimate this would need to be done no more than once per decade.
 
 **The IPLD Canonical format is _canonicalized CBOR with tags_.**
+
+Users of this format should not expect any specific ordering of the keys, as the keys might be ordered differently in non canonical formats.
 
 The legacy canonical format is protocol buffers.
 
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -308,7 +308,7 @@ On the subject of integers, there exist a variety of formats which represent int
 
 IPLD supports a variety of serialized data formats through [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `@link`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
 
-## Serialised CBOR with tags
+### Serialised CBOR with tags
 
 IPLD links can be represented in CBOR using tags which are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4).
 
@@ -320,8 +320,8 @@ A tag `<tag-link-object>` is defined. This tag can be followed by:
 When encoding an IPLD object to CBOR, every IPLD object can be considered to be encoded using `<tag-link-object>` using this algorithm:
 
 - If the IPLD object doesn't contain a link property, it is encoded in CBOR as a map.
-- If the IPLD object contain a link property but it is not a string, it is encoded in CBOR as a map.
-- The link property is extracted and the object is converted to a map that don't contain the link.
+- If the IPLD object contains a link property but it is not a string, it is encoded in CBOR as a map.
+- The link property is extracted and the object is converted to a map that doesn't contain the link.
 - If the link is a valid [multiaddress](https://github.com/jbenet/multiaddr) and converting that link text to the multiaddress binary string and back to text is guaranteed to result to the exact same text, the link is converted to a binary multiaddress stored in CBOR as a byte string (major type 2).
 - Else, the link is stored as text (major type 3)
 - If the map created earlier is empty, the resulting encoding is the `<tag-link-object>` followed by the CBOR representation of the link
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -312,7 +312,10 @@ IPLD supports a variety of serialized data formats through [multicodec](https://
 
 IPLD links can be represented in CBOR using tags which are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4).
 
-A tag `<tag-link-object>` is defined. This tag must be followed by an array (major type 4) containing one or two elements. The first being either a text string (major type 3) or a byte string (major type 2). The second element is defined to be a map (major type 5) and can be omitted if the map is empty. The canonical format is to omit this map if it is empty.
+A tag `<tag-link-object>` is defined. This tag can be followed by:
+
+- a text string (major type 3) or byte string (major type 2) corresponding to the link target. This is the canonical format for links with no link properties.
+- an array (major type 4) containing as first element the link target (text or binary string) and as optional second argument the link properties (a map, major type 5)
 
 When encoding an IPLD object to CBOR, every IPLD object can be considered to be encoded using `<tag-link-object>` using this algorithm:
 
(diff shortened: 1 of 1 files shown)


mildred:
@davidar, any opinions on this? @jbenet? anyone?

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
 
+## Serialised CBOR with tags
+
+IPLD objects can be represented using cbor using the tags described below when possible. Tags are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4):
+
+- `<tag-escaped-key>`: **[If key escaping is necessary]** The string that follows (major type 2 or 3) is interpreted as an escaped string (of the same major type). Every occurrences of `\` are considered to be `\\`, and every occurrences of `@` are considered to be `\@`.
+
+- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string.
```
jbenet: nice! wonder --then-- if we can store the hashes compressed. having the hashes encoded on base58 is a space waste, so if a small conversion can be done transparently, it may be worth it.

(https://github.com/jbenet/multiaddr/ does something like this.  this may be well beyond scope here, and may not be worth the time.)

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+- `<tag-ipfs-path>`: the text string (major type 2) that follows (or the byte string tagged with `<tag-base58>`) is to be interpreted with "`/ipfs/`" added in front of the string.
```

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+- `<tag-link-object>`:  an array (major type 4) must follow. The array must have two elements: a text string (or a byte string tagged using `<tag-ipfs-path>`) followed by a map (major type 5). This whole must be interpreted as a map identical to the map of the array, but with an additional entry. The additional entry would have a text string containing `link` as a key, and the text string contained in the array as value.
+
+**FIXME:** register tags with IANA.
+
+When encoding an IPLD node to CBOR with tags, these tags must be included whenever possible, and avoided if not necessary. This will ensure a unique encoding across implementations. More specifically (and in this order):
+
+- If map key is a text string `s` and `escape(unescape(s)) == s`, then this string is transformed to `unescape(s)` and tagged with `<tag-escaped-key>`
```
jbenet: maybe i'm misreading, did you mean `unescape(escape(s))` ?

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+- if a text string starts with "`/ipfs/`", this prefix is removed and the string is tagged with `<tag-ipfs-path>`.
+
+- If a text string contains a valid base58 encoded value, it is converted to a binary string and tagged with `<tag-base58>`
```
jbenet: so with these two, could we get from `/ipfs/Qm....` to `<binary-version-of-Qm....>` ?

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+- If a map contains an entry which key is the text string "`link`", this entry is removed from the map, an array is created containing the entry value and the map, and this array is prefixed by the tag `<tag-link-object>`. The result is used in place of the map.
```
jbenet: nice way to do it. it's cool thinking in cbor.

one downside of all these tags is that the conversion is not "straightforward cbor to json" but this might be fine with a nice `ipld-*` toolchain that handles all the conversions in the CLI and the web

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+When an IPLD object contains these tags in the way explained here, the multicodec header used to represent the object codec must be `/cbor/ipld-tagsv1` instead of just `/cbor`. Readers will be able to use an optimized reading process to detect links using these tags.
```
jbenet: perfect :+1:

merkledag/ipld.md
```diff
@@ -302,11 +302,37 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
 ### Canonical Format
 
 In order to preserve merkle-linking's power, we must ensure that there is a single **_canonical_** serialized representation of an IPLD document. This ensures that applications arrive at the same cryptographic hashes. It should be noted --though-- that this is a system-wide parameter. Future systems might change it to evolve representations. However we estimate this would need to be done no more than once per decade.
 
-**The IPLD Canonical format is _canonicalized CBOR_.**
+**The IPLD Canonical format is _canonicalized CBOR with tags_.**
```

merkledag/ipld.md
mildred: No, we are considered `s` to be the escaped string that is described in the IPLD model or that is found in the JSON data structure. You check that `escape(unescape(s))` is equal to `s` to check that the original string had all its special characters escaped.

For example:

- `"abc\@def\ghi"` can be considered to be an escape version of `"abc@def\ghi"`, but that's not the canonical form. `escape(unescape("abc\@def\ghi")) == "abc\@def\\ghi"` which is different. In this case, we don't tag the string.

- `"abc\@def\\ghi"` is the canonical escaped version of `"abc@def\ghi"`. It can be stored as `<tag-escaped-string>"abc@def\ghi"` because by escaping it we can find back the original string.

The reference string is still the escaped string.

jbenet: ah that's right, thanks

merkledag/ipld.md
mildred: This spec describes how to transform textual base58 encoded data to actual binary data in the wire format. It can be done with other textual formats that could be represented in binary, only if the function to convert between the two is bijective (there is only one canonical way to convert).

We can add this later, that's why the multicodec header contains a version number (`/cbor/ipld-tagsv1`)

merkledag/ipld.md
mildred: That's exactly the idea

merkledag/ipld.md
mildred: These tags will be registered with IANA, so anyone would be able to implement them. But there is already so many tags... So we'll have to handle the conversion. But to access these CBOR objects, you already have to talk to an IPFS program.

And these are optional, if we want to avoid them, just use the naked `/cbor` multicodec header and don't use these tags.

merkledag/ipld.md
mildred: This will perhaps be a little more tricky to implement tags in IPLD implementations than if we just had used the naked CBOR format.

davidar:
@mildred @jbenet This LGTM for the most part. I had a slightly different tagging system in my head - I suspect it's more or less equivalent to what you're doing, but here it is for completeness:

Assuming no escaping, I think we could get by (perhaps?) with a single extra tag:

- `<tag-link-object>`:  Either 
  - a pair (size-2 array, major type 4) must follow. The first element is a byte string representing a (binary) multihash (i.e. for an address "/ipfs/QmFooBar", strip the "/ipfs/" prefix and base58-decode the remainder). The second element is a map (major type 5) containing any additional metadata attributes.
  - if there is no additional metadata, the tag may be followed by a single (binary multihash) bytestring (instead of a pair of the bytestring and the empty map)

If we *do* have escaping, would it would make more sense to tag the *un*escaped keys rather than the escaped ones? In other words, should we tag the "special" `@keys` rather than the "regular" `keys`? For instance:

- `<tag-special-key>`: The string that follows (major type 2 or 3) is interpreted as a *special* `@directive` key. For example, a JSON key `"@mode"` is translated into CBOR as `<tag-special-key><string>mode`. **Untagged** keys are assumed to be taken literally, and hence `@` and `\` characters should be escaped when encoded as JSON, e.g. CBOR `<string>foo@bar\baz` -> JSON `"foo\@bar\\baz"`

Please let me know if the assumptions I'm making about valid mlink addresses and/or escaping are incorrect.

mildred:
@davidar I thought about the three tags after [these](/ipfs/specs/pull/37#issuecomment-169775469) [comments](https://github.com/ipfs/specs/pull/37#issuecomment-170091858) from issue #37. The idea was to be able to express not only /ipfs/ links, but also other kind of links. The idea was also to allow compressing base58 data independently from the fact that it is a link. is that over-engineering?

Now that I think about it, the solution presented in this PR is not ideal since new tags must be registered with IANA when we need to express other kind of links (`/ipld/...` or others).

As for the escaping, I don't know if we should give it special treatment. I think we should not. Currently, escaping is only needed to convert legacy protocol buffer objects (see PR #59). On new IPLD objects, escaping will not be used at all (or will be defined by individual applications). Merkle-paths are no longer involved in key escaping if you look at PR #62.

mildred:
I would still like to keep a separate base58 tag that is using the same idea as the existing base64 tag defined in the [CBOR RFC section 2.4.4.2](http://tools.ietf.org/html/rfc7049#section-2.4.4.2). So perhaps we should define at least two tags:

- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string.

- `<tag-link-object>`: an array of three items must follow containing a link prefix (text string), a link hash (text string or base58 tag) and a map. The tag when interpreted is a map with a first element with key `link` and value `link_prefix + link_hash`, and the other elements as defined by the map (3rd element).

davidar:
> I would still like to keep a separate base58 tag that is using the same idea as the existing base64 tag 

Sounds reasonable. We might need some way of specifying which variant of base58 we're using though.

> when we need to express other kind of links (/ipld/... or others).

Hmm, that does make things more tricky. Embedding textual prefixes seems a little bit wasteful, but numeric identifiers might be too rigid.

On that topic, I'm having some difficulties resolving the inconsistency between multihash and path addresses, since we're mixing two different prefixing strategies. Something like `/ipfs/QmFooBar` would be more consistent as either

- `/ipfs/sha256/FooBar`, or
- `IpQmFooBar` (or whatever)

@jbenet could you clarify the situation here?

mildred:
> Sounds reasonable. We might need some way of specifying which variant of base58 we're using though.

I know there are multiple alphabets, but perhaps we shouldn't encourage using too much variations. To handle different base58 alphabets, I would introduce a new tag  for each. What do you think?

> Hmm, that does make things more tricky. Embedding textual prefixes seems a little bit wasteful, but numeric identifiers might be too rigid.

We can also introduce a numerical mapping, but defined in this spec (and further iteration of this spec). And fallback to a string if the numerical mapping is not enough. We avoid registering every new link with IANA and allow extensions if necessary.

> As for the escaping, I don't know if we should give it special treatment. I think we should not.

If we do, adding tags to the non escaped keys is more difficult because the IPLD data model defines keys as escaped. I prefer to take the approach of adding tags to the values we modified compared to the IPLD data model (the escaped keys we are unescape) rather than adding tags to the unmodified values (the non escaped values).

What I would do to add tags to escaped keys only would be:

- add `<escaped-map>` tag to CBOR maps that contain escaped keys
- keys in this map that are escaped in IPLD data model (`escape(unescape(key)) == key`) are stored unescaped in that map
- keys that do not contain all escape sequences (`escape(unescape(key)) != key`) are stored with `<tag-unescaped>` in front

mildred:
Pushed a new version which has a simpler escaping mechanism (more flexible) and removed tagging of escaped keys (until it is more needed).

mildred:
@davidar Is it still LGTM? it is for me at this point.

davidar:
waiting for clarification from @jbenet first

jbenet:
### On the path addresses

- A multihash is explicitly designed _not_ to have textual hash names, it is precisely done this way to include the function along with the digest value and ensure it is transferred always in-band correctly. many systems have no out of band way of specifying such values. see https://github.com/jbenet/multihash
- to solve the annoying "what base is this" question and to allow other bases, a new version of multihash _may_ gain a base encoding character prefix, see discussion here https://github.com/jbenet/multihash/issues/14 (we may be able to get away with this while being backwards compatible)
- the `/ipfs/<hash>/<path>` address is a path, effectively a "Filesystem URI" -- a URI whose scheme identifier is unix filesystem friendly. the schemes may vary and have different URIs altogether. like `/ipns/<name>/<path>`, or `/dns/<domain>/<path>`. 
- @davidar while it may seem "more consistent" to do all text or all encoded, it's not-- the point is that the hash is a _component_ in the path. note that the path naming has vast implications when dealing with all sorts of systems (filesystems, browsers, etc). and an explicit goal is creating trustable URIs (trustable paths).

merkledag/ipld.md
```diff
@@ -302,11 +302,50 @@ On the subject of integers, there exist a variety of formats which represent int
 
 IPLD supports a variety of serialized data formats through [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `mlink`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
 
+## Serialised CBOR with tags
+
+IPLD objects can be represented using cbor using the tags described below when possible. Tags are defined in [RFC 7049 section 2.4](http://tools.ietf.org/html/rfc7049#section-2.4):
+
+- `<tag-base58>`: the byte string that follows (major type 2) is to be interpreted as a text string instead (major type 3). This text string is the base58 encoded version of the byte string (using the IPFS alphabet).
```
jbenet: Would the presence of https://github.com/jbenet/multihash/issues/14 change the thinking here? (im not set that https://github.com/jbenet/multihash/issues/14 will happen, but it would solve an annoying wrinkle, and speed up a bunch of applications, allowing us to chose the encoding to fit the use case better)

jbenet: i think we should endeavor to always store the compressed version of the multihash. perhaps what we really need is a format like multiaddr's binary and text (compressed/uncompressed) duality: https://github.com/jbenet/multiaddr#binary-format

We could actually get away with doing this with multiaddr directly, we would just need to add types and an int code for all schemes we want to support (`ipfs, ipns, dns` for now). OR we could have a thing like multiaddr, but with a different table, explicitly for URI things...

merkledag/ipld.md
```diff
@@ -302,11 +302,50 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+- `<tag-link-object>`:  an array (major type 4) of two or three elements (link prefix, link hash (optional) and map) must follow:
+
+    - The link prefix must be an integer representing the first path of the link, or a text string appended at the beginning of the link. Available integers are:
+        - `1`: represents the prefix `/ipfs/`
```
jbenet: yeah a multiaddr-like thing would make this much simpler, eh? (and not require two CBOR tags)

davidar: @jbenet yeah, that would resolve my concerns

merkledag/ipld.md
mildred: > Would the presence of jbenet/multihash#14 change the thinking here?

I commented there, but I don't like this idea when applied to lower level components. It creates a few problems, the first being that a single multihash could now be encoded in various different ways. We could always make sure the protocol would only have a canonical version (say binary encoded)... This could work.

And CBOR already have a tag to represent base64 encoded data, it only makes sense to have a similar tag for base58.

As for encoding the path prefix, I admit I don't understand how you imagine them. I had the idea that a unix path was relative only to a single computer, and depending on the computer you have, the path might represent different things. URLs add a prefix (followed by `:`) to the path to express how they should be interpreted. Since day one, I was puzzled by the use of unprefixed paths in IPFS. So I have little idea how they should be encoded.

mildred:
The more I think about jbenet/multihash#14, the more I like it. Still I see so many use cases to the multihash without the encoding prefix I think it should be done as a separate layer.

There is one issue left with that vision of the multihash. We can't encode a binary multihash as a path component because the binary string might contain the byte `/` and that specifically disallowed for path components. When splitting a path into components, that would be a problem.

We can decide that this is becoming too complex and don't strive for binary encoding of paths. However, let's suppose we have solved this for the rest of this post.

If we apply this to our problem here, we could encode links as paths containing binary multihash, having a way to decode them to a textual representation:

- `<tag-link>` followed by a binary string:

    - the first byte would be taken apart and interpreted as a codec as specified in jbenet/multihash#14
    - in the following bytes, take all path components. For each of them, if it can be interpreted as a binary multihash, it is converted to a multihash encoded as the first byte specifies.

    We'll have to make sure we don't transform the path too much. In particular, if there are double `//` in the paths, even though they can be simplified to a single `/`, this simplification should not happen. We strive to get the exact binary compatible representation we had in input.

Now, for encoding:

- Suppose we have a textual link we want to encode in CBOR:

    - examine each path component, if a path component can be interpreted as a multihash with a byte prefix encoding, encode it to binary
    - if there were multiple path components and they were encoded differently (one in base64, the other in base58), abort and keep the original string
    - if any were encoded in a way that cannot be completely reproduced (hexadecimal for example where uppercase/lowercase is cannot be reproduced from binary), abort
    
    If we didn't abort, we have one path component or multiple path components converted to binary, but all with the same original encoding. Take the character encoding and put it in front of the string.

----

This is becoming quite complex. A little too much for my liking. Why not just require paths to be base58 encoded and support no other encoding?

merkledag/ipld.md
jbenet: > And CBOR already have a tag to represent base64 encoded data, it only makes sense to have a similar tag for base58.

nice

> As for encoding the path prefix, I admit I don't understand how you imagine them. I had the idea that a unix path was relative only to a single computer, and depending on the computer you have, the path might represent different things. URLs add a prefix (followed by :) to the path to express how they should be interpreted. Since day one, I was puzzled by the use of unprefixed paths in IPFS. So I have little idea how they should be encoded.

IPFS paths are similar, im sorry for the confusion there-- it's an annoying uphill battle for unixy composability. (someone recorded a video of me explaining in detail (ranting?) about this-- wonder what became of it.) the path `/ipfs/<hash>` is "like" `ipfs://<hash>` it's just that we make the scheme just _the first path component_ instead of a colon-separated structure, so that it fits in more places + is nicely composable/mappable.

jbenet:
> The more I think about jbenet/multihash#14, the more I like it. Still I see so many use cases to the multihash without the encoding prefix I think it should be done as a separate layer.

Agreed on separate layer.


> There is one issue left with that vision of the multihash. We can't encode a binary multihash as a path component because the binary string might contain the byte `/` and that specifically disallowed for path components. When splitting a path into components, that would be a problem.

Yes, this is indeed a problem. a binary encoded multihash in a string path is dangerous and should not be supported. we were already bitten by something like this.


> We can decide that this is becoming too complex and don't strive for binary encoding of paths. However, let's suppose we have solved this for the rest of this post.
>
> If we apply this to our problem here, we could encode links as paths containing binary multihash, having a way to decode them to a textual representation:
>
> - `<tag-link>` followed by a binary string:
>
>     - the first byte would be taken apart and interpreted as a codec as specified in jbenet/multihash#14
>     - in the following bytes, take all path components. For each of them, if it can be interpreted as a binary multihash, it is converted to a multihash encoded as the first byte specifies.
>
>     We'll have to make sure we don't transform the path too much. In particular, if there are double `//` in the paths, even though they can be simplified to a single `/`, this simplification should not happen. We strive to get the exact binary compatible representation we had in input.
>
> Now, for encoding:
>
> - Suppose we have a textual link we want to encode in CBOR:
>
>     - examine each path component, if a path component can be interpreted as a multihash with a byte prefix encoding, encode it to binary
>     - if there were multiple path components and they were encoded differently (one in base64, the other in base58), abort and keep the original string
>     - if any were encoded in a way that cannot be completely reproduced (hexadecimal for example where uppercase/lowercase is cannot be reproduced from binary), abort
>    
>     If we didn't abort, we have one path component or multiple path components converted to binary, but all with the same original encoding. Take the character encoding and put it in front of the string.
>
> ----
>
> This is becoming quite complex. A little too much for my liking. Why not just require paths to be base58 encoded and support no other encoding?

I think the above would work, yes. But agreed. this is getting very complex. I think _paths_ should work like this:

- have a textual representation. no binary multihashes. only text-encoded multihashes (i.e. base58, and possibly other bases if jbenet/multihash#14 succeeds).
- (possibly) have a binary representation for the path itself, similar to --or just using-- `multiaddr`. the binary representation of the path allows for compression of the path.

---

On the compression-- it may be the case that gzipping or huffman coding everything (wire transports and disk representations) would get most of the benefits anyway.

jbenet:
@mildred have you taken a look at how multiaddr's binary encoding works? 

- https://github.com/jbenet/multiaddr
- https://github.com/jbenet/go-multiaddr

its readme is not very good, but the gist is that something like this:

```sh
# get a multiaddr
> ipfs swarm addrs local --id | head -n1
/ip4/127.0.0.1/tcp/4001/ipfs/Qmbatr97vqpmrXg6tubdpUBwJrLzvYPHe9vS7xc9b7y3nk

# save it to var
> addr=/ip4/127.0.0.1/tcp/4001/ipfs/Qmbatr97vqpmrXg6tubdpUBwJrLzvYPHe9vS7xc9b7y3nk

# export the compressed binary representation
> multiaddr --format bytes $(addr) | wc -c
  46
```

So we can have a compressed representation, which also turns a scheme string like "/ipfs" into a single byte. so a path lik `/ipfs/<base58-encoded-string>` could be as small as `<1-byte-for-scheme><1-byte-for-len><binary-string>`. 

The multiaddr route may not be what we want here, but it is something we already have, and solves this exact use case. And it _already_ has support for [base58 text/binary coding](https://github.com/jbenet/go-multiaddr/blob/master/codec.go#L203-L212). (if it is a multihash (i.e. our use case) we have redundant len varints, but useful for base58 things that are not self-described). this facet of multiaddr could be extended with whatever jbenet/multihash#14 becomes.

davidar:
What kind of addresses do we want links to support? Will they just be flat `/ipfs/QmFooBar`, or the full gamut of `/ipfs/QmFooBar/foo/bar/baz.quux`? Personally, the latter seems excessive as it doesn't add any expressive power over the former (unless we want to be linking to fragments of objects, which seems odd). The former would be much easier to implement (as opaque pointers, with a namespace):

```
<cbor-tag-pointer> <cbor-integer=namespace> <cbor-bytestring=address>
                   /ipfs/                   QmFooBar...
```

jbenet:
> the latter seems excessive as it doesn't add any expressive power over the former (unless we want to be linking to fragments of objects, which seems odd).

We do want to link to fragments. that's one of the important parts of IPLD. being able to make larger objects where needed/reasonable, but be able to link seamlessly to parts of them without having to duplicate entities.

davidar:
@jbenet hrm, that does complicate things a bit. We could still encode the `/ipfs/QmFoo` prefix as above, but I suspect encoding the rest of the path isn't going to buy us much over a plain text representation. Another option would be to introduce (yet another) CBOR tag to express "extract this attribute from this object", but that might be over-engineering the problem. Unless that would be a generally useful thing to have in its own right, but I can't think of any cases off the top of my head.

I'm tempted to say, make every fragment semantically a separate object, and only compose them into larger objects as a practical optimisation, but I suspect that would be a rather major change.

[ If we had computable objects already it would also be easier --- just link to the the program that extracts the relevant fragment from an object :p ]

mildred:
I haven't got any opinion on that. I didn't know multiaddr had a binary encoding, and that may be a nice thing to use. We'll have to encode the leading path fragment of course.

We have to keep in mind that any binary encoding we have will have to ensure that any text we encode gets decoded in the exact same way. Possible issues would be double `/`, case insensitive characters, leading or trailing `/` that might be ignored / non significant...

For all `link` we must ensure that `decode(encode(link)) == link`.

mildred:
I am now convinced that we should define the format of the address somewhere else than in IPLD. And it could be multiaddr or an extension of it if we want to express link fragments. Then, the encoding would be simple:

- if the link property is a valid multiaddr and we can guarantee that `decode-maddr(encode-to-maddr(link)) == link` in every case, encode the link to a multiaddr format
- else keep the link in plain text (it's an invalid link anyway)

Are you all ok with that ?

mildred:
@davidar, I pushed a version of the encoding with a single tag as you suggested first. This is much more simple. The only difference is that the binary encoding of the address is not specified. it is left to the multiaddress to specify.

mildred:
I implemented this spec in [go-ipd/coding/cbor](/mildred/go-ipld/tree/reader/coding/cbor).

davidar:
@mildred LGTM :)

jbenet:
@mildred 

> if the link property is a valid multiaddr and we can guarantee that decode-maddr(encode-to- maddr(link)) == link in every case, encode the link to a multiaddr format
> else keep the link in plain text (it's an invalid link anyway)
> Are you all ok with that ?

yes sgtm

merkledag/ipld.md
```diff
@@ -300,13 +300,53 @@ On the subject of integers, there exist a variety of formats which represent int
...
+IPLD supports a variety of serialized data formats through [multicodec](https://github.com/jbenet/multicodec). These can be used however is idiomatic to the format, for example in `CBOR`, we can use `CBOR` type tags to represent the merkle-link, and avoid writing out the full string key `@link`. Users are encouraged to use the formats to their fullest, and to store and transmit IPLD data in whatever format makes the most sense. The only requirement **is that there MUST be a well-defined one-to-one mapping with the IPLD Canonical format.** This is so that data can be transformed from one format to another, and back, without changing its meaning nor its cryptographic hashes.
+
+## Serialised CBOR with tags
```
jbenet: this should be `###` level

merkledag/ipld.md
```diff
@@ -300,13 +300,53 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+A tag `<tag-link-object>` is defined. This tag can be followed by:
+
+- a text string (major type 3) or byte string (major type 2) corresponding to the link target. This is the canonical format for links with no link properties.
+- an array (major type 4) containing as first element the link target (text or binary string) and as optional second argument the link properties (a map, major type 5)
+
+When encoding an IPLD object to CBOR, every IPLD object can be considered to be encoded using `<tag-link-object>` using this algorithm:
+
+- If the IPLD object doesn't contain a link property, it is encoded in CBOR as a map.
+- If the IPLD object contain a link property but it is not a string, it is encoded in CBOR as a map.
```
jbenet: `contains`

merkledag/ipld.md
```diff
@@ -300,13 +300,53 @@ On the subject of integers, there exist a variety of formats which represent int
...
+- The link property is extracted and the object is converted to a map that don't contain the link.
```
jbenet: "... that doesn't contain..."

merkledag/ipld.md
```diff
@@ -300,13 +300,53 @@ On the subject of integers, there exist a variety of formats which represent int
...
+- If the link is a valid [multiaddress](https://github.com/jbenet/multiaddr) and converting that link text to the multiaddress binary string and back to text is guaranteed to result to the exact same text, the link is converted to a binary multiaddress stored in CBOR as a byte string (major type 2).
+- Else, the link is stored as text (major type 3)
+- If the map created earlier is empty, the resulting encoding is the `<tag-link-object>` followed by the CBOR representation of the link
+- If the map is not empty, the resulting encoding is the `<tag-link-object>` followed by an array of two elements containing the link followed by the map
```
jbenet: all this SGTM

merkledag/ipld.md
```diff
@@ -300,13 +300,53 @@ On the subject of integers, there exist a variety of formats which represent int
...
+
+When decoding CBOR and converting it to IPLD, each occurences of `<tag-link-object>` is transformed by the following algorithm:
+
+- If the following value is an array, its elements are extracted. First the link followed by the link properties. If there are no link properties, an empty map is used instead.
+- Else, the following value must be the link, which is extracted. The link properties are created as an empty map.
+- If the link is a binary string, it is interpreted as a multiaddress and converted to a textual format. Else, the text string is used directly.
```
jbenet: yeah i think `multiaddr` will work fine, just need to expand its scope to pathing a bit. -- this may be a different format per se (that's equivalent minus the network facilities of multiaddr), like `multiaddr-path` or just `binary-path` / `binpath`. 

im +1 on calling it a multiaddr for now until that o/ gets figured out.

merkledag/ipld.md
jbenet: hmmm yes. it would be a pain to have to convert to hash.  how easy is it to extend cbor codecs out there with new tags?

jbenet:
@mildred minor edits, else im good to merge this.

mildred:
Are multiaddresses the place to define merkle-links ? Or do we need something that works the same but is specifically designed for merkle links?

mildred:
I rebased this branch on top of `ipld-spec`

jbenet:
Thanks @mildred -- this LGTM!

PR #62, opened 2016-01-08
IPLD merkle-path improvements

mildred:
Improves PR #37 and replaces PR #60. The idea is that there is only one kind of _merkle-paths_ that are not to be confused with unixfs paths. These paths are powerful enough to be able to access multiple properties in IPLD objects and resolve merkle links.

### DIFF
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -34,23 +34,18 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
 
 A merkle-path is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and allows access of elements of the referenced node and other nodes transitively.
 
-There is no single merkle-path, but there are two:
+Merkle paths aren't suited to be used in filesystem representations (fuse mounts, HTTP or FTP protocols) as they describe the underlying IPLD data structure. Their use in filesystems is howver well suited for debug purposes (like `/proc` on unix).
 
-- merkle-path for filesystems: this is a merkle-path that is designed to be used in the context of filesystems (that also includes network protocols such as HTTP or FTP). Their idea is to be as close as possible to the traditional filesystem semantic
-- merkle-path for IPLD: this is a merkle-path that can be used to access more elements of the IPLD data model (specifically: link properties) but that doesn't fit within the traditional filesystem model.
+Filesystems are encouraged to design an object model on top of IPLD that would be specialized for file manipulation and have specific path algorithms to query this model
 
-When you use a merkle path, make sure of which one you use. Command line tools are encouraged to allow switching between the two flavors using a switch.
+### How do _merkle-paths_ work?
 
-Filesystem representations (fuse mounts, HTTP or FTP protocols) should use the _filesystem merkle-paths_ if they intend to store arbitrary file. They can allow switching to _IPLD merkle-paths_ using a mount option or a configuration switch to allow object inspection, and turn the filesystem something like `/proc` or `/sys` on unix machines where storing user files is not the objective.
-
-### Filesystem merkle-path
-
-A _filesystem merkle-path_ is a unix-style path which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
+A _merkle-path_ is a unix-style path which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
 
 For example, suppose we have this _merkle-path_:
 
 ```
-/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d
+/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d/
 ```
 
 Where:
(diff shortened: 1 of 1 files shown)


merkledag/ipld.md
```diff
@@ -45,33 +53,83 @@ Suppose also that this path points to the object `{ "hello": "world" }`.
...
+
+For example, suppose we have this _merkle-path_:
+
+```
+/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a.b/c.d
+```
+
+The link will:
+
+- look up the first object `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` that we call `root`
```
jbenet: let's maybe call this one `object0` instead of `root`?

merkledag/ipld.md
```diff
@@ -45,33 +53,83 @@ Suppose also that this path points to the object `{ "hello": "world" }`.
...
+
+Note that if we added a trailing slash to the path (`/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a.b/c.d/`), we would perform a last _merkle-link_ dereferencing:
+
+- look up the first object `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` that we call `root`
+- Look up the key `root["a"]["b"]` and find here a _merkle-link_
+- Dereference this _merkle-link_ to get `object1`
+- Look up the key `object1["c"]["d"]` and find a _merkle-link_
+- Dereference this _merkle-link_ and return the IPLD object as the result
+
+Also, in case the IPLD object does not contain a _merkle-link_, it is possible to use both the `/` or the `.` separator as there is no ambiguity.
```
jbenet: is this to mean that _if there is a merkle-link_ we cannot use `/`? (because of the potential ambiguity)

merkledag/ipld.md
```diff
@@ -45,33 +53,83 @@ Suppose also that this path points to the object `{ "hello": "world" }`.
...
+```
+
+The link will:
+
+- look up the first object `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` that we call `root`
+- Look up the key `root["a"]["b"]` and find here a _merkle-link_
+- Dereference this _merkle-link_ to get `object1`
+- Look up the key `object1["c"]["d"]` which will be returned as the result
+
+Note that if we added a trailing slash to the path (`/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a.b/c.d/`), we would perform a last _merkle-link_ dereferencing:
```
jbenet: yes, this is a strong solution possibility, for these paths. it is not without its problems, as a) people dislike this when applied to symlinks, and b) web servers and clients will often redirect trailing or non-trailing with varying rules.

jbenet:
@mildred thanks, this is a strong solution possibility. 

(re the unixfs paths being different, `/ipfs/` links on the gateway (which are sort of unixfs paths), for example, output the concatenated data of a file, instead of the raw objects, and have directory listings, and we cannot access children links of files there)

merkledag/ipld.md
mildred: If the object is a merkle-link then:

- `/` will follow the link
- `.` will look up properties

if the object is not a merkle link, both separators will look up for properties.

That's a part of the spec that is optional (I forgot to mark it as such)

merkledag/ipld.md
mildred: webservers will only redirect clients when dealing with a real underlying filesystem.

In any case, we need a way to know if the trailing object, if it is a merkle link, will be followed or not. This is a solution, and there are other solutions as well.

mildred:
> the unixfs paths being different

That's a part I missed and it created much confusion, sorry for that.

merkledag/ipld.md
```diff
@@ -26,12 +26,20 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
 
 ### What is a _merkle-path_?
 
-A _merkle-path_ is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
+A merkle-path is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and allows access of elements of the referenced node and other nodes transitively.
+
+Merkle paths aren't suited to be used in filesystem representations (fuse mounts, HTTP or FTP protocols) as they describe the underlying IPLD data structure. Their use in filesystems is howver well suited for debug purposes (like `/proc` on unix).
```
jbenet: when we say this, do we mean:
- these paths are not good to represent _files_ because they allow access to the raw structures underneath
- or, these paths are not good to be accessed via the web or the filesystems, because there are UI problems.

i do want to be able to use these paths via HTTP to inspect the underlying data structures. but it's ok if i cannot get a proper file or directory representation out of it.

the latter part of this statement makes me somewhat more at peace with any of:
- disallowing _access of_ link properties via the paths
- disallowing transparent dereferencing without the use of `$obj/link/`
- separating "the link" and "the link properties" (same as the previous one in a way)

---

i think what we really need here is more concrete datastructure examples, and see how the pathing would want us to get to the stuff. maybe this will show us which solution is the best and guide our thinking.

perhaps we could write:
- an fs example (directories, sharded files, file attributes in dir entries)
- a version control example (commit histories)
- a social network example (users, messages, relationships) (maybe something like a simple version of foaf)
- a crypto network example (keys, derived keys, signatures, certificates, attestations) (i'll write this one at least)
- a blockchain example (block histories, transactions, etc)

jbenet:
> That's a part I missed and it created much confusion, sorry for that.

not at all, they look the same right now, which is definitely confusing. we could:
- leave `/ipfs/...` for unixfs
- use `/ipld/...` for raw ipld data.

not sure. in the end i still have trouble with the unixfs--ipld dichotomy. im not sure how to get unixfs to play well with other things (like commits). it is likely that:
- we will have a `1 + N` datastructure and `1 + N` path world (1 for ipld and N for every datastructure)
- or a `1 + N` datastructure but `1 + 1` path world (1 for ipld and 1 for the derived datastructs together).

the latter is hard but if we can get it it would be much easier to reason about. will be very difficult to think of different datastruct-specific paths :/

jbenet:
i do indeed recognize that there is value to try to get everything to be one path system, for example, give up on link properties and make this:

```json
{
  "cat.jpg": {"link": "Qmcatjpg..."},
  "foo": {"link": "Qmfoo..."},
  "@attrs": {
    "cat.jpg": {
      "mode": 777,
      "owner": "jbenet"
    }
  }
}
```

merkledag/ipld.md
mildred: What I meant is **these paths are not good to represent files because they allow access to the raw structures underneath**. That's why I took the analogy of the `/` proc filesystem which is not good for storing user files, but is still valuable as a filesystem. I might not have been clean enough, it's late at night in France and I need sleep :-)

There might be UI problems as the `.` separator is not understood as such by browsers for example. I don't think this will be a real problem though.

mildred:
Your example works well because it represents a file structure that is well represented with unix-like paths. But what if we try to represent anything else. What if keys contain binary data ? You need an escape mechanism.

Or perhaps I'm not thinking strainght as i am tried. In any case, the path definition is not impacting (or should not impact) the IPLD format in itself. So perhaps that's an issue that can be resolved later?

mildred:
I rephrased the paragraph [you commented here](#discussion_r49253782) to remove ambiguities. You talked about examples, but they are already there at the end of `ipld.md`.

If you want to unify every paths under the same prefix, how would that work in practice? For example, how will /ipns/ be managed (or am I lagging behind old concepts)?

mildred:
>     {
>       "cat.jpg": {"link": "Qmcatjpg..."},
>       "foo": {"link": "Qmfoo..."},
>       "@attrs": {
>         "cat.jpg": {
>           "mode": 777,
>           "owner": "jbenet"
>         }
>       }
>     }

here, you won't be able to access both `@attrs` and `cat.jpg` using the same path mechanism, unless you start telling people they can't name a file `@attrs`.

I don't think you can force people to use the same path for everything. It's not practical. Some applications might need some things not provided by default.

jbenet:
@mildred I've made https://github.com/ipfs/ipld-examples to try and resolve this question. So far i've only made two examples `unixfs` and `post`. take a look, i changed things up somewhat after giving more thought. I'm still not decided, but i'm seeing how horrible some of those options are.

could you please double check my work? im not sure i got everything right-- I may have messed up some of them, given that there are things you mentioned were problems but i didnt run head into them. (escaping `.` when using it as a delimiter, for example. may be that this _could_ be a non issue, but anyway, its separate).

I think so far, my favorites are (8), (4), and (5). -- the other feel odd or are very, very confusing.

Also, see the `unixfs` pathing-- i made it a separate thing there too, and i am more convinced this is the right thing to do.

echo-oddly:
I think `.` should not be used to do traversal. I think `\.` should. This way there is no need to escape for key names such as `notes.txt` which I think would occur more often.

mildred:
I updated this branch to `ipld-spec` and added my understanding of the path mechanism (8)

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
 
 ### What is a _merkle-path_?
 
-A _merkle-path_ is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
+A merkle-path is a unix-style path (e.g. `/a/b/c/d`) which initially dereferences through a _merkle-link_ and allows access of elements of the referenced node and other nodes transitively.
+
+_Merkle-paths_ aren't suited for using them in a general purpose filesystem because it introduces many restrictions on file names. However, it can be used to work on special purpose filesystems. It can be compared to the `/proc` filesystem on unix computers or HTTP Web APIs where the allowed paths is restricted.
```
jbenet: I think we should ditch this paragraph -- it's not totally accurate, FSes can be implemented, just the hoops get ugly (`dirA/@link/dirB/@link/fileC/data`), and we want something cleaner. I think the unixfs spec on top of IPLD can discuss the choices there.

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
...
+
+General purpose filesystems are encouraged to design an object model on top of IPLD that would be specialized for file manipulation and have specific path algorithms to query this model.
```
jbenet: i think we can keep this, and it expresses enough.

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
...
+
+### How do _merkle-paths_ work?
+
+A _merkle-path_ is a unix-style path which initially dereferences through a _merkle-link_ and then follows _named merkle-links_ in the intermediate objects. Following a name means looking into the object, finding the _name_ and resolving the associated _merkle-link_.
 
 For example, suppose we have this _merkle-path_:
 
 ```
-/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d
+/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k/a/b/c/d/
```
jbenet: my instinct is to keep "no trailing slashes". is there a problem still with trailing slashes?

also, we want to add the changes we decided on, namely using `/` and (`//` or `@link`)

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
...
 ```
 
 Where:
 - `ipfs` is a protocol namespace (to allow the computer to discern what to do)
 - `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` is a cryptographic hash.
 - `a/b/c/d` is a path _traversal_, as in unix.
-- this link traverses five objects.
 
-Suppose also that this path points to the object `{ "hello": "world" }`.
+Path traversal can either happen inside a single IPLD object, or can happen between objects throught *merkle-links*. The simple rule is that in case the traversal is possible in the same IPLD object, *merkle-links* should not be followed.
```
jbenet: "objects through *merkle-links*"

jbenet: Are we being _strict_ about always requiring `//` to traverse **merkle-links**? (that's what (8) called for)

--- 

If so, let's say: 

> Paths traversals are divided into two kinds:
> - **in-object traversals** traverse maps within the same object, and is denoted with `/`
> - **cross-object traversals** traverse across objects, _resolving through **merkle-links**_, and is denoted with `//` or with `/@link/`.
>
> We divide the traversals strictly, to avoid ambiguity in accessing properties within a **merkle-link** map itself. This is not transparent resolution, and thus a path reveals the objects it traverses. For example, `a/b//c/d//e` traverses across 3 objects.`
>
> We use `//` or `/@link/` for **cross-object traversals** depending on the filesystem implementation. For example, in unix filesystems, double slashes (`//`) are meaningless and often cleaned into a single slash (`/`). In such a case, the use of `/@link/` is required to traverse links.

---

If not, let's say:

> Paths traversals can be **in-object traversals**, which traverse maps within the same object, and **cross-object traversals**, which traverse across objects, _resolving through **merkle-links**_. 
>
> We use `/` to transparently traverse inside a single IPLD object or traverse across multiple. To avoid potential ambiguity, we MUST use **in-object traversals** wherever possible. A single slash (`/`) ALWAYS traverses **in-object** first, and **cross-object** otherwise .  A double slash (`//`) or `/@link/` ALWAYS traverses **cross-object**. For example, **merkle-links** can themselves carry properties and sub-maps. When `/` path traversals are ambiguous, they default to **in-object** (the local operation). In that case, we must use `//` or `/@link/` to traverse **cross-object**.

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
...
+
+- Use the special path component `/@link/` instead of a simple path separator. That also signifies the path needs to dereference the *merkle-link*
+
+As a consequence, `@link` keys that are not *merkle-links* cannot be referenced in *merkle-paths*.
+
+#### Examples
+
+The IPLD object with the hash `QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k` contains:
+
+    a:
```
jbenet: maybe show it as: 

```
> ipld cat --fmt yml QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k
a: 
...
```

jbenet: (ehhhh maybe not. shrug)

merkledag/ipld.md
```diff
@@ -32,51 +32,73 @@ Objects with merkle-links form a Graph (merkle-graph), which necessarily is both
...
+
+And the object `QmQmkZPNPoRkPd7wj2xUJe5v5DsY6MX33MFaGhZKB2pRSE` contains:
+
+    name: "third"
+
+An example of the paths:
+
+- `/ipfs/QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT/a/b/c` will only traverse the first object and lead to string `d`.
+- `/ipfs/QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT/a/b//c` will traverse both objects and lead to the string `e`
+- `/ipfs/QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT/a/b/@link/c` is equivalent
```
jbenet: is equivalent (will traverse both objects and lead to the string `@`)

merkledag/ipld.md
jbenet: @mildred i may be ok with "not strict", but i'd like to confirm it wont cause trouble to be lenient?

otherwise, we may have to be strict.

merkledag/ipld.md
mildred: Ok, as long as I can store a file named `@link` on my unixfs filesystem (which would at one point in some future would become my root filesystem I'd imagine).

merkledag/ipld.md
mildred: Right

merkledag/ipld.md
mildred: No problem, I'll fix that

merkledag/ipld.md
mildred: I wrote the non strict version on the spec. The only problem I could see with not being strict is that if we expect a property to be there on the same object but it is absent and this causes a cross object traversal.

mildred: > To avoid potential ambiguity, we MUST use in-object traversals wherever possible.

I'd rather say the opposite: To avoid potential ambiguity, we MUST use `//` or `/@link/` traversals wherever possible.

jbenet:
@mildred things look good. I'm good to leave the cases there for now, and decide on the lenient vs strict down the road. we can implement strict for now and see how it goes?

jbenet:
I'm good to merge as is, and continue from there.

jbenet:
@mildred lmk if you are ready too, and i'll merge. else what else to do?

mildred:
I'm ok with this as well.

PR #64, opened 2016-01-10
Rename Merkle links key from mlink to link

mildred:
Merges to #37

Replaces `mlink` occurrences with `link` with a paragraph to explain how merkle-links are formed. [Original comment](/ipfs/specs/pull/37#discussion_r45558988) and [follow up](/ipfs/specs/pull/37#issuecomment-169775469)

### DIFF
--- a/merkledag/ipld.md
+++ b/merkledag/ipld.md
@@ -20,6 +20,10 @@ A _merkle-link_ is a link between two objects which is content-addressed with th
 - **Cryptographic Integrity Checking**: resolving a link's value can be tested by hashing. In turn, this allows wide, secure, trustless exchanges of data (e.g. git or bittorrent), as others cannot give you any data that does not hash to the link's value.
 - **Immutable Datastructures**: data structures with merkle links cannot mutate, which is a nice property for distributed systems. This is useful for versioning, for representing distributed mutable state (eg CRDTs), and for long term archival.
 
+A _merkle-link_ is represented in the IPLD object model by a map containing a key `link` which value is the actual link. When dereferencing the link, the map itself is to be replaced by the object it points to.
+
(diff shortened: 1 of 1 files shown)


jbenet:
this LGTM. though do we want `link` or `@link` to signify importance, allow people to use `link` for other things (and `\@link` if they need to).

mildred:
I didn't understand what you said about `@link` and what does it mean to have an important link.

jbenet:
Sorry, let me rephrase: do we want `link` or `@link`? the `@` helps signify importance, as it is a clearly distinguished value, and we can do the directives + escaping. this also allows people to use the value `link` themselves without creating links accidentally. 

(i myself am leaning towards directives + escaping)

mildred:
If you want to add escaping, we can either let the application do the escaping on their own, or handle the escaping ourselves. Escaping implies that we will be able to provide the application with either the escaped (original) version, or the unescaped (with link information removed) objects. If the application wants to look at links, it will not be able to do so in the unescaped version the `@link` property will be removed). The application will also have to specify which version it wants and this may create confusion. *"Which version should I get?"*

I'd be tempted to say it doesn't matter if the application create links unknowingly. We'll just never use them. So let's use `link`, that's fine.

If we want to provide an easy way to avoid that, just use `@links` but let the application itself do the escaping. In any case, we probably don't want to bother with escaping in IPLD core. We'll just provide a library that the application would be free to use or not.

jbenet:
Very well thought out. thanks

> If we want to provide an easy way to avoid that, just use @links but let the application itself do the escaping. In any case, we probably don't want to bother with escaping in IPLD core. We'll just provide a library that the application would be free to use or not.

Yep totally agreed here, i think this ^ is the right choice.

jbenet:
thanks @mildred

PR #65, opened 2016-01-11
WIP: IPFS API (core+http+cli)

diasdavid:
Upgraded the doc to reflect and indicate how we are organising our API specs. Moved from a `level` org to a `system` org (calling something level 1 and 2, when inside level 1 there are extensions and optional features made things a tad confusing).

Currently, the IPFS Core API (the programmatic interface) is still kind in flux (and that is ok), so no need to pressure ourselves into locking it down. The IPFS CLI happens as needed and as the system evolve. The only API spec that needs to get more stable, is the HTTP API as a lot of other libraries and applications depend on it.

Currently, the HTTP API is a mix of RPC and RESTful (it really isn't either), which makes it weird to assess. Having a RPC API gives us the chance of layering it over any other transport, while not having a RESTful one makes it weird for Web Apps libraries.

Let's continue the endeavour of documenting the current HTTP API through Apiary (here http://github.com/ipfs/api) so that then we can migrate through PR and reviews

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -1,14 +1,36 @@
-# IPFS API Spec
+IPFS API Specifications (core + http + cli)
+===========================================
 
-> Official spec of the APIs that ipfs implementations must conform to, as well as test suites to ensure they do.
+> This document presentes the specifications for the IPFS APIs, namely: 'core', the programmatic interface, 'http', a networked API interface over HTTP and 'cli', the command line interface to interact with IPFS.
 
-> **WARNING** This is a work in *progress*. There be dragons.
+Authors:
 
-The IPFS HTTP API Spec, in Apiary format, can be found in [ipfs/http-api-spec](//github.com/ipfs/http-api-spec).
+- David Dias
+- Richard Littauer
 
-The IPFS API has two levels. The first level is a general description based on function signatures and the second one is the transport specific description.
+Reviewers:
 
-* [Level 1 Spec - Transport Agnostic](level1.md)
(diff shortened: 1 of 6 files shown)


lgierth:
> mix of RPC and RESTful (it really isn't either)

I agree that it's very low in the scale of RESTfulness, why call it RESTful at all though? The explicit goal is to mirror the CLI as much as possible.

api/README.md
```diff
@@ -1,14 +1,36 @@
...
+
+> **This spec is a Work In Progress (WIP).**
+
+# Organization of this document
+
+This RFC is organized by chapters described on the *Table of contents* section. Each of the chapters can be found in its own file.
+
+# Table of contents
+
+- [%N% Introduction]()
```
Dignifiedquire: what is `%N%`?

diasdavid: It is just a placeholder for the future number of the chapter

api/README.md
```diff
@@ -1,14 +1,36 @@
-# IPFS API Spec
+IPFS API Specifications (core + http + cli)
+===========================================
 
-> Official spec of the APIs that ipfs implementations must conform to, as well as test suites to ensure they do.
+> This document presentes the specifications for the IPFS APIs, namely: 'core', the programmatic interface, 'http', a networked API interface over HTTP and 'cli', the command line interface to interact with IPFS.
```
RichardLitt: presentes >> presents

RichardLitt:
Added one spelling note, otherwise, looks good to me.

PR #75, opened 2016-02-11
Application document to register CBOR tag to IANA

mildred:
In relation to #61, this is the application document to register the CBOR tags.

### DIFF
--- a/merkledag/cbor_tag_merkle_link.md
+++ b/merkledag/cbor_tag_merkle_link.md
@@ -1,0 +1,51 @@
+Merke Links
+===========
+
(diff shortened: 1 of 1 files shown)


mildred:
registration is here: http://www.iana.org/form/protocol-assignment

davidar:
If IANA considers this to be too application-specific, we could rephrase it as "tagged opaque pointers", which could be utilised for other purposes in other applications. Otherwise LGTM :)

diasdavid:
Let's move this doc to IPLD. Created a issue and tagged @nicola for it https://github.com/ipfs/specs/pull/75

PR #80, opened 2016-02-23
Update ipld.md

wasserfuhr:
fixed typo


diasdavid:
Thank you @wasserfuhr, it seems that we moved the spec before merging your PR. Now the IPLD spec lives at: https://github.com/ipld/specs

PR #81, opened 2016-02-28
Use absolute link to github

clkao:
So they are actually clickable in https://github.com/ipfs/specs/tree/ipld-spec/merkledag

### DIFF
--- a/merkledag/README.md
+++ b/merkledag/README.md
@@ -1,11 +1,11 @@
 # The merkledag
 
 
-Authors: [Juan Benet](github.com/jbenet)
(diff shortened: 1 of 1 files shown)


Kubuxu:
It would be better to use protocol less version IMO, so `//github.com/...`.

clkao:
There are a few other places (including top level README) using https. I suppose we can do a cleanup afterwards for all those places.

RichardLitt:
clkao: sounds like a good idea to me. Let me know if you want to do that, or want help. For now, I'll just merge this.

PR #83, opened 2016-03-09
IPFS Node Init: 1st Draft

noffle:
Heya! This aims to document how to initialize an IPFS node, based on the de facto code in go-ipfs. I'd love reviews & comments in particular from

- @whyrusleeping: for review of technical correctness
- @diasdavid: for feedback from the perspective of someone unfamiliar with init who'd be implementing it
- @richardlitt: for English & technical proof-reading

### DIFF
--- a/api/internals/init.md
+++ b/api/internals/init.md
@@ -86,13 +86,11 @@ The default bootstrap peers form the following config tree:
 
 #### Version
 
-This is the `"Version"` config subtree, which is a set of static values *except* for `"Current"`, which evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`.
+This is the `"Version"` config subtree, which is just the key `"Current"`, which
+evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`.
 
 ```json
 "Version": {
-  "AutoUpdate": "minor",
-  "Check": "error",
-  "CheckPeriod": "172800000000000",  // 48 hours
   "Current": "v0.4.0-dev"
 }
 ```
@@ -105,10 +103,10 @@ There are some other various static values that also must be included in the con
 // setup the node's default addresses.
 // Note: two swarm listen addrs, one tcp, one utp.
 "Addresses": {
-	"Swarm": {
+	"Swarm": [
 		"/ip4/0.0.0.0/tcp/4001",
 		"/ip6/::/tcp/4001"
-	},
+	],
 	"API": "/ip4/127.0.0.1/tcp/5001",
 	"Gateway": "/ip4/127.0.0.1/tcp/8080"
 },
--- a/api/internals/init.md
+++ b/api/internals/init.md
@@ -87,7 +87,7 @@ The default bootstrap peers form the following config tree:
 #### Version
 
 This is the `"Version"` config subtree, which is just the key `"Current"`, which
-evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`.
+evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`).
 
 ```json
 "Version": {
@@ -150,7 +150,7 @@ If the flag `--empty-repo` was provided, this step should be skipped.
 
 Otherwise, add the `init-doc` directory to the repo and pin them.
 
-Finally, write
+If successful, write
 ```
 to get started, enter:
 
--- a/api/internals/init.md
+++ b/api/internals/init.md
@@ -1,0 +1,207 @@
+# IPFS Node Init Specification
+
+> This is the specification for the steps performed when a new IPFS node is initialized.
+
+*Note that initializing an IPFS node differs from initializing an IPFS repo. For more on repo initialization, read the [Repo specification](https://github.com/ipfs/specs/tree/master/repo).*
+
+## Status
+
+First draft.
+
+For now, [`ipfs/go-ipfs/cmd/ipfs/init.go`](https://github.com/ipfs/go-ipfs/blob/master/cmd/ipfs/init.go) is the source of truth.
+
+## Flags
+
+There are some optional flags that may be supplied to `ipfs init`:
+
+```
+ -f|--force: whether to overwrite an existing repo config. Defaults to false.
+ -e|--empty-repo: whether to add+pin ipfs info help files to the repo. Defaults to false.
+ -b|--bits: # of bits to use in the generated RSA private key. Defaults to 2048.
+```
+
+## Initialization Sequence
+
+### 1. Verify Initial Conditions
+
+Before initialization can begin, the following must be true:
+
+1. The above input flags are parsed and validated.
+2. The [config file](https://github.com/ipfs/specs/tree/master/repo#config) has not been locked by an active daemon process.
+3. The repo root is writeable.
+4. The repo either doesn't exist and hasn't been initialized, or does already exist but the `--force` flag is supplied.
+
+If all of these are satisfied, output `"initializing ipfs node at %s\n", repoRoot` to standard out and proceed.
+
+### 2. Generate Repo Configuration
+
+These are the steps to producing the default [config tree](https://github.com/ipfs/specs/tree/master/repo#config). The actual output format will depend on the repo backend (i.e. likely JSON if fs-repo). Check the specification for the relevant backend if unsure.
+
+The following sections describe the various subtrees in the config and how to generate them. For simplicity and consistency, JSON will be used.
+
+#### Identity
+
+The RSA keypair must be >= 1024 bits.
+
+Before beginning key generation, write `"generating %v-bit RSA keypair...", nbits` to standard out. (no newline)
+
+Generate the keypair. If successful, append `"done\n"` to standard out.
+
+The private key must be formatted as the base64 encoding of its bytes. It is currently stored unencrypted.
+
+The PeerID must be formatted as
+
+```json
+Base58.Encode(Multihash('SHA2-256', PublicKey.Bytes))
+```
+
+Resulting in the following config tree entry:
+
+```json
+"Identity": {
+	"PeerID": "<PeerID>",
+	"PrivKey": "<PrivKey>"
+}
+```
+
+Finally, write `"peer identity: %s\n", ident.PeerID` to standard out.
+
+#### Bootstrap Peers
+
+The default bootstrap peers form the following config tree:
+
+```json
+"Bootstrap": [
+  // mars.i.ipfs.io
+  "/ip4/104.131.131.82/tcp/4001/ipfs/QmaCpDMGvV2BGHeYERUEnRQAwe3N8SzbUtfsmvsqQLuvuJ",
+
+  // neptune.i.ipfs.io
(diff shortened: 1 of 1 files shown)


api/internals/init.md
```diff
@@ -0,0 +1,207 @@
...
+
+Finally, write `"peer identity: %s\n", ident.PeerID` to standard out.
+
+#### Bootstrap Peers
+
+The default bootstrap peers form the following config tree:
+
+```json
+"Bootstrap": [
+  // mars.i.ipfs.io
```
diasdavid: note that comments are not valid JSON (unfortunately)

api/internals/init.md
```diff
@@ -0,0 +1,207 @@
...
+  "/ip4/104.236.151.122/tcp/4001/ipfs/QmSoLju6m7xTh3DuokvT3886QRYqxAzb1kShaanJgW36yx"
+]
+```
+
+#### Supernode Routing
+
+The default supernode routing nodes (for delegated federated routing) form the following config subtree:
+
+```
+SupernodeRouting: {
```
diasdavid: I'm not aware of this being currently used of ir it is just a go-ipfs experiment. Might not be necessary to add this to the spec for now (avoiding confusion)

diasdavid:
This looks pretty sweet @noffle. Really throughout specification of the init process, perfect to be followed by other implementations. The building blocks to implement this in JS are there (peer-info, peer-id, ipfs-repo that works in Node.js and the browser), let me know if you encounter any obstacle :)

api/internals/init.md
noffle: Removed. :(

api/internals/init.md
noffle: Thanks -- I wasn't sure if it was used or not. Removed!

api/internals/init.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Version
+
+This is the `"Version"` config subtree, which is a set of static values *except* for `"Current"`, which evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`.
+
+```json
+"Version": {
+  "AutoUpdate": "minor",
+  "Check": "error",
+  "CheckPeriod": "172800000000000",  // 48 hours
```
whyrusleeping: those three values arent currently used. should either do something with them or remove them

api/internals/init.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Miscellaneous
+
+There are some other various static values that also must be included in the config tree:
+
+```json
+// setup the node's default addresses.
+// Note: two swarm listen addrs, one tcp, one utp.
+"Addresses": {
+	"Swarm": {
```
whyrusleeping: list instead of object here

whyrusleeping:
This looks good to me! thanks for documenting

api/internals/init.md
noffle: whoops -- thanks!

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+
+## Flags
+
+There are some optional flags that may be supplied to `ipfs init`:
+
+```
+ -f|--force: whether to overwrite an existing repo config. Defaults to false.
+ -e|--empty-repo: whether to add+pin ipfs info help files to the repo. Defaults to false.
+ -b|--bits: # of bits to use in the generated RSA private key. Defaults to 2048.
+```
```
jbenet: there's an important difference between:
- an ipfs node (logical idea of what an ipfs node is)
- a commandline tool that inits an ipfs node (a projection of the abstract concept into a concrete implementation, in this case a cli tool based one)

the upshot is that we may need two docs:
- one doc that describes initializing an ipfs node, namely might specify "options", but not "commandline flags". for example, `IPFS_PATH` env var would NOT belong here. (may be too abstract for this doc).
- a doc that describes how initializing an ipfs node happens through cli tools like go-ipfs and js-ipfs (more this doc)

jbenet: this is similar to how we have "a repo spec" (abstract) and an "fs-repo spec" (concrete)

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+
+### 2. Generate Repo Configuration
+
+These are the steps to producing the default [config tree](https://github.com/ipfs/specs/tree/master/repo#config). The actual output format will depend on the repo backend (i.e. likely JSON if fs-repo). Check the specification for the relevant backend if unsure.
+
+The following sections describe the various subtrees in the config and how to generate them. For simplicity and consistency, JSON will be used.
+
+#### Identity
+
+The RSA keypair must be >= 1024 bits.
```
jbenet: It does not have to be RSA at all. go-ipfs doesnt support more, but we will. note that:
- "this is how it happens now"
- "this is what we're going to"

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+```
+
+Resulting in the following config tree entry:
+
+```json
+"Identity": {
+	"PeerID": "<PeerID>",
+	"PrivKey": "<PrivKey>"
+}
+```
```
jbenet: this may assume too much of the config or how it's stored. it may not have to be this way in all implementations.

jbenet: to expand: it may be ok to specify it this way, but the tricky bit will be in getting agreement about how the config keys and structure MUST be represented. maybe just leave it this way and note that some other spec will describe how these keys are actually stored in the repo.

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+// setup the node's default addresses.
+// Note: two swarm listen addrs, one tcp, one utp.
+"Addresses": {
+	"Swarm": [
+		"/ip4/0.0.0.0/tcp/4001",
+		"/ip6/::/tcp/4001"
+	],
+	"API": "/ip4/127.0.0.1/tcp/5001",
+	"Gateway": "/ip4/127.0.0.1/tcp/8080"
+},
```
jbenet: this is specific to go-ipfs. does it belong in this spec? (maybe, maybe not, not sure)

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+
+"Ipns": {
+	"ResolveCacheSize": 128,
+},
+
+"Gateway": {
+	"RootRedirect": "",
+	"Writable": false
+},
+```
```
jbenet: all of these are specific to go-ipfs. do they belong in this spec? (maybe, maybe not, not sure)

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+
+### 3. Initialize the Repo
+
+The exact process will depend on the repo backend ([fs-repo](https://github.com/ipfs/specs/tree/master/repo/fs-repo), mem-repo, s3-repo) being used.
+
+If at this point the repo already exists, destroy it. We can do this safely because at this point either a) there is no pre-existing repo, or b) there is a repo, but `--force` was specified, giving permission for it to be wiped.
```
jbenet: will this _always_ be the case? there may be an init that does not blow out the object repo.

jbenet:
good idea to document all this! we need to:
- (at some point) make a spec for a "logical" node (abstract)
- (now) decide whether some of these things we want ALL implementations to do or not.

if we dont mean ALL impls to use it, only go-ipfs, i dont think it is very useful to have this separate from the go-ipfs codebase. if its go-ipfs specific, it should be represented by code and tests and maybe comments.

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+	"PeerID": "<PeerID>",
+	"PrivKey": "<PrivKey>"
+}
+```
+
+Finally, write `"peer identity: %s\n", ident.PeerID` to standard out.
+
+#### Bootstrap Peers
+
+The default bootstrap peers form the following config tree:
```
RichardLitt: Would be good to have a short writeup about what these are -- servers or nodes hosted by IPFS.

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+  "/ip4/178.62.158.247/tcp/4001/ipfs/QmSoLer265NRgSp2LA3dPaeykiS1J6DifTC88f5uVQKNAd",
+  "/ip4/178.62.61.185/tcp/4001/ipfs/QmSoLMeWqB7YGVLJN3pNLQpmmEk35v6wYtsMGLzSr5QBU3",
+  "/ip4/104.236.151.122/tcp/4001/ipfs/QmSoLju6m7xTh3DuokvT3886QRYqxAzb1kShaanJgW36yx"
+]
+```
+
+#### Version
+
+This is the `"Version"` config subtree, which is just the key `"Current"`, which
+evaluates to the IPFS program's current version literal (e.g. `"0.4.0-dev"`.
```
RichardLitt: Missing )

api/internals/init.md
```diff
@@ -0,0 +1,169 @@
...
+
+### 4. Add Default Assets
+
+The exact set of assets to include is the directory structure defined in [`go-ipfs/assets/init-doc`](https://github.com/ipfs/go-ipfs/tree/master/assets/init-doc).
+
+If the flag `--empty-repo` was provided, this step should be skipped.
+
+Otherwise, add the `init-doc` directory to the repo and pin them.
+
+Finally, write
```
RichardLitt: Two finally, right after each other. Remove one.

api/internals/init.md
noffle: It was my understanding that one could init a repo with go-ipfs, then launch js-ipfs and point it at the config created by go-ipfs. If this is true -- that configs ought to be compatible across implementations -- we need to have a very consistent format.

api/internals/init.md
noffle: Why is this specific to go-ipfs? I think I may be unclear on what the responsibilities of an IPFS node are. Don't all IPFS nodes have ports for connecting to the swarm? Is it not the case that IPFS nodes offer an API and gateway? If not, whose responsibility is that? I think I need some clarification on where the boundaries lie.

noffle:
> (at some point) make a spec for a "logical" node (abstract)

I sort of asked about this in another comment, but where are the points of differentiation between a logical node and [whatever this doc currently describes]?

> (now) decide whether some of these things we want ALL implementations to do or not.

Agreed.

> if we dont mean ALL impls to use it, only go-ipfs, i dont think it is very useful to have this separate from the go-ipfs codebase. if its go-ipfs specific, it should be represented by code and tests and maybe comments.

Agreed. I intend for this document to generalize to any IPFS implementation.

api/internals/init.md
noffle: Good catch.

api/internals/init.md
noffle: :+1:

noffle:
Thanks everyone.

Based on @jbenet's comments, it sounds like this doc needs to be more abstract: covering initialization on a generic IPFS node that may/may not have things like a CLI or RSA keypairs or even standard out.

noffle:
I've integrated this doc as a [comment](https://github.com/ipfs/js-ipfs/issues/42#issuecomment-190470930) for now, since it is too concrete to be a general "IPFS Node Init" specification. It however serves well enough as a "here's how to do init if you're writing an IPFS implementation that is very go-ipfs-like (unixfs, has a CLI).

PR #93, opened 2016-04-13
Added a "raw" tag for an edge node in the chunked file example.

jefft0:
This is in part to resolve #2053 where jbenet said "by extending IPLD to allow just raw data edge nodes, we can make #875 easy to implement".

### DIFF
--- a/ipld/README.md
+++ b/ipld/README.md
@@ -437,6 +437,7 @@ Split into multiple independent sub-Files.
     },
     {
       "@link": "QmAA1...",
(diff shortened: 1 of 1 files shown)


davidar:
I don't think this would work with the new links-without-properties. I've made some suggestions in #111 that would hopefully address this issue.

jefft0:
I can see IPLD has moved in another direction (which also will support raw leaf nodes). I'll withdraw and close this pull request.

PR #96, opened 2016-04-19
write down swarm dialing logic

whyrusleeping:
I thought i'd take a stab at describing the rather complicated dialing logic in a spec

### DIFF
--- a/libp2p/8-implementations.md
+++ b/libp2p/8-implementations.md
@@ -61,6 +61,7 @@ Current known implementations (or WIP) are:
 
 The swarm dialer manages making a successful connection to a target peer, given a stream of addresses as inputs, and making sure to respect any and all rate limits imposed. To this end, we have designed the following logic for dialing:
 
+```
 DialPeer(peerID) {
 	if PeerIsBeingDialed(peerID) {
 		waitForDialToComplete(peerID)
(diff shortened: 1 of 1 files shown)


RichardLitt:
The code should be indented or in ```.

diasdavid:
Was there anything added on https://github.com/ipfs/go-libp2p/pull/38 that wasn't described in these notes, @whyrusleeping ? Otherwise LGTM for merge :)

whyrusleeping:
@diasdavid no, i think the gist of it is here

libp2p/8-implementations.md
```diff
@@ -55,3 +55,68 @@ Current known implementations (or WIP) are:
...
+		} else {
+			rateLimitScheduleDial(peerID, addr)
+		}
+	})
+}
+
+// doDialAsync starts dialing to a specific address without blocking.
+// when the dial returns, it releases rate limit tokens, and if it
+// succeeded, will finalize the dial process.
+doDialAsync(peerID, addr) {
```
whyrusleeping: One 'bug' i just caught in writing more tests for this is that we shouldnt start dials if they have been cancelled. For example, I queue up 100 dials, and one in the first ten or so completes successfully we don't want to continue dialing the remaining 90 queued addresses for that peer.

cc @diasdavid

diasdavid: This finding goes along the lines of https://github.com/diasdavid/js-libp2p-swarm/issues/24

Issue #97, opened 2016-04-30
IPLD Pathing Redux


jbenet:
A week or so ago, @nicola @dignifiedquire @Stebalien and I discussed lots of IPLD things in detail. 

Among them, we discussed pathing again. One result here was that we revisited the "two pathing delimiters" and link-properties problem. @Stebalien and I got to something in the end, and many approve (including @dignifiedquire and @diasdavid). @mildred i would be very interested in your thoughts. And I believe this is more in line with your original thinking too, or at least things you and I considered.

In the meeting we recalled many observations from previous conversations, including:

- we want the format to be as absolutely simple as possible, requiring very little conceptual overhead
- pathing with two different delimiters gets complicated for users
- the `//` shorthand for `/@link/` increases conceptual complexity
- breaking apart into layers (Layers 2, 3, 4) helps address graph shaping problems
- the `{"@link": "hash"}` construction is getting around the problem that JSON does not allow representing a different value type natively (eg a link value). Other formats (CBOR, YML) do not have this problem.

#### Right now we have:

```js
{
  "mode": 0444,
} // hash = Qmfoo

{
  "foo": {
    "@link": "Qmfoo",
    "mode": 0755,
  }
}

> /foo
{
  "@link": "Qmfoo",
  "mode": 0755,
}

> /foo/mode
0755

> /foo/@link/mode
0444

> /foo//mode
0444
```

Recalled and new observations where

- The "link properties" problem is introduced by using an object to represent links.
- We use an object to represent links because other JSON formats (like EJSON) use objects like `{"$date": "2016-01-01"}` to represent other _Value Types_.
- The "link objects" (such as a _directory entry_ with a mode in a _directory_) are different from "link values". The current way to do things mixes these two.
- We'll already have to re-path in layer 4 or layer 3 to address things like `/foo/bar/baz` representing a unixfs directory traversal. (ie. the shape of the data, and without `//` or `/@link/` traversals)
- Users will have to use `/@link/` a lot all over the place.
- There is no way to address the _Link Value_ on it's own.

We found a very nice middle ground between #1 in https://github.com/ipfs/ipld-examples/ and the current way to do things. 

#### The proposal is to move to something like:

```js
{
  "mode": 0444,
} // hash = Qmfoo

{
  "foo": {
    "entry": {"&": "Qmfoo"},
    "mode": 0755,
  }
}

> /foo/entry
{"&": "Qmfoo"}

> /foo/mode
0755

> /foo/entry/mode
0444
```

[...]

PR #101, opened 2016-05-22
update ipld spec with new link fmt

jbenet:
This changes the spec to update the link fmt. Please review it carefully and suggest changes. This was a great simplification. thanks everyone.

See:
- https://github.com/ipfs/specs/issues/97
- https://github.com/ipfs/js-ipld/pull/17

### DIFF
--- a/ipld/README.md
+++ b/ipld/README.md
@@ -1,7 +1,9 @@
 IPLD -- the "thin-waist" merkle dag format
 ==========================================
 
-![](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
+![](https://img.shields.io/badge/status-draft-green.svg?style=flat-square)
+
+UPDATE: we re-drafted this spec to deal with links. We hope to re-finalize it shortly. Sorry for any inconvenience. This was an important change to do before implementations shipped.
 
 There are a variety of systems that use merkle-tree and hash-chain inspired datastructures (e.g. git, bittorrent, ipfs, tahoe-lafs, sfsro). IPLD defines:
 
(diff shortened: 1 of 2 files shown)


dignifiedquire:
There is still a paragraph referencing `@link`:

> As a consequence of using the @link path component to denote cross-object traversals, this becomes a reserved path component and makes it impossible to access arbitrary @link keys that are not otherwise merkle-links. Escaping can be used to render access to those keys possible if so desired.

dignifiedquire:
The section on `Escaping Algorithm` still refers to `@link`

ipld/README.md
```diff
@@ -23,11 +25,45 @@ A _merkle-link_ is a link between two objects which is content-addressed with th
...
+```js
+{
+  "foo": {
+    "bar": "/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k", // not a link
+    "baz": {"/": "/ipfs/QmUmg7BZC1YP1ca66rRtWKxpXp77WgVHrnv263JtDuvs2k"} // link
+  }
+}
 
-If an application wants to use the `@link` key for other purposes, the application itself is responsible to escape the keys in the IPLD object so that the application keys do not conflict with the `@link` key. When discussing application specific paths, it may be worth escaping all keys starting with `@` in case future versions of IPLD make use of other keys.
+**Object with pseudo "link object" at `files/cat.jpg` and actual link at `files/cat.jpg/link`**
```
dignifiedquire: This seems to be new to me and problematic when building it into the resolver. I was under the impression key like `"link"` would not have a special meaning, but with this you are effectively giving them one. 
Also this seems ambiguous to me, consider this object:

```js
{
  "files: {
    "cat.jpg": {
      "coolLink": {"/": "/ipfs/Qm..AAA"},
      "link": {"/": "/ipfs/QmBBB"},
      "mode": 0755
    }
  }
}
```

then how should `files/cat.jpg` be resolved? Objects are not inherently ordered in JavaScript so we can not guarantee the same resolution if we say first come, first serve.

jbenet: @dignifiedquire no, there's nothing special going on. the `link` keyword here is not special at all, it just shows how someone might model a data structure that requires associating properties with a link. Ex:

```
> ipld get .../files/cat.jpg
 {
  "coolLink": {"/": "/ipfs/Qm..AAA"},
  "link": {"/": "/ipfs/QmBBB"},
  "mode": 0755
}

> ipld get .../files/cat.jpg/coolLink/
<resolves through coolLink>

> ipld get .../files/cat.jpg/link/
<resolves through link>
```

dignifiedquire: Okay great, then I just misunderstood

jbenet:
@dignifiedquire thanks, removed those `@link` refs.

PR #104, opened 2016-06-09
Expand the IPLD Abbreviation in first paragraph

xioustic:
This drove me nuts when I first read thru this spec, lol. I couldn't find the expansion anywhere else in the doc.


noffle:
Thanks for noticing and PRing this. :)

PR #103, opened 2016-06-09
Minor correction to example

adrianhopebailie:

PR #105, opened 2016-06-10
ipld/README: fix path example

chriscool:
The QmV76pUdAAukxEHt9Wp2xwyTpiCmzJCvjnMxyQBreaUeKT object has no key "e".
Its only key that gives "e" is "c".


chriscool:
Ooops, it looks like this duplicates PR #103 !

dignifiedquire:
Thanks

PR #123, opened 2016-07-06
Add multigram draft 1

lgierth:
Braindump related to packet switching (ipfs/notes#143).

### DIFF
--- a/multigram/README.md
+++ b/multigram/README.md
@@ -1,0 +1,96 @@
+# Multigram -- protocol negotiation and multiplexing over datagrams
+
+For multiplexing different protocols on the same datagram socket, multigram prepends a 1-byte header to every packet. This header represents an index in a table of protocols shared between both endpoints. This protocol table is negotiated by exchanging the intersection of the endpoint's supported protocols. The protocol table's size of 256 tuples can be increased by nesting multiple multigram headers.
+
+TODO: analyze the properties vs. other approaches (e.g. an identifier on every packet).
+
+# Packet layout
+
+```
+                  1               2               3               4
+   0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
+  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+0 |  Table Index  |                                               |
+  +-+-+-+-+-+-+-+-+             Variable-length data              +
+4 |                                                               |
+  +
+```
+
+## Protocol table
+
+The protocol table MUST be append-only and immutable. It MUST initially contain exactly one tuple:
+
+```
+0x00,/multigram-table/0.1.0
+```
+
+The `/multigram-table` protocol is used for appending to the shared protocol table.
+
+```
+                  1               2               3               4
+   0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
+  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+0 |     0x00      |                                               |
+  +-+-+-+-+-+-+-+-+           Operation as Multicodec             +
+4 |                                                               |
+  +
+```
+
+- Note how the Table Index field is set to `0x00`, selecting the `/multigram-table` protocol.
+- The Operation field MUST support at least the `/cbor/` multicodec format. It SHOULD support the `/protobuf/` and `/json/` formats.
+- In the future, an `/ipfs/` format can be used to resolve code and specification for supporting the format.
+
+## Protocol table operations / multigram-table
+
+Either endpoint can send `append` proposals, and the other endpoint will reply with the result, based on their own supported protocols.
+
+(1) Endpoint A sends a proposal to Endpoint B. A doesn't commit this proposal to its own view of the protocol table. It dismisses it right away and waits for B's reply.
+```
+0x00
+/json/
+{"0x01":"/foo/1.0.0","0x02":"/bar/1.0.0","0x03":"/baz/1.0.0"}
+```
+
+(2) Endpoint B forms the intersection of this proposal and its own supported protocols, appends to its own view of the protocol table, and replies.
+```
+0x00
+/json/
+{"0x01":"/foo/1.0.0","0x02":"/bar/1.0.0"}
+```
+
+(3) We can list the table by sending the other endpoint an empty proposal.
+```
+0x00
+/json/
+{}
+```
+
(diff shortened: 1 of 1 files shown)


lgierth:
I'd be grateful for pointers to prior art, and most of all people who review this, point out open questions, etc. :)

multigram/README.md
```diff
@@ -0,0 +1,96 @@
+# Multigram -- protocol negotiation and multiplexing over datagrams
+
+For multiplexing different protocols on the same datagram connection, multigram prepends a 1-byte header to every packet. This header represents an index in a table of protocols shared between both endpoints. This protocol table is negotiated by exchanging the intersection of the endpoint's supported protocols. The protocol table's size of 256 tuples can be increased by nesting multiple multigram headers.
```
whyrusleeping: Why not use a varint instead of a single byte?

lgierth: Parsing is not so much a problem, but the per-packet overhead should be constant -- for some cases, we'll want to inform the upper protocols about the overhead, so they can avoid fragmentation, or batch multiple small messages into one packet.

We could fit a larger table into to 2 bytes, but I expect that a connection with more than 255 protocols on the same layer will be extremely rare. These two bytes multiply the deeper you nest protocols, in IPFS we currently have 3 layers of multistream I think?

Does this make any sense?

lgierth: The maximum packet size would be bubbled up from the transports through all the protocol wrappers, e.g. `512 (UDP) - 1 (multigram) - 20 (cryptoauth) - 1 (multigram) - 12 (switch) - 1 (multigram) = 477`

whyrusleeping: hrm... okay. I see the importance of the single byte, but you do mention that you can 'expand' the table by nesting. Nesting seems more complicated than using a varint and informing other layers how much space they have left to consume

lgierth: Yes, with a good night's sleep, you're absolutely right.

lgierth: To explain what my concern was: I was thinking the varint width would vary on a packet-per-packet basis, while it'll really only change if you add protocols to the protocol table (i.e. change the multigram connection's state).

Kubuxu: I think varint will make packet handling more complex then it should be, you introduce offset handling and unframing that depends on the state of connection.
If the state goes desync for what ever reason getting it back to sync will be extremely hard.
Having constant header file is very nice feature.

Also 255 overlay protocols (that are negotiated on need basis should be more than enough, if need layering is still an option).

multigram/README.md
```diff
@@ -0,0 +1,96 @@
...
+Either endpoint can send `append` proposals, and the other endpoint will reply with the result, based on their own supported protocols.
+
+(1) Endpoint A sends a proposal to Endpoint B. A doesn't commit this proposal to its own view of the protocol table. It dismisses it right away and waits for B's reply.
+```
+0x00
+/json/
+{"0x01":"/foo/1.0.0","0x02":"/bar/1.0.0","0x03":"/baz/1.0.0"}
+```
+
+(2) Endpoint B forms the intersection of this proposal and its own supported protocols, appends to its own view of the protocol table, and replies.
```
Kubuxu: The problem I see with this is how we react when two sides start appending at the same time?
Also as we will be using it over bare UDP we should deal with datagram loss and reordering.

multigram/README.md
```diff
@@ -0,0 +1,96 @@
...
+   0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7 0 1 2 3 4 5 6 7
+  +-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+0 |     0x00      |                                               |
+  +-+-+-+-+-+-+-+-+           Operation as Multicodec             +
+4 |                                                               |
+  +
+```
+
+- Note how the Table Index field is set to `0x00`, selecting the `/multigram-setup` protocol.
+- The Operation field MUST support at least the `/cbor/` multicodec format. It SHOULD support the `/protobuf/` and `/json/` formats.
```
Kubuxu: I would love to see bencode added to the format list as for it simplicity. CBOR, protobuf and JSON definitions are all tenths of pages long. In contrast bencode is both short to define and to write parser for (most implementations are sub 500 LOC). 

http://jonas.nitro.dk/bittorrent/bittorrent-rfc.html#anchor7

lgierth: Yeah we should totally discuss which formats MUST and SHOULD be included, and I agree that bencode should be in there somewhere.

multigram/README.md
lgierth: The endpoint should respond with some kind of error if it gets a proposal with conflicts with its own pending proposal. So these two would cancel each other out.

multigram/README.md
lgierth: Mh yep we're trading off between a complication of header parsing, and a complication of protocol semantics. We should decide whether we favor performance (int8 field and nested multigrams) over a simpler protocol (varint and no nesting).

We should figure out exactly *how complicated* nesting is (make experiments), and also *how heavy* parsing a varint is vs. an int8 (benchmarks). My gut feeling is that we should probably go with an int8, because we'll have to do this for *every* packet, and multiple times per packet in many use cases.

Not-so-appealing idea: a varint degrades to an int8 nicely, so there might a way to leave it up to implementations.

lgierth:
### todo

- [ ] decide int8 vs. varint
  - [ ] nesting experiment
  - [ ] int8 vs. varint benchmarks
- [ ] elaborate on the table operations
  - [ ] the exact operations
  - [ ] the reason for table operations (packets are small, so in bigger use cases, the table has to be fragmented)
- [ ] success responses
- [ ] error responses
- [ ] ability to reset the table?

RichardLitt:
This should be added to the now stubbed out https://github.com/multiformats/multigram, and work should continue there.

lgierth:
Will open another PR at https://github.com/multiformats/multigram -- plenty of great changes resulted from the discussions had in Lisbon.

ben221199:
Is multigram still a thing or should it be deprecated? In that case, the https://github.com/multiformats/multigram should be archived and the README of https://github.com/multiformats/multiformats and the website should be updated. The last activity was around 2016/2017.

rvagg:
@marten-seemann @mxinden thoughts? If we're not using it I agree with @ben221199 that we should make it clear they're not an ongoing concern.

mxinden:
Thanks for the ping @rvagg. I have not been aware of the multigram project.

With QUIC and https://github.com/libp2p/specs/pull/349 I don't see us pushing for a datagram based protocol negotiation protocol any time soon. Thus I am fine with marking this effort deprecated.

rvagg:
archived it, thanks for raising it @ben221199

ben221199:
Maybe I can also make a pull request for https://github.com/multiformats/website.

marten-seemann:
Sorry for the late reply. I agree with closing / archiving.

Datagram transports are probably better implemented on top of QUIC, which provides native datagrams (encrypted, congestion controlled, forward-secure) at the transport level.

PR #124, opened 2016-07-16
Native English speaker fixes

JesseWeinstein:
This shouldn't change the meaning at all; it's just minor copy-editing tweaks.

### DIFF
--- a/overviews/implement-api-bindings.md
+++ b/overviews/implement-api-bindings.md
@@ -10,16 +10,16 @@ Sections:
 
 ## IPFS Types
 
-IPFS uses a set of type value that is useful to enumerate up front:
+IPFS uses a set of type values that are useful to enumerate up front:
 
-- `<ipfs-path>` is unix-style path, beginning with `/ipfs/<hash>/...` or `/ipns/<hash>/...` or `/ipns/<domain>/...`.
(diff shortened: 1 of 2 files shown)


overviews/implement-api-bindings.md
```diff
@@ -171,5 +168,4 @@ foo
 
 ```
 
-Which produces: http://gateway.ipfs.io/ipfs/QmNtpA5TBNqHrKf3cLQ1AiUKXiE4JmUodbG5gXrajg8wdv
-
+which produces: http://gateway.ipfs.io/ipfs/QmNtpA5TBNqHrKf3cLQ1AiUKXiE4JmUodbG5gXrajg8wdv
```
RichardLitt: Why the lowercase?

RichardLitt:
Looks good to me. One small nitpick.

overviews/implement-api-bindings.md
JesseWeinstein: Because it's part of the sentence (far) above, that begins: "Basically, we end up with HTTP requests like this:" rather than a separate sentence. If we wanted a separate sentence, something like: "Such a set of requests produces: http://gateway..." would work.

RichardLitt: Works for me.

RichardLitt:
Thanks Jesse.

PR #125, opened 2016-07-17
Copyediting of the libp2p spec

JesseWeinstein:


RichardLitt:
Who originally wrote this? Because they should review, this is great feedback on English writing. 

Also, @JesseWeinstein, can you collapse these commits?

lgierth:
cc @diasdavid

JesseWeinstein:
I'm glad to combine these commits into fewer -- would you like just one commit, or?

JesseWeinstein:
OK, I've now squashed all the commits into one. Note that there are a few places where I did a bit of rewording, which needs to be reviewed to make sure I got the meaning correct.

RichardLitt:
Thanks. @diasdavid This is gtg, but you should read it if you like!

diasdavid:
Thank you @JesseWeinstein ! Really appreciate the edits. Hope that that reading spec was helpful :)

JesseWeinstein:
It was, thank you!

PR #134, opened 2016-09-13
Move libp2p to libp2p/spec

RichardLitt:
I have moved libp2p to github.com/libp2p/spec, keeping all of the commit history there. Some justification given in libp2p/website#8, but mostly this is clear; we have a libp2p organization, and the spec should be there, not in IPFS, now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ The specs contained in this repository are:
 - [overviews](/overviews) - quick overviews of the various parts of IPFS
 
 **Networking layer:**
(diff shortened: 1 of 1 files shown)


diasdavid:
needs rebase, otherwise LGTM

RichardLitt:
Looks like you copied my efforts before. :P Shouldn't we just delete the libp2p folder? No need for it to have a redirect README.

diasdavid:
@RichardLitt it helps people that were linking to our ipfs/specs, it is kind of an "archaic" redirect :)

Issue #138, opened 2016-09-30
Spec of ipfs: and fs: URI schemes 


Kubuxu:
As a follow up of https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-157478515 we should have clear specs for those URI schemes, in including `web+` prefixed ones.

lidel:
Year later, I think this issue should be renamed to have _"and `dweb:`"_ :)

Ref. https://github.com/ipfs/specs/pull/139#issuecomment-339935336, https://github.com/ipfs/specs/pull/152#issuecomment-284628862

lgierth:
Thanks! Good catch

da2x:
Any update on this?

lidel:
See current consensus at:
- https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md

Mid-term plan is to move it to docs.ipfs.io: https://github.com/ipfs/docs/issues/91#issuecomment-423851456

fulldecent:
This issue can please also include documentation of URLs starting with https://ipfs.io/ipfs/

As per  https://github.com/ipfs/specs/issues/307

PR #139, opened 2016-09-30
First draft of ipfs: URI spec, see #138

pawal:


lgierth:
Hey, thanks so much, and sorry for keeping this waiting so long. I'll take a proper look and provide in the next 2 weeks

diasdavid:
@nicola this should be on the IPLD org, correct?

lgierth:
> @nicola this should be on the IPLD org, correct?

Not sure, the fs: URI scheme doesn't seem coupled to IPLD. It sits at the root of the path namespace (it *defines* the path), and IPLD is just in `/ipfs` or `/ipld`.

nicola:
Although it does not belong to ipfs either..
IPFS in the browser?

pawal:
Not sure exactly what you guys are referring to when you say that which sub-project this belongs to. The spec does not refer to either fs: or ipld.

Is there something I should change in the document, or are you thinking of where to publish the spec?

lgierth:
Hey @pawal, I'm just getting back into the addressing scheme discussion and it looks like we're pretty close to consensus on the ipfs:// and fs: schemes, their semantics, and the upgrade path. Which is awesome, given the complexity of the issues, and the constraints \o/

Just want to let you know that I'm sorry we weren't very responsive to your PR and contribution (and even stumbled in here with unrelated stuff). Let's wrap up the other discussion and then update the IANA registration docs. I definitely wanna get these sent off :) It's just that wrapping up the addressing scheme discussion is a prerequisite.

From what I understand it'll likely be three documents in the end, for ipfs://, ipns://, and fs:. Do you know how strict IANA are when it comes to registering schemes? Is there a chance they will tell us to revisit everything and come back with just *one* scheme?

pawal:
Thank you for your update. Just let me know what changes you are planning.

In this case IANA makes us of expert reviewers for registration of new types, and several of them are friendly enough. There is no problem having ipfs and ipns now, and later come in with another URI registration. But we should probably not change ipfs or ipns too much if we want to register them. (Errata is ok.)

You can have a look here and read some of the non-RFC documents: http://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

Also read [RFC6920](https://tools.ietf.org/html/rfc6920) as background material, and reference it in your conclusions.

nichtich:
Any progress on this issue? I need to reference datasets via URIs so IPFS would only be an option if there is an URI scheme

lidel:
I think it is worth noting that a plan was created, see  [four stages of the upgrade path for path addressing](https://github.com/ipfs/specs/pull/152#issuecomment-284628862) and 

**tl;dr** 

Short-term, URLs:
> `ipfs://{hash}` → `http://127.0.0.1:8080/ipfs/{hash}`
> `ipns://{hash}` → `http://127.0.0.1:8080/ipns/{hash}`

Long-term, URI:
> `dweb:/ip[f|n]s/{hash}` → `http://127.0.0.1:8080/ip[f|n]s/{hash}`

pawal:
So this PR is essentially ok?

uri/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+Reviewers:
+
+* * *
+
+# Abstract
+
+For different applications to handle ipfs references there need to be URIs registered. In order to for applications to understand how to handle the references, the URIs needs to have specificaitions. The generic syntax for URIs are described in [RFC3986](https://tools.ietf.org/html/rfc3986).
+
+The consensus of which URIs need to be specified can be found here: https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-157478515
```
lidel: I think this was updated and superseded by @lgierth's comment at https://github.com/ipfs/specs/pull/152#issuecomment-284628862, no?

uri/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+# Status of this spec
+
+![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+Note: These specifications are currently in the wip phase, things are likely to change.
+
+# Overview
+
+The discussion on what URI schemes that are needed is concluded [here](https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-157478515). The discussion lead to the consensus that these URI schemas are needed:
```
lidel: I would add https://github.com/ipfs/specs/pull/152#issuecomment-284628862 here too

uri/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+ * **canonical NURI** is a file system **path**. Does not need a URI specification. Examples, where **Qm...** is a [multihash](https://github.com/multiformats/multihash):
+    * /ipfs/Qm...
+    * /ipns/Qm...
+    * /ipfs/Qm.../filename
+    * /ipns/Qm.../dir/filename
+
+ * **[ipfs://](./ipfs.md)** style URI that references IPFS content with the multihash and filepath.
+
+ * **[fs://](./fs.md)** style URI referencing ipfs-specific AND non-ipfs specific hash resolution mechanisms
```
lidel: According to https://github.com/ipfs/specs/pull/152#issuecomment-284628862 `fs://` URL is replaced by `dweb:` URI

uri/ipfs.md
```diff
@@ -0,0 +1,34 @@
+(last updated 2016-09-30)
+
+Resource Identifier (RI) Scheme name: ipfs 
+Status: provisional
+
+Applications/protocols that use this scheme name: ipfs
+
+Scheme syntax:
+   ipfs://ipfs/&lt;multihash&gt;
```
lidel: I think we just go with `ipfs://{cid}` these days

uri/ipfs.md
```diff
@@ -0,0 +1,34 @@
...
+   ipfs://ipfs/&lt;multihash&gt;/path
+   ipfs://ipns/&lt;multihash&gt;
```
lidel: ..and `ipns://{peerid}`  instead of `ipfs://ipns/{peerid}`

daviddias:
@pawal I was going to rebase this PR onto master, but found that it got created in your fork. Would you prefer:
- rebasing yourself
- creating a new branch from this repo so that I can then rebase it myself?

pawal:
I could perhaps do the rebase later this week. Thanks for the heads up.

pawal:
@daviddias Hi, I rebased my fork now. Not sure that everything worked out ok, but it looked fine to me.

lidel:
IIRC this PR was adding two files in `uri/` and now changes stuff all over the place.
@pawal  Try copying `uri/` somewhere safe, then delete local branch and recreate it from latest master + add `uri/` back and force-push.

pawal:
@lidel - ok, will resubmit. There has been some proposed changes in this PR, but I will redo my original PR and we'll start from there.

lidel:
Thank you! Let's continue in https://github.com/ipfs/specs/pull/222

PR #140, opened 2016-09-30
change r.Expires from time.Duration to time.Time

keks:
All the code treats it as a `time.Time` and that's the only thing that makes sense here.

I know it's just an example but good, correct examples help understanding the ideas and reasoning behind a spec.


keks:
"CR" please :)

PR #147, opened 2016-12-29
match description identifiers

atfornes:
change formating to serializing in Record System definition for consistency.


jbenet:
thnaks

PR #148, opened 2017-01-06
rough draft describing pubsub wire protocol 

whyrusleeping:
This is description of the current protocol *as it exists in the codebase now*. Please review with respect to this specs accuracy to the implementation, and leave comments about protocol changes for another PR.

### DIFF
--- a/pubsub/spec.md
+++ b/pubsub/spec.md
@@ -1,0 +1,164 @@
+# libp2p pubsub specification
+
+This is the specification for generalized pubsub over libp2p. Pubsub in libp2p
+is currently still experimental and this specification is subject to change.
+This document does not go over specific implementation of pubsub routing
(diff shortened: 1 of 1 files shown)


hsanjuan:
The RPC description seems clear enough to me.

Should it be part of the spec an explanation of how these RPCs are expected to circulate around, and how the different options affect that circulation? My impression after quick reading is that I know what the message format looks like, but still I have little idea of how the pubsub protocol works.

hsanjuan:
Also, thanks for dedicating some time for this!!

diasdavid:
@whyrusleeping you are awesome! keep the specs coming :) 

I'm moving this one to libp2p/specs, here: https://github.com/libp2p/specs/pull/7 Giving a clean up to the place :)

Issue #131, opened 2016-08-24
Feedback for the Suborigin working group


ghost:
@jbenet mentioned in Lisbon that _now_ is the time to direct feedback at the Suborigin working group [1]. In fact it might already be too late as their 31-Jul-2016 was the "recommendation" milestone date. Anyhow I wanna write up our use case and requirements and possible need for changes, and send it their way.

Let me know if this repo is not a good place for this.

[1] https://www.w3.org/2014/12/webappsec-charter-2015.html

lgierth:
> In fact it might already be too late as their 31-Jul-2016 was the "recommendation" milestone date. 

Wrong about this one. I'll try and figure out what the actual milestones for suborigin are. Their website lists Dec-2015 for the respective recommendation milestone, so that's clearly out of date.

lgierth:
I think Suborigins still fit our use case and I don't have pressing feedback right now. What's breaking the current suborigins implementation in go-ipfs is the recent requirement that identifiers be lowercase, in order to be able to construct RFC6454-compatible origins.

It looks like we'll need HSHCA here, a "multihash encoded in Base32 RFC 4648, downcased and stripped of its padding", which is fit to be used in hostnames and should yield RFC6454-compatible suborigin-names. Thanks @kyledrake :):)

```
LOWERALPHA = %x61-7A   ; a-z
suborigin-name = 1*( LOWERALPHA / DIGIT / "-" )   ; DIGIT as defined by RFC5234
```

---

Here's a reading list:
- editor's draft
  - https://w3c.github.io/webappsec-suborigins/
  - https://github.com/w3c/webappsec-suborigins
- RFC6454, the web origin concept: https://tools.ietf.org/html/rfc6454
- browsers
  - https://bugs.chromium.org/p/chromium/issues/detail?id=555117
  - https://bugzilla.mozilla.org/show_bug.cgi?id=1231225
- ipfs-firefox: https://github.com/lidel/ipfs-firefox-addon/issues/71
- a writeup: https://chloe.re/2016/09/01/sub-origins-an-introduction-to-suborigins/

lgierth:
cc @jbenet @diasdavid @dignifiedquire

dignifiedquire:
Thanks @lgierth for this! I will read through the links this week and comment afterwards.

Kubuxu:
We use base32 without padding in the datastore already so it shouldn't be hard to implement.

As we do that we should support both ipns and ipfs.

[...]

PR #152, opened 2017-03-03
Background on Address Scheme discussions (dweb, ipfs:// and more)

flyingzumwalt:
This still needs work, but it gives a starting point for people to clarify the discussion. Please submit comments and/or changes.

ref ipfs/ipfs#227 ipfs/in-web-browsers#4 ipfs/in-web-browsers#28 https://github.com/ipfs/specs/pull/139


flyingzumwalt:
@lgierth @jbenet @nicola @gozala @lidel @jefft0 and anyone else, please help me make this document accurate and complete.

addresses-scheme/readme.md
```diff
@@ -0,0 +1,72 @@
...
+> A minor reason is not having to force people to swallow N shemes (ipfs:// ipns:// ipld:// and counting), and instead use one that muxes.
+
+### ... but don't let it prevent pragmatism.
+
+@gozala encouraged pragmatism:
+> While I think that’s a very noble goal, I think it would be hard to sell for a very pragmatic crowd like browser vendors. I frequently see standardization process taking specs into least ambitious and most pragmatic direction, I often disagree, but I think often times that’s only way to make progress. Maybe some version of this goal could be articulated in [less] perfectionistic manner and in a more pragmatic one ?
+
+@jbenet agreed to that pragmatism:
+> **These goals are secondary in time to getting browser adoption. Meaning that we CAN do things like recommend ipfs:// ipns:// ipld://** IF browser vendors think that it's unlikely to get adoption this way now. We can work on unifying the fs-db-web rift later. **We're not dogmatic, we're pragmatic.** But we want to make sure we push in the right places and try to make as much as we can better.
+
```
jbenet: Also, worth noting:
- browsers currently assume `<origin>` is a _case insensitive_ thing. 
- so a construction like `dweb://ipfs/<base58-hash>` or even `ipfs://<base58-hash>` doesn't work without redirection or translation in the implementation.
- today, given firefox constraints, what is _actually_ fed into firefox as the proper URL is `ipfs://<base16-or-32-hash>` (this does not have to be the thing that shows up to the user or that we encourage people to use as links).

timthelion:
I felt kind of ignored, so I added a section on my concerns with `/ipfs` in a new PR here: https://github.com/ipfs/specs/pull/153/files

I know it is long, but I've put time into making real examples with real testable code and concrete examples.

addresses-scheme/readme.md
```diff
@@ -0,0 +1,72 @@
...
+also 
+
+> we'd like to improve things, particularly since "a whole new world" of things is joining the internet (blockchains, ipfs, other decentralized web things). It would be nice if there's a nice compatible way to bridge with the web's expectations (dweb://...) but work towards fixing things more broadly.
+
+also 
+
+> A minor reason is not having to force people to swallow N shemes (ipfs:// ipns:// ipld:// and counting), and instead use one that muxes.
```
lidel: I feel it should be explicitly stated in the document if this single scheme aims to be for IPFS ecosystem only, or something that can be adopted by other distributed systems. 

If we want to share it with others, eg. Ethereum, Tor Onion Services etc, then I see potential problem with single scheme: who is responsible for muxing when multiple vendors are in play?
Let's say we have: `dweb://ipfs/Qmbar..` and `dweb://foo/buz`.  If IPFS browser add-on provides handler for `dweb://`, FOO is unable to handle its own URIs.

timthelion:
Which I guess, means that `dweb` support has to be in its own package, 
so that tor wouldn't depend on ipfs...

addresses-scheme/readme.md
timthelion: Which I guess, means that `dweb` support has to be in its own package, so that tor wouldn't depend on ipfs...

flyingzumwalt:
I've added info from everyone's comments, except @timthelion -- provided suggestions on his PR. I also rearranged the document so that (hopefully) it will be easy for people to submit further corrections/additions, etc. Please keep your additions concise with lots of links out to the ongoing discussions in GH issues and other PRs.

Note: I'm going to be very busy traveling and presenting at a conference this week, so I will not have much bandwidth to guide this discussion or to keep it moving forward.

timthelion:
Just to let you konw, I will not be commenting further on this issue.

flyingzumwalt:
Why @timthelion? You're making good points. I'm trying to gather the info and discussions so that your contributions move this discussion forward, rather than being yet another github issue that lies unresolved. Have I done something to discourage you?

timthelion:
@flyingzumwalt don't worry, you didn't do anything wrong or anything to discourage me. I put several hours into explaining my position, I explained it as well as I could and I don't beleive that I could do so better than I have already.

But the real reason why I do not want to continue, is that this discussion is making me feel unhappy, frustrated, and combative. For the sake of my own mental health, I would prefer not to continue with this discussion.

Its not anything any of you said. I just feel like this sometimes.

flyingzumwalt:
I can certainly empathize with that @timthelion. 😓 If it's any consolation, the delay on getting a response to your final posix/unix questions was because the people who can give a good answer are busy with other stuff right now, but I'm going to make sure we get good answers on the record eventually.  I wanted to make sure they answer your _new_ questions rather than re-answering stuff that's already been discussed elsewhere or re-hashing debates that already happened. In order to do that, I needed to gather everything into a document that provides the context/history.

So please don't delete your fork with those comments and examples. If you're willing, it will help me cite your comments and avoid losing them over time if you put your comments into a separate file next to the readme.md -- ie. a file called "timthelion-comments.md" and submit a PR with that file in it. Alternatively, it's perfect material for a blog post. If you put it on your blog we can cite it there and save a copy on ipfs.

flyingzumwalt:
And I completely agree about the coldness of text-based communication. It can be really rough on the human heart.

addresses-scheme/readme.md
```diff
@@ -0,0 +1,134 @@
...
+#### Strength: Fits with Established Conventions & Expectations around URLs
+
+#### Strength: Fits with the way browsers identify origins
+
+When enforcing the Single Origin Policy, some current browsers expect: `origin = a:/b`, even though that's not what the URI spec calls for. The `ipfs:` approach fits with that expectation, with URIs like `ipfs:/<root-hash>/foo/bar`, the origin would be `ipfs:/<root-hash>`
+
+### Criticisms of this Approach
+
+#### Criticism 1: We want IPFS, IPNS and IPLD to be handled by a single schema
+Creating an `ipfs:` schema would not be enough because `ipfs:` only refers to mutable content. You would, at the very least, need an `ipns:` schema too.
```
jefft0: `ipfs:` only refers to mutable content

This is a typo, right? It should be: 

    `ipfs:` only refers to immutable content

jefft0:
@jbenet mentions the case insensitivity problem. I think the "Strengths" or "Criticisms" section of each approach should say whether the multihash is in the case insensitive part of the URI. (If a scheme requires a browser to hack the URI to not conflate `ipfs://QmAAA` with `ipfs://Qmaaa`, then it's a pretty strong criticism of the approach.) Do you want me to add comments to this effect in the document?

timthelion:
@jefft0 "URLs should treat upper case letters as equivalent to lower case **in scheme names**" https://www.ietf.org/rfc/rfc1738.txt

So long as http://gateway.ipfs.io/ipfs/Qmaqx89FcFCQ3EocZZEbzGf7mJZpKBe8XAZt8ijoLXkojN works, then why would `ipfs:/` be any different?

addresses-scheme/readme.md
```diff
@@ -0,0 +1,134 @@
...
+
+The short-term fix that people reach for is to create an `ipfs:` schema, as proposed in https://github.com/ipfs/specs/pull/139. That approach seems simple at first, but it's got problems.
+
+### Strengths of this Approach
+
+#### Strength: Fits with Established Conventions & Expectations around URLs
+
+#### Strength: Fits with the way browsers identify origins
+
+When enforcing the Single Origin Policy, some current browsers expect: `origin = a:/b`, even though that's not what the URI spec calls for. The `ipfs:` approach fits with that expectation, with URIs like `ipfs:/<root-hash>/foo/bar`, the origin would be `ipfs:/<root-hash>`
```
jefft0: I'm confused. The [spec in the pull request](https://github.com/pawal/specs/blob/f697953df6e36d0dffbb7849cb1d90f007c1354d/uri/ipfs.md) looks like `ipfs://ipfs/<root-hash>/foo/bar`, not `ipfs:/<root-hash>/foo/bar` as you wrote here. So the origin would be `ipfs://ipfs`. Do you want to refer to a different draft of the `ipfs:` scheme?

jefft0:
@timthelion: Yes, that's what the RFC says about the scheme part of `scheme://host`. The point is that browsers also treat the host part as case insensitive so that `http://github.com` and `http://GitHub.COM` are the same. If the multihash is in the "host" part of the URI, there could be problems.

addresses-scheme/readme.md
flyingzumwalt: Interesting point. 
A couple of things I will have to update the doc to reflect:

1. @pawal's spec at https://github.com/ipfs/specs/pull/139 includes a brief reference to the `fs:` scheme which is fleshed out in @lgierth's https://github.com/ipfs/specs/tree/doc/fs-paths/fs-paths 
2. A couple of points are implicit in @jbenet's arguments, which hang off the fact that an `ipfs:` scheme should, due to its name, only return ipfs content: 
    * if you create an `ipfs:` scheme, the `/ipfs/` in the path is redundant. 
    * if you create an `ipfs:` scheme you will end up needing more schemes: `ipns:`, `ipld:`, etc.

addresses-scheme/readme.md
flyingzumwalt: yes that's a typo. you're right. I need to fix that.

lgierth:
Thanks so much @flyingzumwalt, excellent write up :heart: and thanks for covering for me. I failed at producing any useful writeup during the in-web-browsers sprint. I'll try my best to reframe this from a closer-to-implementation perspective that'll help us quickly wrap up the discussion.

I won't repeat the arguments for and against the various ideas proposed here and in older discussions, this PR already does a good job at that. Also @jbenet is right:

> This is the current address scheme, and is not an "unresolved thing waiting for a decision". There are suggestions to _undo that decision_ and change to something else, for the sake of simplifying implementation of browser bridges

We've had lengthy and fruitful discussions spanning years, and consensus about the long term goal of path addressing has been reached long ago (see https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-157478515). What is up for **technical debate** is what we usually call the upgrade path -- how to integrate with existing protocols and infrastructure in a way that:

1. Makes it easy to simultaneously use the thing in its native environment, as well as the classic environments.
1. Opens a long term perspective for transition.
2. Yields an immediate benefit even within existing infrastructure.

Big protocol migrations happen over decades, and it's fundamentally important that new protocols interoperate well, otherwise the migration cost is simply too high. For example the IPv4/IPv6 migration came up with half a dozen IPv4-over-IPv6 and vice versa sub-protocols.

Note that for the upgrade path, we're not constrained to exactly what currently is. Many of the web's protocols (HTML, URLs, etc.) are so-called living standards, which aim to evolve relatively organically. We can propose new APIs and features, weigh in on discussions, and generally participate in the specification and development effort, as we did for example with the Suborigin working group.

Browser developers appear to not be generally opposed to the technicalities of dweb: URIs (and sometimes even inviting), if and only if we make an effort to make them play nice with the existing browser platform.

While browsers play a big role for ipfs (see ipfs/in-web-browsers) and are mentioned here a lot, we also want dweb: URIs to be supported in all kinds of tools/applications/libraries. One example is @Kubuxu's weechat plugin, which makes the URIs clickable.

## The four stages of the upgrade path for path addressing.

Note that whenever we talk about these, it's important to be very exact in the notation of the address. The number of slashes matters a great deal.

1. Current: HTTP-to-IPFS gateway
  - e.g. https://ipfs.io/ipfs/$hash
  - Has been working well for 2 years
  - Supported by many standard tools and applications out there because it's just HTTP
  - Issues:
    - Centralized in multiple ways: HTTP, Certificate Authorities, DNS
2. Short term: URL
  - e.g. ipfs://$hash
  - Well understood, adhering to WHATWG URL standard
  - Straightforward to implement
  - $hash is host/origin
  - Issues:
    - Hash needs to be base32 encoded, because URL hosts are case-insensitive.
      - Solution 1/2: Needs a redirect to the base32 hash when pasting any non-base32 hashes.
      - Solution 2/2: Probably needs a special base32 CID that retains information about the original non-base32 CID, so that we can avoid confusing UX around changing hashes.
    - Doesn't retain the path, but instead needs conversion step to/from URI.
      - Solution: The URI path can be derived from URL scheme and host and path.
3. Mid term: URI
  - e.g. dweb:/ipfs/$hash
  - No host/origin
  - Origin for Content Security Policy in browsers can be derived from path
  - Experimental implementation in @lidel's firefox addon
  - WebExtensions
    - Custom scheme/protocol handlers are not yet defined for WebExtensions
    - We need to **weigh in** here: https://bugzilla.mozilla.org/show_bug.cgi?id=1271553#c47
  - Note: this used to be called fs:
4. Long term: NURI
  - e.g. /ipfs/$hash
  - a.k.a. Nestable URI a.k.a. path addressing
  - Not just a noble goal, we do want to get there and lay the groundwork for it today.
  - You bring up good cases where path addressing / NURIs don't play well with current tooling and operating systems, @timthelion.
  - The FUSE integration of go-ipfs is only a very simple attempt at making NURIs work today.
  - Apart from FUSE, I can't think of any other significant effort we've made so far on this stage of the upgrade path.

These are not fixed serial steps, instead there's more of a wibbly-wobbly concurrence/simultaneity of all four stages. Networks are a wild mess of parts developing into differing directions, at varying speed (just like a [rhizome](https://en.wikipedia.org/wiki/Rhizome_(philosophy))). We definitely want to make the first three work in the foreseeable future, and also the fourth, although that one admittedly looks like a long shot at the moment :)

Regarding NURIs, I'd like to emphasize a comment @btrask [wrote](https://github.com/ipfs/go-ipfs/issues/1678#issuecomment-157500542),
which I think is helpful to understand the ubiquity of NURIs and how they help fix the FS/databases/web rift.

> > - canonical NURI is the path.
> >   - where NURI means Nestable URI. more universal than URIs. if you do not understand read the issue.
> >   - `/ipfs/Qm.../foo/bar`, `/ipns/Qm.../foo/bar`, and `/dns/foo.com/bar`.
> >   - use this everywhere you can!
>
> An IPFS path works as a relative URL. That means it is a resource name without a canonical location. Browsers can interpret it without conversion as relative to the current server, which presumably supports the IPFS interface.

As a reference, the filesystem ideas of [Plan 9](https://en.wikipedia.org/wiki/Plan_9_from_Bell_Labs) have been the inspiration for NURIs.

jefft0:
Is there a CID spec that shows how to unambiguously distinguish a base32 CID from a base58 CID? (Can't find the CID spec anywhere.)

jefft0:
> distinguish a base32 CID from a base58 CID

Never mind. The multibase code handles it. With Google I found https://github.com/ipld/cid . (It wasn't linked to from https://github.com/ipfs/specs or https://github.com/ipld/specs/blob/master/README.md or https://github.com/ipld/specs/blob/master/ipld/README.md . Maybe it should be?)

lidel:
I made a related comment in https://github.com/ipfs/in-web-browsers/issues/6#issuecomment-284666731
There should be tooling out-of-the-box for this, otherwise people will reinvent the wheel over and over.  🙃

mib-kd743naq:
> Is there a CID spec that shows how to unambiguously distinguish a base32 CID from a base58 CID?

@jefft0 a couple months ago I walked someone through that in detail: https://botbot.me/freenode/ipfs/2017-01-04/?msg=78819903&page=9

I never found the time afterwards to put this in a documente however :/

lgierth:
There's a remaining issue mentioned above with the ipfs:// URL scheme:

- Hash needs to be base32 encoded, because URL hosts are case-insensitive.
  - Solution 1/2: Needs a redirect to the base32 hash when pasting any non-base32 hashes.
  - Solution 2/2: Probably needs a special base32 CID that retains information about the original non-base32 CID, so that we can avoid confusing UX around changing hashes.

diasdavid:
I believe now is the time (more than ever) for us to coalesce our thoughts and ideas into this single document that can be used to present the case. There are browsers waiting for a clear path to decide if they should or not implement this.

I found myself looking for information across the issues on https://github.com/ipfs/specs/issues/131, https://github.com/ipfs/in-web-browsers/issues/29, https://github.com/ipfs/in-web-browsers/issues/8 and https://github.com/ipfs/in-web-browsers/issues/6 and got lost in the midst of all discussions.

We need a PRI to lead this to a ready to merge state. @flyingzumwalt @lgierth you have worked on this the most, do you want to lead the end of the effort? Can I enable you to get this done sooner?

dag:
Reposting from [ipfs-companion #164](https://github.com/ipfs/ipfs-companion/issues/164#issuecomment-341584995):

> 
> The `ipfs` scheme has no "authority" as per the RFC so it should not include the double forward slashes. It is debatable whether the `ipns` scheme includes an authority; it includes a domain name, but it's all routed through IPFS and for example there is no port number and the domain isn't necessarily a peer. Thus even with IPNS the domain is merely an identifier and not an "authority". I therefore propose that it would be more in compliance with the specification to have:
> 
> * `ipfs:{hash}` and
> * `ipns:{domain}`
> 
> Double slashes are a wart of the URI spec, so it's nice to be able to avoid them. The spec defines authority thus:
> 
> ```
> authority   = [ userinfo "@" ] host [ ":" port ]
> ```
> 
> IPFS is an unauthenticated peer-to-peer network, so there is no userinfo and no singular host or port.
> 
> ### Why I could be wrong
> 
> I don't think I'm wrong for the `ipfs` scheme, but I'm less certain for the `ipns` scheme. It could be argued that when the RFC defines `host` to include "registered name", it is very vague as to what that means and it could apply to IPNS. But I think most people would expect that if there is two slashes followed by a domain name, the computer will connect to that domain name, and it could be followed by a port number. Neither of those are necessarily or at all true with IPNS. Double slashes signify the existence of a centralized authority; the antithesis of IPFS.
> 
> https://tools.ietf.org/html/rfc3986#section-3.2
> 
> ### URN?
> 
> It could also be argued that the proper scheme to use is `urn`. That's what's used for things like ISBN and UUID, which are a lot like IPFS. `urn:isbn:0451450523`, `urn:ipfs:QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`... Just as ISBN tells you what to look up rather than where, so does an IPFS hash or IPNS name tell you what to look up, not which server to connect to. Personally I always thought URNs were kind of redundant though, why not just `isbn:{num}`?

I'll also note that both of these variants solve the issue with URL hosts being case insensitive. I personally think the proposed NURI variant or the `dweb` scheme are the wrong solutions because they don't follow any existing specifications, and I see no need to invent something new. I hope I'm not stepping on any toes. I feel that having `/ipfs` follow `dweb:` is sort of arbitrary; it would be more in line with existing standards to make it `dweb:ipfs:{hash}` but then you've just reinvented URNs.

dag:
OK I see now the idea with NURI is you do things like,

```
ipfs swarm connect /ip4/104.236.176.52/tcp/4001/ipfs/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z
```

which presumably you can't do without "nestable" URIs. But you can do at least the above with URIs if you don't use the authority part for the hash:

* P2P-network IPFS: `ipfs:QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z`
* Peer-specific IPFS: `ipfs://104.236.176.52:4001/QmSoLnSGccFuZQJzRadHn95W2CrSFmZuTdDWP8HXaHca9z`

The fact that it's a TCP connection over IPv4 is obvious from the scheme and host.

Now it's also clear that IPNS too should not put the domain in the authority part, or you can't express this:

* `ipns://104.236.176.52:4001/domain.com`

where that domain might not resolve to that IP at all.

lgierth:
> I hope I'm not stepping on any toes

Absolutely not! Super useful feedback, thanks :)

I'm super busy right now but I'll find time next week to address your suggestions.

In very brief words: URLs very strongly couple content and location, and that's something we're trying to break up to everyone's benefit. The dweb: URI scheme is meant for content addressing, and [multiaddr](https://github.com/multiformats/multiaddr) is meant for network addressing.

That's why you'd never address /ipns/somehash together with a network address. To the content layer the location is completely transparent.

---

The documentation regarding addressing is very much in flux but will stabilize until end-of-year. Multiaddrs should have never had a /ipfs protocol in them because it leads to confusion with the /ipfs namespace in content addresses. Multiaddrs will move to using /p2p instead soon to avoid this confusion in the future.

lgierth:
> To the content layer the location is completely transparent.

This is also why it's hashes are the right thing to use for an authority in a URL in our context. The hash *is the authority*. There is no server anywhere deciding what to serve you for a given request path. This is location-addressing (traditional web) vs. content-addressing (decentralized web).

dag:
I get what you're going for saying the hash is the authority, but I don't think that's how the RFC is meant to be read. If it was, we'd have `isbn://{num}` rather than `urn:isbn:{num}`. In the case of ISBN, you could say the number is the authority. But that's not what's implied by authority in this case. I strongly get the sense they really do mean "central authority on the network" in the sense of "server", which is also how every other URL scheme interprets it. And as I noted, if you *don't* use authority for hash you get benefits like a cleaner URI and no need to go base-32, and you communicate the distributed authorityless nature better: it's obvious from `ipfs:{hash}` that there is no central authority, and that the hash is a unique content identifier. It's obvious from `ipns:{domain}` that you're not necessarily connecting to the domain, and it's instead a unique content pointer. Plus as I also noted, when you actually do want to express a peer-specific connection, this way you can make that distinction.

This is also why we have `mailto:{user@host}` rather than `mailto://{user}@{host}` because the latter would imply you're connecting to `host` and logging in as `user` when in fact you're connecting to your own mail server, logging in as yourself and asking the server to send a mail. It's why we have `tel:{num}` rather than `tel://{num}` even though you're arguably "connecting" to the telephone number when you call it, because it's not a central network authority, the number is not an IP number or a name that resolves to one, and having a port number would be nonsensical. It's why we have `bitcoin:{addr}` rather than `bitcoin://{addr}`, again the address is an identifier not a central authority. And it is why we *do* have two forward slashes in `http`, `ftp` and `irc` etc, because those all *do* have a central network authority you talk to to get at the content.

> In very brief words: URLs very strongly couple content and location, and that's something we're trying to break up to everyone's benefit. The dweb: URI scheme is meant for content addressing, and multiaddr is meant for network addressing.

Certainly true for URLs, being "locators", but not for URIs, being "identifiers". Why not simply an `ipfs:` URI for content addressing, and an `ipfs://` URL for network addressing? It's consistent, it's spec-compliant, it doesn't invent a new wheel, and it makes it obvious to everyone what you're trying to communicate.

And if you want, you could still do dweb and multiaddr as well, as experiments. But both this spec, and ipfs-companion that I originally commented on, additionally suggest an `ipfs` URI scheme, and my main interest here is to comment on that, irrespective of dweb and multiaddr. Certainly multiaddr seems to have some interesting ideas and I don't mean to dismiss those out of hand.

RangerMauve:
I really like the ideas that @dag  is bringing up with using `ipfs:{hash}`. One worry that I have is that libraries might not be parsing that use case properly. A lot of environments (like node and the browser) come with URL support out of the box, and people might be tempted to make use of it. But if you use `url.parse(`ipfs:FOOBAR`)` it returns an object with an empty path and a lowercase hostname of `foobar`.

I think if the project does go in this direction, they'll need to be careful to inform people that it's not a URL and that they shouldn't be using URL libraries to parse it.

lgierth:
@dag 

> but I don't think that's how the RFC is meant to be read.

The RFC is not the relevant standards document regarding URLs nowadays. Instead browser vendors are implementing the WHATWG's spec: https://url.spec.whatwg.org/ - as we've been explicitly told by multiple browser developers.

In particular see 3.1 Host representations:

> A host is a domain, an IPv4 address, an IPv6 address, an opaque host, or an empty host. Typically a host serves as a network address, but it is sometimes used as opaque identifier in URLs where a network address is not necessary.

Hashes in `ipfs://` and `ipns://` are totally valid "opaque hosts".

---

> Why not simply an ipfs: URI for content addressing, and an ipfs:// URL for network addressing

IPFS doesn't concern itself with network addressing, and we want to keep the /ipfs and /ipns namespaces purely about content/data.

There's also one important thing that you miss and which is generally underexplained so far: the benefits of path addressing and a universal path namespace. Plan9 has demonstrated this beautifully, but there are also real world pains that it solves:

1. Scheme namespace pollution
2. The rift between addressing on the web, on the filesystem, and on databases
3. Balkanized data
4. Tools that *always* only support working with a tiny subset of data, just because they can't possibly implement all the URI and URL schemes out there.

The ideal longterm result is widespread use of paths a la `/ipfs/QmFoo/some/file.txt` or `/ipns/example.com` instead of URLs. These two content-addressed namespaces (immutable data and mutable names) nicely fit nicely into the overall universal namespace at `/`.

The benefits of a path namespace might not be immediately obvious just from IPFS itself, but note that IPLD is capable of addressing *ALL* content-addressed data out there, and you can address Git objects, BTC/Ethereum/Zcash transactions and blocks, as well as Torrent files, right now, today.

Other namespaces than /ipfs and /ipns can have different semantics, or add semantics on top of `/ipfs`. For example there's various other naming systems out there. And you can also express existing protocols in paths: `/http/exampe.com/some/file.txt`

Granted, widespread use of paths is a long stretch, and until we get there we'll try very hard to represent addresses as URLs and URIs:
- URIs do have the advantage that they can actually represent our paths without altering them, in exchange we still need to come up with a comprehensive security model though. (Examples: `dweb:/ipfs/QmFoo/some/file.txt`, `dweb:/ipns/example.com`)
- URLs need to alter the paths to represent, but in exchange we get an existing security model, and seamless integration with web browsers. (Examples: `ipfs://QmFoo`, `ipns://example.com`)

When reading these steps, remember that they aren't serial steps, instead they're a wibbly-wobbly concurrent. That means we aim to have all 4 steps of the upgrade path (HTTP-to-IPFS gateway, ipfs:// and ipns:// URLs, dweb: URI, plain universal paths) working simultaneously. This will maximize the exposure to all kinds of existing systems, and significantly lowers the adoption costs since migrations can happen in small steps along the way.

---

I'd like to close by saying that location-based addressing is a serious problem to how the web develops, probably the single biggest problem we're dealing with right now. With the general lack of encryption and signatures across all layers of the stack being a close second.

I also disagree that location-based addressing is the way that users expect addresses to work like. Users want names, you can observe it every time someone types a domain name into google search.

dag:
> url.parse

Yeah, it's not a URL. Is there perchance a `uri.parse` that doesn't expect a locator authority? How do libraries and browsers deal with `bitcoin:{addr}` links? Also it seems easier to fix broken URI libraries than to propose that everyone write new libraries for multiaddr...

> IPFS doesn't concern itself with network addressing

I based that off of the example from the documentation using `ipfs swarm connect`. My point was that even if we have an authorityless URI, we can still express the argument to `swarm connect` with a URL that does include the peer as an authority. I don't know how useful that is, but it's apparently something that IPFS already does, and multiaddr expresses. We can do it with just the URI standard.

> path addressing and a universal path namespace

I'm mainly concerned with the fact that we're still proposing to use URLs in the interim, so even if the long-term goal is to replace URLs, shouldn't we at least use URLs correctly while we still do use them?

Also, it's hard to beat URLs for crosslinking. If I type `abc:[//]xyz` everyone knows it's a reference to something. Humans know it, machines know it. If we replace that with `/abc/xyz` it's going to be much less obvious what it is, harder to automatically turn into links reliably, and it will clash with many text markup languages that take slashes to mean italics, etc. At least the `dweb` scheme keeps its status as a reference obvious, while still supporting this new multiaddr format.

...but I'm also a bit concerned with the format of multiaddr. It seems to arbitrarily privilege HTTP.

Take `/dns/foo.com/http/foo/bar/baz`:

Shouldn't this mean the `bar` protocol encapsulated inside the HTTP protocol? Why does parsing suddenly stop after `http`? And why is only HTTP paths part of the hierarchy of the path? Why don't we have `/dns/com/foo` to resolve `foo.com`? `/ip4/127/0/0/1` to resolve an IP?

I don't mean to criticize someone's baby, but it seems important to get links right for the future of distributed applications.

> I'd like to close by saying that location-based addressing is a serious problem to how the web develops, probably the single biggest problem we're dealing with right now. [...]

> I also disagree that location-based addressing is the way that users expect addresses to work like. Users want names, you can observe it every time someone types a domain name into google search.

Now I'm confused, because it seems to me that *I'm* the one advocating against location-based addressing, and advocating for name and reference based resolution. I'm suggesting, stop trying to treat content hashes as locations and start treating them as references. This is already well-supported in existing standards. And IPNS names are just pointers to hashes, so they should get the same treatment. An IPNS name or IPFS hash is not a location you connect to, it's a reference you look up in the network you're already connected to.

RangerMauve:
> url.parse

Node.js just has the `url` module which behaves as I mentioned, but the URL constructor in browsers treats `ipfs:ABCD` as 

```
origin: "ipfs://"
pathname: "ABCD"
protocol: "ipfs:"
```

kevina:
One problem which I think needs to be considered but it doesn't seam to be mentioned (and apologizes if it is I only skimmed the document and searched for the word absolute) is that most web sites use absolute paths in the urls (that is `href="/html/page1.html"`).  If we take a website and store it on ipfs, the website will not function if it is addressed by `http://ipfs.io/HASH/index.html` but will if it was `http://HASH.ipfs.dweb.link/` or something similar.

Was this discussed at all?

dag:
Another point to consider w.r.t. absolute paths is that it's how browsers will interpret multiaddr links. How do you make browsers support `href="/ipfs/{hash}"` without breaking support for absolute paths?

lamarpavel:
@dag If the browser interprets `href="/ipfs/{hash}"` as an absolute path to the local file system, then [mounting](https://ipfs.io/docs/commands/#ipfs-mount) ipfs to `/ipfs` should solve the problem without a need to add any support to browsers.

dag:
No, it would interpret it as an absolute path on the current host. [Try this link for example.](/ipns/ipfs.io) That was a link to `/ipns/ipfs.io`, but your browser will try to connect you with `https://github.com/ipns/ipfs.io`.

RangerMauve:
Yeah, the spec for [href](https://www.w3.org/TR/html4/struct/links.html#adef-href) specifically expects it to be a URI, so if you want to use this new format you'll likely need a new attribute added to the HTML spec.

lgierth:
It'll work just fine with `href="ipfs://<hash>"` and `href="dweb:/ipfs/<hash>"` -- I don't feel like this discussion is going anywhere at the moment, these are things that have been agreed on years ago.

I'll ping this thread once the DWeb Addressing spec is finally there (give me a week or two more). An outline and introduction have been merged in #167. We're looking for open questions and spots that need clarification.

lidel:
Note to self: use this PR as a place to incorporate existing resources listed at https://github.com/ipfs/in-web-browsers/issues/147 (including [ADDRESSING.md](https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md) / https://docs.ipfs.io/guides/guides/addressing/)

hsanjuan:
@lidel alive check on this...

lidel:
Ack, specs consolidation was on the backburner due to other priorities but I hope to tackle it in 2020Q2 at some point (after subdomain gateways land everywhere).

I think we should close this outdated PR just to decrease noise and point people at https://github.com/ipfs/in-web-browsers/issues/147 which has a proper timeline of all decisions, making it a much better entry point than this PR. 
Possibly should move mentioned issue to ipfs/specs too. 

@hsanjuan can you close this and move mentioned issue to ipfs/specs?

PR #157, opened 2017-04-29
Update links to multiformats

musoke:
Some repos have moved from github.com/jbenet to github.com/multiformats


lgierth:
Thanks :+1:

PR #167, opened 2017-11-11
Add DWeb Addressing intro and table of contents

lgierth:
This is a table of contents and intro for the addressing spec. (Some of this can surely be pulled out into a blog post or so later on.)

Please take a look and give some early feedback, I have a bunch more which isn't as coherent yet. A simple LGTM is sufficient too :)

cc @diasdavid @lidel @kyledrake @stebalien @whyrusleeping @Kubuxu @flyingzumwalt

### DIFF
--- a/dweb-addressing/README.md
+++ b/dweb-addressing/README.md
@@ -1,0 +1,39 @@
+# Addressing on the Decentralized Web
+
+> Note: This document is work in progress and largely incomplete.
+
+Authors:
+- Lars Gierth \<lgierth@ipfs.io>
+
+Table of contents:
+- Introduction
+- The precarious web
+  - Link competition and link rot
+  - The addressing rift
+- DWeb Addressing
+- Namespaces
+  - /ipfs -- immutable data
+  - /ipns -- mutable pointers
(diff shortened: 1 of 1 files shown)


addressing/README.md
```diff
@@ -0,0 +1,43 @@
...
+  - The addressing rift
+- DWeb Addressing
+- Namespaces
+  - /ipfs -- immutable data
+  - /ipns -- mutable pointers
+  - Addressing data from other content-addressed systems
+  - Network addressing
+- Interoperability
+  - DWeb Addressing with HTTP
+  - URL schemes
```
lidel: I would explicitly put names here ("URL schemes: `ipfs://` and `ipns://`"), just like we have `dweb:` below.
It is what people usually ask about, and it makes sense to have these names pop.

lidel:
@lgierth  are we able to create shared branch and a PR from it so that multiple people can work on this spec document? or should I just fork the repo and create a PR from my fork? 
(I tried to create `dweb-addressing` branch here, but I don't have permissions to this repo)

PR #168, opened 2017-11-15
bitswap/README.md: Cleaning up the format

dgrisham:
Going to start updating this document soon, so I figured I'd handle the reformatting before I made any content changes to keep commits/updates clean.

PR #171, opened 2017-12-13
Bitswap spec update

dgrisham:
Editing and updating the Bitswap spec. Work in progress. Starts from https://github.com/ipfs/specs/pull/168

### DIFF
--- a/bitswap/README.md
+++ b/bitswap/README.md
@@ -25,67 +25,54 @@ Bitswap is the data trading module for IPFS. Its purpose is to request blocks fr
 
 # Organization of this document
 
-TODO
+  - [Introduction](#introduction)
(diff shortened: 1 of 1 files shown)


dgrisham:
@diasdavid This is ready for a first-pass review :)

diasdavid:
Skimmed through it, so far so good :)

diasdavid:
@dgrisham when will the final version be ready?

dgrisham:
I'll work on getting the Bitswap sessions section together within the next week or so when I have time. That and the API are the most substantial parts left -- is the API far enough along to be set, or should that be left unfinished for now in lieu of future changes?

diasdavid:
sounds good @dgrisham. Let's take it as it is and continue from there.

PR #182, opened 2018-05-31
fix IPLD broken link

marcogbarcellos:
The link for the IPLD github is returning 404: https://github.com/ipld/specs/tree/master/ipld
I believe the right link is https://github.com/ipld/specs/tree/master

Hope it helps!


marcogbarcellos:
Another open PR already fixes the problem.

PR #188, opened 2018-08-22
remove unwant from the public API

Stebalien:
Update for https://github.com/ipfs/go-ipfs/pull/5308


alanshaw:
refs https://github.com/ipfs/interface-ipfs-core/issues/339

PR #199, opened 2019-04-24
Add spec for file bits of unixfs v1

whyrusleeping:
Just the bits I need right now for the filecoin spec (we use unixfs as a transfer mechanism). If anyone wants to take this over from me and flesh out the rest of it, be my guest.


Stebalien:
Good enough to merge.

Issue #217, opened 2019-08-08
Adding additional file metadata to UnixFSv1


mikeal:
Current UnixFSv1 importers do not encode most of the standard file metadata from most file systems.

This has been a particular challenge for package managers since they already rely on some of this metadata.

The goal of this issue is to surface all the necessary discussion points in order to drive a new PR against the [unixfs spec](https://github.com/ipfs/specs/tree/master/unixfs).

## Potential metadata

* [Permissions](https://github.com/ipfs/unixfs-v2/issues/14)
  * Executable bit
  * Ownership (user and group)
* Filename *in* file object
* mtime
* ctime
* atime

### Additional considerations

For time stamps (mtime, ctime, atime) we need to decide if we’re going to use high precision times or not. Most systems expect a 32-bit integer (low precision) while other use cases may need a 64-bit integer (high precision).

Do we want to store additional metadata of the directory? How do we handle updating this when someone updates only a single file in the directory?

## Where do we store this metadata?

In terms of the [data format](https://github.com/ipfs/specs/tree/master/unixfs#data-format), should these properties be added to the `File` message or the `Data` message?

## History

The history of this feature as well as meeting notes where this feature was prioritized are [available here](https://github.com/ipfs/unixfs-v2/issues/28).

lidel:
Could this also include support for opt-in setting of content type?

[The spec @ 12a3d57](https://github.com/ipfs/specs/blob/12a3d57028f37086e211cc460be684da5f5333cf/unixfs/README.md) already has a field for this:
```go
message Metadata {
	optional string MimeType = 1;
}
```


This would enable people to solve false-positives in content-type sniffing before v2 lands (https://github.com/ipfs/unixfs-v2/issues/11)

mikeal:
Can someone more familiar with the original spec and implementations explain how the Metadata message is currently used? It seems obvious that we should leverage it and also start using the MimeType field but without knowing a bit more about the history and current usage I can’t tell if we’re likely to break anything.

warpfork:
I'm not sure what to make of the MimeType idea.  Unixy filesystems don't have a concept of MimeType; that's much higher level.

It certainly seems like an sizable embiggening of scope from the bullet points at the top.

mib-kd743naq:
> Can someone more familiar with the original spec and implementations explain how the Metadata message is currently used?

[...]

PR #220, opened 2019-08-30
Add file metadata and permissions to UnixFSv1

mikeal:
Prior discussion in #217

### DIFF
--- a/unixfs/README.md
+++ b/unixfs/README.md
@@ -38,13 +38,11 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	
-	/* file metadata */
-	optional uint32 mtime = 7;
-	optional uint32 atime = 8;
-	optional uint32 ctime = 9;
-	
 	/* file permissions */
-	optional uint32 mode = 10;
+	optional uint32 mode = 7;
+	
+	/* file time data */
+	optional uint32 mtime = 8;
 }
 
 message Metadata {
--- a/unixfs/README.md
+++ b/unixfs/README.md
@@ -58,7 +58,7 @@ For files that are comprised of more than a single block, the 'Type' field will 
 
 This data is serialized and placed inside the 'Data' field of the outer merkledag protobuf, which also contains the actual links to the child nodes of this object.
 
-The `mode` is for optionally persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation).
+The `mode` is for optionally persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) \[[spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html)\].
 
 ## Importing
 
@@ -77,7 +77,6 @@ Chunking strategy currently has two different options, 'fixed size' and 'rabin'.
 Fixed size chunking will chunk the input data into pieces of a given size. Rabin
 chunking will chunk the input data using rabin fingerprinting to determine the
 boundaries between chunks.
-
 
 ### Layout
 
--- a/unixfs/README.md
+++ b/unixfs/README.md
@@ -37,6 +37,16 @@ message Data {
 
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
+	
+	/* file metadata */
+	optional uint32 mtime = 7;
+	optional uint32 atime = 8;
+	optional uint32 ctime = 9;
+	
+	/* file permissions */
+	optional uint32 permissions = 10;
+	optional uint32 uid = 11;
+	optional uint32 gid = 12;
 }
 
 message Metadata {
--- a/unixfs/README.md
+++ b/unixfs/README.md
@@ -58,6 +58,8 @@ For files that are comprised of more than a single block, the 'Type' field will 
 
 This data is serialized and placed inside the 'Data' field of the outer merkledag protobuf, which also contains the actual links to the child nodes of this object.
 
+The `mode` is for optionally persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation).
+
 ## Importing
 
 Importing a file into unixfs is split up into two parts. The first is chunking, the second is layout.
--- a/unixfs/README.md
+++ b/unixfs/README.md
@@ -44,9 +44,7 @@ message Data {
 	optional uint32 ctime = 9;
 	
 	/* file permissions */
-	optional uint32 permissions = 10;
-	optional uint32 uid = 11;
-	optional uint32 gid = 12;
+	optional uint32 mode = 10;
 }
 
 message Metadata {


mikeal:
Question: what happens when a message containing these new properties is parsed by someone using the old protobuf schema? Are the properties ignored or does it fail?

unixfs/README.md
```diff
@@ -37,6 +37,16 @@ message Data {
 
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
+	
+	/* file metadata */
+	optional uint32 mtime = 7;
+	optional uint32 atime = 8;
+	optional uint32 ctime = 9;
```
Stebalien: @djdv pointed out that windows supports creation time.

unixfs/README.md
```diff
@@ -37,6 +37,16 @@ message Data {
...
+	
+	/* file permissions */
+	optional uint32 permissions = 10;
+	optional uint32 uid = 11;
+	optional uint32 gid = 12;
```
Stebalien: For now, we can probably stick with modification times and permissions. We probably don't need uid/gid.

Eventually, we may want user _names_ and group _names_ but I don't believe that's a pressing issue.

@andrew?

Stebalien:
> Question: what happens when a message containing these new properties is parsed by someone using the old protobuf schema? Are the properties ignored or does it fail?

Ignored.

unixfs/README.md
mib-kd743naq: Not supporting fractional time is an oversight. Especially provided we might be stuck with this stop-gap for quite a while. I urge you to consider doing the following:
```
message Timespec {
       // Use of int64 is deliberate - negative epoch is super unlikely
       // Using sint64 would introduce zig-zag encoding ( harder to eyeball )
       // The varint representing the time of writing this is 5 bytes long
       // It will remain so until October 26, 3058 ( 34,359,738,367 )
       required int64 UnixEpoch = 1;

       // fixed32 ( always 4 bytes, no varint ), as nanosecs are often > 2^28
       // https://developers.google.com/protocol-buffers/docs/proto#scalar
       // NOTE: on the wire this value will be *little* endian
       optional fixed32 UnixEpochNsec = 2;
}
```

And using `Timespec` for `mtime`/`ctime`/etc ( `atime` is a really odd choice to carry but since it is optional it's not a big deal )

unixfs/README.md
```diff
@@ -37,6 +37,16 @@ message Data {
...
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
+	
+	/* file metadata */
+	optional uint32 mtime = 7;
+	optional uint32 atime = 8;
+	optional uint32 ctime = 9;
+	
+	/* file permissions */
+	optional uint32 permissions = 10;
```
mib-kd743naq: Instead of `permissions`, I recommend calling this`filemode` and have it be the literal spec/32-bit representation of golang's `FileMode`. It already [captures the interesting "data bits" really well](https://golang.org/pkg/os/#FileMode) for our use cases, and is also a stable format: https://github.com/golang/go/issues/25422#issuecomment-524078451

achingbrain: Things that aren't files have permissions too, so I don't think `filemode` is a great name.

`permissions` is the name everyone uses when talking about permissions so it seems fine to me.

If not that then how about `accessmode`?

mib-kd743naq: > permissions is the name everyone uses when talking about permissions

I am not only talking about permissions though :) I am talking about the entirety of `mode_t` encoded as a 32bit field. The nomenclature is a bit confusing, but this is POSIX for you: among other things a **`directory`** is one of the [available **file** types](https://www.gnu.org/software/libc/manual/html_node/Testing-File-Type.html#Testing-File-Type).

The reason it is extremely interesting to have the filetype info on each individual entry ( just like in common VFS `inode`s ) is so that after fetching a directory block you can **instantly** tell how any files, subdirectories, links and other weirdness it contains. With no further trips to the network.

achingbrain: How about just `mode`?

unixfs/README.md
warpfork: I'd like to offer some perspective from the field in terms of what I have and have not, ever, needed when working with package management as well as containerland:

- mtime comes up sometimes.

- atime is completely useless.  It will often get set again in the course of... anything else you're doing.  The `atime` vs `relatime` vs `noatime` flags on common filesystems have also muddied the field beyond all recognition.  There is approximately one program in the universe known to look at atime, and it's something you've never heard of.  atime should be disregarded.  atime provides purely negative value.

- ctime is completely useless.  It can't even be expressed unless you're writing filesystem drivers.  While it so happens that we are, and we can, that doesn't mean it's a good idea, that it's useful, or that we should.  For example, if we were to support commands that just unpack/manifest files on an existing (e.g. ext4) filesystem, we would not be able to preserve ctime anyway.  Most other tooling exists within this limitation as well, and as a result, the world has evolved to effectively disregard ctime for any significant purposes; e.g., if there was a piece of software that would not work correctly if ctimes changed, it would fail to work in any container system, anything that uses tarballs, etc; effectively, it would be nonviable, and resultingly, such software tends not to exist.

I'd think long and hard before perpetuating these pieces of metadata just because we 'can'.  They're *remarkably* useless, unused, and in many contexts unus*able*.

- fractional timestamps I also have very mixed feelings on.  We can observe that most tar systems don't do them, in practice.  Yes, I know there's an extension for that.  Go check if your tar command actually does that by default.  Go stdlib tar doesn't.  Docker and such systems have elided fractional components for ages as well, since they've been using the go stdlib tar libraries.  By and large, nobody seems to have noticed or cared.  We can look at the rate of serious problems being reported about timestamp precision in e.g. docker and use it as a barometer for how much fractional timestamps matter.  I'd say it's "not much".

unixfs/README.md
warpfork: I like `mode` as a name.  :+1: 

I'd refrain from referring directly to the golang filemode bits for docs.  Three reasons: They've combined several other things in that struct which are not sensible for us to combine (namely, file _types_, which we represent with different messages entirely, right?).  They've used a very unusual bit position for the 'sticky' bit (this has caused me painful bugs in the past).  And finally, it's probably easier to just write our own doc labeling each bit than it is to outsource.

We also need to document what is and is not in this field.  Either "permissions" or the "mode" name are very likely to carry the connotations of more than just the 0777/rwxrwxrwx bits for many readers.

The typical bits beyond the 0777 mask are: setuid, setguid, and sticky.  We should document clearly if we're doing those or not, and what position they're in. In other systems, these (again, 'typically') pack into the 07000 mask.  [I've done digging on this before](https://github.com/polydawn/rio/blob/4fef789ffc851aaf3ac291a91c21e477e6367811/fs/metadata.go#L29-L50) and if we're going to do these as a bitmask, I'd hew to the common form used there.  But we shouldn't skip the question of if we're doing them at all; and the question of if we're doing them as a bitmask.

My personal weakly-held opinion is to advocate for considering support for these additional bits, but not including them in a bitmask.  None of these three additional bits is frequently used in bulk (whole computers generally have a few dozen setuid/setgid things on the whole filesystem, and approximately *one* sticky bit (on `/tmp`)), so there's no reason to work about minimizing their encoded size; and they're also *extremely* important bits to Handle Carefully because of their relationship to system security, so setting them apart and handling them with oven mitts so to speak is a good idea.

warpfork:
I'd like to just mention a couple links to prior art that's not merely prior art, but also particularly easy to read and review for inspirations:

- the casync project persists a lot of things -- and this is particularly interesting because some of them deeply silly IMO (even as flags), but nonetheless, it's quite a list of things you *could* preserve: https://github.com/systemd/casync/blob/e4a3c5efc8f11e0e99f8cc97bd417665d92b40a9/src/caformat.h#L82-L125  It's such a list of things you *could* preserve that I think it's a fantastic highlight of why the *should*-vs-*could* discussion is always, always important :)
- the reproducible builds group has done some comparisons of archive systems and what they support which is very comprehensive in its considerations: https://reproducible-builds.org/events/berlin2017/ReproducibleSummitIIIEventDocumentation/ , heading "Mapping out archive formats" (though this is badly munged, sorry to say)

Both of these (as well as the specs of tar, which I'm assuming everyone's at least given a cursory glance at already) are highly worth a quick skim just to see what other people have covered when trying to map this terrain.

ivan386:
I think all file or dir metadata must be in separate metadata block. This will be more flexible than put it in block with links to sub blocks. Don't make mistakes as in bit torrent. Separate metadata block can be included in dirrectory block for each file or directory that  it contains. It can wrap first file block if it need. It can contain links to blocks in different format (raw).

mib-kd743naq:
> It will protect links between dir block and data that it contains.

@ivan386 could you elaborate what do you mean by this?

@achingbrain @warpfork note that my suggestions are just that: suggestions, based on the obscenely long time I've spent thinking about all this. The proposed PR is not usable by me in its entirety ( due to 1:1 mapping of entries/metadata-objects ), and that is fine: a properly of a robust framework is the ability to have multiple experiments run simultaneously. Please do not take my suggestions seriously enough to potentially derail this PR ;)

ivan386:
> It will protect links between dir block and data that it contains.
> 
> @ivan386 could you elaborate what do you mean by this?

Every change of metadata change block hash that contain it. If include file/dir metadata to it root block it will be many root hashes for same file/dir with diffrent metadata. Leaf nodes will be same but root block different. This root block can be lost becouse it wery rare. It break chain and file will be inacceseble from parent block. All file data will be awaliable in IPFS but we can't get it without root block.

Only way to return that root block in IPFS is to generate it from file data again. But if file root block will contain ctime or other metadata we can't gues what it was. And we will generate diffrent root hash without ability recover old root hash for that file.

Separate Metadata block wery small for each file/dir it can be included in parent dir block by using of [identity hash](https://github.com/ipfs/go-ipfs/issues/4697). This dir block will contain all files metadata and links to file/dir root block without metadata that always same for same data and not so rare. In this case we can lost links to root blocks to files/dirs only with parent dir block. And can recover root file block only by using file data.

unixfs/README.md
mikeal: I only want to store what is needed for the top level use cases (package managers) given this is not a permanent or even our ideal solution to these problems. My assumption is that they need something close to what `stat` gives them, which includes `atime` and `ctime`, but if nobody needs this we shouldn’t preserve this data.

My experience with fractional timestamps is that most FS API’s don’t give them to you by default and require an additional flag or option in order to see them, which didn’t meet the bar I had set, but again, if package managers need this then we can do it.

unixfs/README.md
mikeal: I cut them for now but I’d like @andrew to verify they aren’t needed before merging.

unixfs/README.md
mikeal: changed to `mode`.

mikeal:
> Every change of metadata change block hash that contain it. If include file/dir metadata to it root block it will be many root hashes for same file/dir with diffrent metadata. Leaf nodes will be same but root block different. This root block can be lost becouse it wery rare. It break chain and file will be inacceseble from parent block. All file data will be awaliable in IPFS but we can't get it without root block.

The file node is already, effectively, a root node of metadata pointing to linked file data. It’s not clear to me what another layer of indirection (metadata message) would preserve if it’s still what everyone needs to link to. The current use case for the metadata message is to define a `content-type` for HTTP which makes sense since filesystems do not store this information and people may want to serve the same file with different content types. This new metadata is all related to the representation of the file in a filesystem, which is what the File message already is, and adding a layer of indirection would degrade performance.

[We discussed this in the issue](https://github.com/ipfs/specs/issues/217) and the consensus seemed to be that this was the best path forward.

unixfs/README.md
mikeal: Also, note that high precision 64bit timestamps **are** in the [current schema](https://github.com/ipfs/unixfs-v2/issues/29) for `unixfsv2`.

ivan386:
@mikeal content-type is optional field. mtime can be used in HTTP to.

momack2:
Wanted to quickly check whether "mode" will fit the use case described for guix such that executable bit can be contained within file metadata? According to @ambrevar at IPFS camp, that's one of the few remaining blocker to get the [guix-ipfs-cache](https://github.com/fps/guix-ipfs-cache) over the finish line - https://github.com/ipfs/camp/blob/master/DEEP_DIVES/package-managers/README.md#guix-patch-notes-that-may-be-helpful-in-other-package-manager-efforts

Stebalien:
@momack2 yes. The mode allows encoding the executable bit along with all the other bits (sticky, read, write, etc.). Although, we should probably define how _we're_ using the mode as we don't want bits like "directory".

unixfs/README.md
mikeal: I’ve added a note about `mode` being in standard numeric notation and linked to the wikipedia article describing the behavior. https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation

Stebalien: The official POSIX definition is: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html.

I'd say that only the lower 9 bits (ugo, rwx) are defined for now.

mikeal: I’ll add this to the comment as well.

Ambrevar:
@momack2: Just to confirm what @Stebalien said, yes that would do it!  Note that the link to https://github.com/fps/guix-ipfs-cache is a different (more hacky) proof of concept.  The proper patch waiting to be merged can be reviewed at https://issues.guix.gnu.org/issue/33899.

unixfs/README.md
djdv: >@djdv pointed out that windows supports creation time.

(sorry for the late reply)
Just to be clear this attribute also exists on a range of systems I'm aware of, and probably more.
Windows(CreationTime), Solaris (crtime), FreeBSD (st_birthtim), Darwin (st_birthtimespec).

mikeal: what’s the consensus here? keep all 3 or cut down to ctime and mtime?

warpfork: IMO a prerequisite for adding these info like 'ctime' to unixfs should be to have documented answers to:

- If a command like `ipfs get` is going to unpack content from unixfs into a tree of dirs and files on a regular (say, ext4, just to pick something) filesystem, and this were run by a non root user, what should the behavior be?
- Is there any use to ctime other than being effectively a random number generator in the input data, which users cannot for any practical purposes control?  Should commands to add files to unixfs have a default behavior to set it to a "zero" value?  If so, does that change the expected behavior of either 'unpack' or mount commands?

... because whoever implements code around this *will* discover these questions about five minutes after we merge this.

(Or maybe not, and we quietly introduce a defacto random number generator to our nice content addressable system without noticing, which would be... worse.)

warpfork: I should note I have fairly strong and (I'd like to think) fairly informed opinions about this, because I've previously [wrestled with it in the Timeless project](https://github.com/polydawn/go-timeless-api/blob/0ece408663edb9dbc6109ef8690f8425f0d8f5f4/filesetFilters.go)... and while I'm not here to tell you I had The Solution over there, I *will* tell you that underestimating the complexity of the normalizing filters system that's almost certain to end  up growing around this area would be a serious mistake.  It's a touchy thing and has been a source of many headaches and bugs for me.  Treat it cautiously and especially be wary of adding data if there's no plan for what its normalizers would look and act like.

achingbrain: We could split this PR into two parts - merge the `mode` property (which seems relatively uncontentious and solves some real-world problems our users have today) and bottom out the `*time` related stuff separately?

mikeal: There’s two questions here that seem to be getting merged together:

1. Should *anyone ever* be allowed to add these time properties to the File.
2. How should IPFS *use* these properties, and should that behavior be optional or by default.

I’m inclined to follow @achingbrain’s advise and split it at this point. However, I’m going to keep `mtime` *in the schema* since it seems to be the least controversial and it gives us some room to experiment with `*time` issues in IPFS. Without it being in the schema we have no room to even experiment sufficiently.

We can work through the `*time` issues in the initial implementations and decide before we ship whether or not we actually want to use it in IPFS.

mib-kd743naq:
Want to raise one last question from my corner of the ring, I am afraid this is being lost...

In https://github.com/ipfs/specs/pull/220#issuecomment-527669755 @Stebalien said:
> Although, we should probably define how we're using the mode as we don't want bits like "directory".

Given that this detour is taken for the sake of the package management / nix-like-things groups: does it actually make sense to do this all as a new generic wrapper node that we will **have to support forever**?

To reiterate on (again @Stebalien)'s point in https://github.com/ipfs/specs/issues/217#issuecomment-525973631:
> Names are already a part of the directory. Making metadata a part of the directory isn't all that odd. I'd expect most tools to reference files relative to directories anyways.

As far as I can tell, for the original set of users waiting on this, there is no usecase where a piece of data needs POSIX attributes **without a filename**. The filename is already only available as a **directory** wrapper node, there is no other way to attach name to an "anonymous stream of data". What makes other attributes different enough to warrant a non-directory anonymous wrapper-node?

unixfs/README.md
djdv: >@djdv pointed out that windows supports creation time.

For some reason my previous comment is gone now. I had a list of operating systems with links to their properties...

This property is present in other systems as well, usually under the name `crtime`, or `birthtime`.

mikeal:
Update:

1. This change has been given the version number “1.5” ;)
2. After a good sync discussion it was decided that a better approach is to put this metadata in the directory rather than the file message. This PR can be closed and @Stebalien is working on a new proposal.

unixfs/README.md
```diff
@@ -37,6 +37,12 @@ message Data {
 
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
+	
+	/* file permissions */
+	optional uint32 mode = 7;
```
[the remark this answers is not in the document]
momack2: @achingbrain may know this too...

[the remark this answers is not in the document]
achingbrain: Executable is the main one, I think everything else is "nice to have".

achingbrain:
> After a good sync discussion it was decided that...

@mikeal was this documented anywhere?

PR #226, opened 2019-11-14
docs: adds spec sections for mode and mtime metadata

achingbrain:
Following our Redux Redux meeting, here's what we decided (as I recall).  It largely swings back towards #220 with added sections on deduping and inlining and is more specific about the sizes of things.

The spec is updated with metadata implementation details and a decision rationale section has been added at the end.

Follows on from/supersedes #223

### DIFF
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -1,4 +1,4 @@
-# ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) UnixFS
+# ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) UnixFS  <!-- omit in toc -->
 
 **Author(s)**:
 - NA
@@ -11,15 +11,31 @@ UnixFS is a [protocol-buffers](https://developers.google.com/protocol-buffers/) 
 
 Draft work and discussion on a specification for the upcoming version 2 of the UnixFS format is happening in the [`ipfs/unixfs-v2` repo](https://github.com/ipfs/unixfs-v2). Please see the issues there for discussion and PRs for drafts. When the specification is completed there, it will be copied back to this repo and replace this document.
 
-## Table of Contents
+## Table of Contents <!-- omit in toc -->
 
-TODO
+- [Implementations](#implementations)
+- [Data Format](#data-format)
+- [Metadata](#metadata)
+	- [Inheritance](#inheritance)
+	- [Deduplication and inlining](#deduplication-and-inlining)
+- [Importing](#importing)
+	- [Chunking](#chunking)
+	- [Layout](#layout)
+- [Exporting](#exporting)
+- [Design decision rationale](#design-decision-rationale)
+	- [Metadata](#metadata-1)
+		- [Separate Metadata node](#separate-metadata-node)
+		- [Metadata in the directory](#metadata-in-the-directory)
+		- [Metadata in the file](#metadata-in-the-file)
+		- [Side trees](#side-trees)
+		- [Side database](#side-database)
 
 ## Implementations
 
 - JavaScript
   - Data Formats - [unixfs](https://github.com/ipfs/js-ipfs-unixfs)
-  - Importers and Exporters - [unixfs-engine](https://github.com/ipfs/js-ipfs-unixfs-engine)
+  - Importer - [unixfs-importer](https://github.com/ipfs/js-ipfs-unixfs-importer)
+  - Exporter - [unixfs-exporter](https://github.com/ipfs/js-ipfs-unixfs-exporter)
 - Go
   - [`ipfs/go-ipfs/unixfs`](https://github.com/ipfs/go-ipfs/tree/b3faaad1310bcc32dc3dd24e1919e9edf51edba8/unixfs)
   - Protocol Buffer Definitions - [`ipfs/go-ipfs/unixfs/pb`](https://github.com/ipfs/go-ipfs/blob/b3faaad1310bcc32dc3dd24e1919e9edf51edba8/unixfs/pb/unixfs.proto)
@@ -66,11 +82,10 @@ The serialized size of a UnixFS node must not exceed 256KiB in order to work wil
 
 ## Metadata
 
-UnixFS currently supports three optional metadata fields:
+UnixFS currently supports two optional metadata fields:
 
 * `mode` -- The `mode` is for persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) \[[spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html)\]. If unspecified this defaults to `0755` for directories/HAMT shards and `0644` for all other types where applicable.
 * `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified.
-* `MimeType` -- The mime type of the file. This field is deprecated, as is the `Metadata` message and neither will be present in `UnixFSv2` because mime types are not typically properties of a file system.
 
 ### Inheritance
 
@@ -82,7 +97,7 @@ Where the file data is small it would normally be stored in the `Data` field of 
 
 To aid in deduplication of data even for small files, file data can be stored in a separate node linked to from the `File` node in order for the data to have a constant [CID] regardless of the metadata associated with it.
 
-As a further optimization, if the `File` node's serialized size is small, it may be inlined into it's v1 [CID] by using the [`identity`](https://github.com/multiformats/multicodec/blob/master/table.csv) [multihash].
+As a further optimization, if the `File` node's serialized size is small, it may be inlined into its v1 [CID] by using the [`identity`](https://github.com/multiformats/multicodec/blob/master/table.csv) [multihash].
 
 Such [CID]s must consist of 23 bytes or fewer in order for them to fit inside the 63 character limit for a DNS label when encoded in base32 (see [RFC1035 Section 2.3.1](https://tools.ietf.org/html/rfc1035#section-2.3.1)).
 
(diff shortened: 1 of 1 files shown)


UNIXFS.md
```diff
@@ -43,9 +43,10 @@ message Data {
 	optional bytes Data = 2;
 	optional uint64 filesize = 3;
 	repeated uint64 blocksizes = 4;
-
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
+	optional uint32 mode = 7;
+	optional int64 mtime = 8;
```
achingbrain: These are here and not in the `Metadata` message because it's deprecated so let's stay well away from it.

achingbrain:
Cc @ribasushi @ianopolous - I can't add you to the review box for some reason.

achingbrain:
Cc @ribasushi @ianopolous @rvagg - I can't add you to the review box for some reason.

UNIXFS.md
```diff
@@ -59,6 +60,32 @@ For files that are comprised of more than a single block, the 'Type' field will
...
+
+The serialized size of a UnixFS node must not exceed 256KiB in order to work will with the [Bitswap] protocol.
+
+## Metadata
+
+UnixFS currently supports three optional metadata fields:
+
+* `mode` -- The `mode` is for persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) \[[spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html)\]. If unspecified this defaults to `0755` for directories/HAMT shards and `0644` for all other types where applicable.
+* `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified.
+* `MimeType` -- The mime type of the file. This field is deprecated, as is the `Metadata` message and neither will be present in `UnixFSv2` because mime types are not typically properties of a file system.
```
Stebalien: We can probably just drop this comment now that we aren't even including it in the protobuf.

UNIXFS.md
```diff
@@ -59,6 +60,32 @@ For files that are comprised of more than a single block, the 'Type' field will
...
+
+### Deduplication and inlining
+
+Where the file data is small it would normally be stored in the `Data` field of the UnixFS `File` node.
+
+To aid in deduplication of data even for small files, file data can be stored in a separate node linked to from the `File` node in order for the data to have a constant [CID] regardless of the metadata associated with it.
+
+As a further optimization, if the `File` node's serialized size is small, it may be inlined into it's v1 [CID] by using the [`identity`](https://github.com/multiformats/multicodec/blob/master/table.csv) [multihash].
+
+Such [CID]s must consist of 23 bytes or fewer in order for them to fit inside the 63 character limit for a DNS label when encoded in base32 (see [RFC1035 Section 2.3.1](https://tools.ietf.org/html/rfc1035#section-2.3.1)).
```
Stebalien: * Could you do a quick check to see if this is actually possible?
* In terms of domains, we really need to generalize to multiple subdomains (different issue).

Stebalien:
> I didn't include the pros and cons of the other options because I don't think they should be included in the spec. That is, if I'm implementing this, I read the spec to understand what I should implement and how it should behave, I don't want to read a bunch of exposition on how we got to where we are.

These need to be included _somewhere_.

momack2:
See the spec Raul did for Eth2.0's use of libp2p: https://github.com/ethereum/eth2.0-specs/blob/dev/specs/networking/p2p-interface.md#design-decision-rationale -- does a really good job capturing design rationale in the spec itself, which I think is warranted here given how foundational these decisions are

achingbrain:
@Stebalien @momack2 Raul's is a nice format - I've added the decision rationale at the bottom in a similar style as suggested.

UNIXFS.md
achingbrain: I think so, as long as we don't get to the extremes of any number values.  For 'reasonable' values it should work:

```javascript
const CID = require('cids')
const multihashing = require('multihashing-async')
const UnixFS = require('ipfs-unixfs')

async function main () {
  const data = new UnixFS('file')
  data.addBlockSize(256)

  const multihash = await multihashing(data.marshal(), 'identity')
  const cid = new CID(1, 'dag-pb', multihash)

  console.info(cid.toBaseEncodedString('base32').length)
  // prints '21'
}

main()
```

For the following configurations:

```javascript
const data = new UnixFS('file')
data.addBlockSize(256)
data.fanout = 4294967295  // e.g. max uint32

// prints '30'
```

Stebalien: Nice!

UNIXFS.md
```diff
@@ -59,6 +76,31 @@ For files that are comprised of more than a single block, the 'Type' field will
...
 
+For files comprised of a single block, the 'Type' field will be set to 'File', 'filesize' will be set to the total number of bytes in the file and the file data will be stored in the 'Data' field.
+
+The serialized size of a UnixFS node must not exceed 256KiB in order to work will with the [Bitswap] protocol.
+
+## Metadata
+
+UnixFS currently supports two optional metadata fields:
+
+* `mode` -- The `mode` is for persisting the [file permissions in numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) \[[spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html)\]. If unspecified this defaults to `0755` for directories/HAMT shards and `0644` for all other types where applicable.
```
Stebalien: Let's note that, for now, only the `ugo-rwx` bits are defined.

UNIXFS.md
```diff
@@ -59,6 +76,31 @@ For files that are comprised of more than a single block, the 'Type' field will
...
+* `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified.
+
+### Inheritance
+
+When traversing down through a UnixFSv1.5 directory, child entries without metadata fields will inherit those of their direct ascendants.
```
Stebalien: Is this really what we want?

UNIXFS.md
```diff
@@ -86,3 +128,72 @@ If there is only a single chunk, no intermediate unixfs file nodes are created,
...
+In this scheme, the existing `Metadata` message is expanded to include additional metadata types (`mtime`, `mode`, etc).  It then contains links to the actual file data but never the file data itself.
+
+This was ultimately rejected for a number of reasons:
+
+1. You would always need to retrieve an additional node to access file data which limits the kind of optimizations that are possible.
+
+	For example many files are under the 256KiB block size limit, so we tend to inline them into the describing UnixFS `File` node.  This would not be possible with an intermediate `Metadata` node.
+
+1. The `File` node already contains some metadata (e.g. the file size) so you'd end up with metadata stored in multiple places
+
```
Stebalien: also note: this would have had forwards-compatibility issues that would have been annoying.

UNIXFS.md
achingbrain: I think so?  Maybe the wording is wrong - you can have perms set at the root of a tree, which will cascade to all descendants, then you can specialise sections of the tree by setting perms on intermediate directories, and/or you can set perms on individual files for granular control.  You then save space by not having to set the perms on every node.

UNIXFS.md
achingbrain: I've added that.  Also setuid, setgid and the sticky bit for completeness.

UNIXFS.md
achingbrain: If you copied a file that had inherited perms from one UnixFS directory into another, presumably you'd want the option of preserving those perms, so you'd resolve them from the parent and apply them to the resulting node which would change its CID.  I think that's ok.  Bit weird, but ok.

Stebalien: This has the same problems as putting the permissions inside directories instead of inside files: if I give you a CID, it'll have different permissions/timestamps.

achingbrain: I had the same thought while travelling today. I’ll remove this section.

UNIXFS.md
```diff
@@ -59,6 +75,31 @@ For files that are comprised of more than a single block, the 'Type' field will
 
 This data is serialized and placed inside the 'Data' field of the outer merkledag protobuf, which also contains the actual links to the child nodes of this object.
 
+For files comprised of a single block, the 'Type' field will be set to 'File', 'filesize' will be set to the total number of bytes in the file and the file data will be stored in the 'Data' field.
+
+The serialized size of a UnixFS node must not exceed 256KiB in order to work will with the [Bitswap] protocol.
```
Stebalien: not quite: It can be up to 1MiB (bitswap will transfer 2 (4?) MiB blocks).

UNIXFS.md
Stebalien: nit: should probably 23 + a single link to a block.

UNIXFS.md
achingbrain: I've removed this last paragraph, we can always add it in later when the bitswap spec is more fleshed out.

UNIXFS.md
achingbrain: I've pulled this out too because the generated identity CID is too long to be used in DNS labels after all:

```javascript
const CID = require('cids')
const multihashing = require('multihashing-async')
const UnixFS = require('ipfs-unixfs')
const dagPB = require('ipld-dag-pb')
const { DAGLink, DAGNode } = dagPB

async function main () {
  // store some file data
  const fileData = Buffer.from([0, 1, 2, 3, 4])
  const fileDataHash = await multihashing(fileData, 'sha2-256')
  const fileDataCid = new CID(1, 'raw', fileDataHash)

  // create the UnixFS node that points to the file data
  const file = new UnixFS('file')
  file.addBlockSize(5)
  const node = new DAGNode(file.marshal(), [
    new DAGLink('', 5, fileDataCid)
  ])

  // inline the UnixFS node into it's CID
  const multihash = await multihashing(dagPB.util.serialize(node), 'identity')
  const cid = new CID(1, 'dag-pb', multihash)

  console.info(cid.toBaseEncodedString('base32').length)
  // prints '91'
}

main()

```

Stebalien:
:rocket:

PR #229, opened 2019-12-12
docs: change mode and mtime to strings

achingbrain:
As currently defined the closest we can come to removing a `mode`/`mtime` previously set on a file is to set it's value to `0`, which works in that the CID returns to what it would have been pre-metadata, but when combined with default file/dir modes also has the side effect of making it impossible to actually set the file mode to `0`.

Worse, if you set the file mode to `0` you actually get the default file mode back of `0644` which is probably not what you intended.

This is because protocol buffers assign [default values to all fields](https://developers.google.com/protocol-buffers/docs/proto3#default) so we cannot differentiate between the situation where a user has set a thing to `0` vs when they haven't set anything at all.

The proposed solution is to store the `mode`/`mtime` as a string which can then be interpreted as a base 8 number (base 10 for `mtime`), which is a little icky but does solve the problem:

| Value | Interpretation |
| -------- | -------- |
| `''` | No mode has been set |
| `'0'` | Mode has been set to zero |
| `'0755'` | Mode has been set to `0755` |

`mtime` is changed too because if I add some metadata to a file, I should be able to remove it - it shouldn't be a one-way thing.  Unless we are happy with the `mtime` of all entries defaulting to 1970.

It also removes the default `mode` values as otherwise it's not an opt-in upgrade.

### DIFF
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -60,8 +60,8 @@ message Data {
 	repeated uint64 blocksizes = 4;
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
-	optional uint32 mode = 7;
-	optional int64 mtime = 8;
+	optional string mode = 7;
+	optional string mtime = 8;
 }
 
 message Metadata {
@@ -82,11 +82,12 @@ For files comprised of a single block, the 'Type' field will be set to 'File', '
 UnixFS currently supports two optional metadata fields:
 
 * `mode` -- The `mode` is for persisting the file permissions in [numeric notation](https://en.wikipedia.org/wiki/File_system_permissions#Numeric_notation) \[[spec](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html)\].
-  If unspecified this defaults to `0755` for directories/HAMT shards and `0644` for all other types where applicable
+  It is stored as a string that is intended to be interpreted as a `uint32` number in base 8, e.g. `'0544'`
   The nine least significant bits represent  `ugo-rwx`
   The next three least significant bits represent `setuid`, `setgid` and the `sticky bit`
   All others are reserved for future use
 * `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified
+  It is stored as a string that is intended to be interpreted as a `int64` number in base 10
 
 ### Deduplication and inlining
 
@@ -188,7 +189,18 @@ This scheme would see metadata stored in an external database.
 
 The downsides to this are that metadata would not be transferred from one node to another when syncing as [Bitswap] is not aware of the database, and in-tree metadata
 
+#### Why are `mode` and `mtime` stored as strings?
+
+UnixFS uses [protocol buffers] as its internal storage format, which do not have the concept of null, instead [defaulting all fields to certain values](https://developers.google.com/protocol-buffers/docs/proto3#default) - numbers as `0`, strings as `''`, etc.
+
+The problem is `0` is a valid mode for a file, so we need a way of representing a file where no mode has been set, as well as when the user has set the mode to what would have been the default value for the field, if the field was a number.
+
+So we use `''` for when no mode has been set, `'0'` for when the mode was set to `0`, `'0644'` for when the mode was set to `0644`, etc.
+
+Similarly `''` represents the absence of an `mtime`, `'0'` is the start of the Unix Epoch, `'1'` is one second after the start of the Unix Epoch.
+
 [multihash]: https://tools.ietf.org/html/draft-multiformats-multihash-00
 [CID]: https://docs.ipfs.io/guides/concepts/cid/
 [Bitswap]: https://github.com/ipfs/specs/blob/master/BITSWAP.md
 [MFS]: https://docs.ipfs.io/guides/concepts/mfs/
+[protocol buffers]: https://developers.google.com/protocol-buffers


achingbrain:
cc @ribasushi who still doesn't appear in the review box for reasons beyond me.

alanshaw:
String for `mode` feels like the least worst solution (@achingbrain and I discussed a few) but I'm open to other suggestions.

Not sure I agree/understand the reasoning to do it for `mtime` as well.

achingbrain:
> Not sure I agree/understand the reasoning to do it for mtime as well.

The reason behind my thinking is that metadata should be opt-in, it should be as correct as possible and that we should be able to undo everything that we do.

So:

```console
$ jsipfs files mkdir /foo                          // make a dir
$ jsipfs files stat /qux                           // what's the CID?
QmUNLLsPACCz1vLxQVkXqqLX5R1X345qqfHbsf67hvA3Nn
... other stuff
$ jsipfs files ls -l
----------	-	qux 0                      // no mode or mtime
$ jsipfs files touch /qux
$ jsipfs files ls -l
---------- Dec 12, 2019, 02:08:08 PM GMT qux 0     // has mtime
$ jsipfs files stat /qux                           // what's the CID now?
QmZzCcaozsXHTgzMavQcYLCVqyWm6dfAGVDMdrK2CEG74x
```

How do I get back to `QmUNLLs...`?

```console
$ jsipfs files touch -m 0 /qux
$ jsipfs files ls -l
----------	-	qux 0                      // no mode or mtime
```

What if I wanted `mtime` to be `0`?

We could display `Jan 1, 1970, 01:00:00 AM GMT+1` for `mtime` whenever it's `0`, but that would give everything a default `mtime` that was not necessarily the time it was last modified which seems wrong to me.

aschmahmann:
@achingbrain to clarify do we want nullable, optional or both (take a look at https://specs.ipld.io/schemas/schema-kinds.html#value-type-modifiers) for both mtime and mode?

Once we know that we can figure out what we want the representation to be. We could use strings, we could use a union, etc. depending on how compact vs readable we want the data to be.

dirkmc:
Would it be possible to use an enum for mode?

ribasushi:
> Value	Interpretation
> `''`	No mode has been set
> `'0'`	Mode has been set to zero
> `'0755'`	Mode has been set to `0755`

This feels **really** icky... :(

> This is because protocol buffers assign [default values to all fields](https://developers.google.com/protocol-buffers/docs/proto3#default) so we cannot differentiate between the situation where a user has set a thing to `0` vs when they haven't set anything at all.

You could simply do one level of indrection ( will cost 4 extra bytes per metadata entry, nothing in the granrd scheme of things). Then you can say "is the structure nil? - moving on".

Essentially one of old proposal mirroing the Metadata subtype:
```
message Data {
...
	required DataType Type = 1;
	optional bytes Data = 2;
	optional uint64 filesize = 3;
	optional uint64 hashType = 5;
	optional uint64 fanout = 6;
        optional PosixMeta = 7;
}

message PosixMeta {
	optional uint32 mode = 1;
        // it just occured to me wjile typing this, that we can have cake and eat it:
        // no value = no mtime
        // 1 value = epoch based seconds
        // 2 values = high resolution mtime for these who want it
        // 3+ values = reject as invalid
	repeated int64 mtime = 2;  
}

**Comment by achingbrain** [comment]
> do we want nullable, optional or both

Ideally it'd be optional, but optional values in protocol buffers always have a value so it doesn't seem possible.

**Comment by achingbrain** [comment]
> Would it be possible to use an enum for mode?

I don't think so, not from a practical perspective as you'd have to define enum values for every value of a uint32 + 'not set' in the worst case and 0000-7777 (base 8) + 'not set' in the best case.

**Comment by achingbrain** [comment]
> You could simply do one level of indirection

Actually this is not terrible.  The protobuf docs say:

> For message fields, the field is not set. Its exact value is language-dependent.

..so we can set them as null values:

```
message Data {
...
	required DataType Type = 1;
	optional bytes Data = 2;
	optional uint64 filesize = 3;
	optional uint64 hashType = 5;
	optional uint64 fanout = 6;
        optional Mode mode = 7;
        optional Mtime mode = 8;
}

message Mode {
	optional uint32 value = 1;
}

message Mtime {
        // 1 value = epoch based seconds
        // 2 values = high resolution mtime for these who want it
        // 3+ values = reject as invalid
	repeated int64 value = 1;
}
```

**Comment by achingbrain** [comment]
Sample implementation of using protobuf messages for mtime and mode: https://github.com/ipfs/js-ipfs-unixfs/pull/38

**Comment by aschmahmann** [comment]
> Ideally it'd be optional

Are you sure, it sounds like you mean `nullable` (or really `optional + nullable`)? Protobuf already gives us `optional` which allows us to process structs where the optional field is missing (we like this property). What we're missing is a way of encoding that we want to know if that value was set whether it was set to null/ignore me or a real value.

On to the byte hackery stage where we can figure out if/what a compact representation of this data is and then decide if we'd prefer a few extra bits for sanity's sake or not.

If `mtime` and `mode` each completely fill up a byte (i.e. their data spans a multiple of 2^8 values) and we want to express both `optional mtime` and `optional mode` then we need to encode two more values. This means adding some bytes somewhere.

We can add the extra bytes per metadata field (like with the optional suggestion above), or we could even group them together and have an enum defining whether `mtime`, `mode`, both or neither are set. I'm pretty sure if we wanted we could even have it setup so that we don't have extra bytes with no metadata or with non-zero metadata, but only have extra bytes in the unlikely events of mtime = Jan 1 1970, or mode = 0.

**Comment by achingbrain** [comment]
> Are you sure, it sounds like you mean nullable (or really optional + nullable)?

I think so?  I want either a value that means it's been set (`0`) or the lack of a value that means it hasn't been set (`undefined`), not a value that isn't a value (`null`).

> Protobuf already gives us optional which allows us to process structs where the optional field is missing 

I think this is what I was missing - that unlike simple values (numeric types, strings, etc) optional Messages do not get default values, instead it's up to the language to specify what value they have.

**Comment by aschmahmann** [comment]
A fairly compact, but maybe annoying option:

```
message Data {
         ...
        optional PosixMetadata pmeta = 7;
}

message PosixMetadata {
        enum OverrideDefaults {
               NONE = 0;
               ALL = 1;
               MODE = 2;
               MTIME = 3;
        }
        optional OverrideDefaults  override = 1;
	optional uint32 mode = 2;
        optional int64 mtime = 3;
}
```

Then if we actually wanted to use the `mtime = 0` or `mode = 0` we could add the flags to explicitly define that. This is gross and just a first stab at the problem, but I think is fairly compact on the wire if that's what we're interested in.

achingbrain:
Hmm, I'd prefer plain values if possible - it makes everything a bit simpler.

Is it significantly more efficient than the modified version of @ribasushi's proposal in https://github.com/ipfs/specs/pull/229#issuecomment-565068736 ?

achingbrain:
Superseded by #230

PR #235, opened 2019-12-23
docs: support high resolution mtime metadata

achingbrain:
Stores modification times as two values seconds and nanoseconds since or before the unix epoch.

The `EpochSeconds` field represents a fraction of a second rather nanoseconds since the epoch.

Developed from mine and @ribasushi's conversation on https://github.com/ipfs/specs/pull/232#discussion_r359897261

### DIFF
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -61,7 +61,12 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	optional uint32 mode = 7;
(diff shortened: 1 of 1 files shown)


ribasushi:
🤦‍♂ I did not see this PR before I opened mine :(

Mine has a bit more prose, and the data type is slightly different ( with a comment why ). Wat do we do? :)

achingbrain:
Closing in favour of #236

PR #236, opened 2019-12-23
Add timespec structured mtime metadata

ribasushi:
Augment the mtime specification with a well-defined universal datastructure capable of expressing both arbitrarily distant points in time, and sufficiently arbitrary ( in computing terms ) sub-second precision. The very slight increase in on-wire size and the mild extra parsing complexity allows for:

- A structure representing time, that is unlikely to change any time in the future, aiding implementation convergence
- A way to encode any filesystem-related timestamp within the wider problem domain of information systems

### DIFF
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -61,11 +61,24 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	optional uint32 mode = 7;
-	optional int64 mtime = 8;
+	optional TimSpec mtime = 8;
 }
 
 message Metadata {
 	optional string MimeType = 1;
+}
+
+message TimeSpec {
+	// Use of int64 is deliberate - negative epoch is super unlikely
+	// Using sint64 would introduce zig-zag encoding ( harder to eyeball )
+	// The varint representing the time of writing this is 5 bytes long
+	// It will remain so until October 26, 3058 ( 34,359,738,367 )
+	required int64 EpochSeconds = 1;
+
+	// fixed32 ( always 4 bytes, no varint ), as nanosecs are often > 2^28
+	// https://developers.google.com/protocol-buffers/docs/proto#scalar
+	// NOTE: on the wire this value will be *little* endian
+	optional fixed32 EpochNanoseconds = 2;
 }
 ```
 
@@ -90,7 +103,9 @@ UnixFS currently supports two optional metadata fields:
   - The remaining 20 bits are reserved for future use, and are subject to change. Spec implementations **MUST** handle bits they do not expect as follows:
     - For future-proofing the (de)serialization layer must preserve the entire uint32 value during clone/copy operations, modifying only bit values that have a well defined meaning: `clonedValue = ( modifiedBits & 07777 ) | ( originalValue & 0xFFFFF000 )`
     - Implementations of this spec must proactively mask off bits without a defined meaning in the implemented version of the spec: `interpretedValue = originalValue & 07777`
-* `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified
+* `mtime` -- A two-element structure ( `EpochSeconds`, `EpochNanoseconds` ) representing the modification time in seconds relative to the unix epoch `1970-01-01T00:00:00Z`. In contexts where an mtime is mandatory ( e.g. FUSE interfaces ) implementations must treat an unspecified mtime as `0`.
+  - `EpochSeconds` represents the amount of seconds after **or before** the epoch. Implementations must be able to gracefully handle negative mtime, even if such a value is not applicable within their domain ( e.g. a POSIX filesystem )
+  - `EpochNanoseconds` represents the fractional part of the mtime as the amount of nanoseconds. The valid range for this value is the integer range `[1, 999999999]`. If a fractional part outside of this range is encountered, implementations should consider the entire metadata block invalid and abort processing it. Note that **a fractional value of `0` is NOT valid** - omit the nanosecond value altogether to represent whole seconds.
 
 ### Deduplication and inlining
 
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -61,7 +61,7 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	optional uint32 mode = 7;
-	optional TimSpec mtime = 8;
+	optional TimeSpec mtime = 8;
 }
 
 message Metadata {
@@ -69,15 +69,8 @@ message Metadata {
 }
 
 message TimeSpec {
-	// Use of int64 is deliberate - negative epoch is super unlikely
-	// Using sint64 would introduce zig-zag encoding ( harder to eyeball )
-	// The varint representing the time of writing this is 5 bytes long
-	// It will remain so until October 26, 3058 ( 34,359,738,367 )
 	required int64 EpochSeconds = 1;
 
-	// fixed32 ( always 4 bytes, no varint ), as nanosecs are often > 2^28
-	// https://developers.google.com/protocol-buffers/docs/proto#scalar
-	// NOTE: on the wire this value will be *little* endian
 	optional fixed32 EpochNanoseconds = 2;
 }
 ```
@@ -206,6 +199,21 @@ This was rejected due to concerns about added complexity, recovery after system 
 This scheme would see metadata stored in an external database.
 
 The downsides to this are that metadata would not be transferred from one node to another when syncing as [Bitswap] is not aware of the database, and in-tree metadata
+
+### TimeSpec protobuf datatype rationale
+
+#### EpochSeconds
+
+The integer portion of the epoch is represented on the wire using a varint encoding. While this is inefficient for
+negative values, it avoids introducing zig-zag encoding. Negative epoch values will be exceedingly rare, and there
+could very well be value in having such cases stand out, while at the same keeping the "usual" positive values easy
+to eyeball. The varint representing the time of writing this text is 5 bytes long. It will remain so until
+October 26, 3058 ( 34,359,738,367 )
+
+#### EpichNanoseconds
+Since fractional values will very often be > 2^28 nanoseconds, that part is represented as a 4-byte `fixed32`,
+[as per google's recommendation](https://developers.google.com/protocol-buffers/docs/proto#scalar).
+
 
 [multihash]: https://tools.ietf.org/html/draft-multiformats-multihash-00
 [CID]: https://docs.ipfs.io/guides/concepts/cid/
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -61,17 +61,16 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	optional uint32 mode = 7;
-	optional TimeSpec mtime = 8;
+	optional UnixTime mtime = 8;
 }
 
 message Metadata {
 	optional string MimeType = 1;
 }
 
-message TimeSpec {
-	required int64 EpochSeconds = 1;
-
-	optional fixed32 EpochNanoseconds = 2;
+message UnixTime {
+	required int64 Seconds = 1;
+	optional fixed32 FractionalNanoseconds = 2;
 }
 ```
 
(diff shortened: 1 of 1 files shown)


UNIXFS.md
```diff
@@ -61,12 +61,25 @@ message Data {
 	optional uint64 hashType = 5;
 	optional uint64 fanout = 6;
 	optional uint32 mode = 7;
-	optional int64 mtime = 8;
+	optional TimSpec mtime = 8;
```
achingbrain:
```suggestion
	optional TimeSpec mtime = 8;
```

UNIXFS.md
```diff
@@ -61,12 +61,25 @@ message Data {
...
 }
 
 message Metadata {
 	optional string MimeType = 1;
 }
+
+message TimeSpec {
+	// Use of int64 is deliberate - negative epoch is super unlikely
```
achingbrain: This sort of discussion should move to the `Design decision rationale` section

UNIXFS.md
```diff
@@ -61,12 +61,25 @@ message Data {
...
+	// Using sint64 would introduce zig-zag encoding ( harder to eyeball )
+	// The varint representing the time of writing this is 5 bytes long
+	// It will remain so until October 26, 3058 ( 34,359,738,367 )
+	required int64 EpochSeconds = 1;
+
+	// fixed32 ( always 4 bytes, no varint ), as nanosecs are often > 2^28
```
achingbrain: This sort of discussion should move to the `Design decision rationale` section

UNIXFS.md
```diff
@@ -90,7 +103,9 @@ UnixFS currently supports two optional metadata fields:
   - The remaining 20 bits are reserved for future use, and are subject to change. Spec implementations **MUST** handle bits they do not expect as follows:
     - For future-proofing the (de)serialization layer must preserve the entire uint32 value during clone/copy operations, modifying only bit values that have a well defined meaning: `clonedValue = ( modifiedBits & 07777 ) | ( originalValue & 0xFFFFF000 )`
     - Implementations of this spec must proactively mask off bits without a defined meaning in the implemented version of the spec: `interpretedValue = originalValue & 07777`
-* `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified
+* `mtime` -- A two-element structure ( `EpochSeconds`, `EpochNanoseconds` ) representing the modification time in seconds relative to the unix epoch `1970-01-01T00:00:00Z`. In contexts where an mtime is mandatory ( e.g. FUSE interfaces ) implementations must treat an unspecified mtime as `0`.
```
achingbrain: I think we should just default this to `0` as it is currently.  The `go-ipfs` and `js-ipfs` implementations cannot control which context they will be used from so should default to the most conservative approach so that might as well be in the spec.

UNIXFS.md
```diff
@@ -90,7 +103,9 @@ UnixFS currently supports two optional metadata fields:
...
+  - `EpochSeconds` represents the amount of seconds after **or before** the epoch. Implementations must be able to gracefully handle negative mtime, even if such a value is not applicable within their domain ( e.g. a POSIX filesystem )
+  - `EpochNanoseconds` represents the fractional part of the mtime as the amount of nanoseconds. The valid range for this value is the integer range `[1, 999999999]`. If a fractional part outside of this range is encountered, implementations should consider the entire metadata block invalid and abort processing it. Note that **a fractional value of `0` is NOT valid** - omit the nanosecond value altogether to represent whole seconds.
```
achingbrain: Why not allow fractional values of `0`?

UNIXFS.md
ribasushi: My hesitation placing this in the spec is that it implicitly means `unknown mtime is not a thing`. E.g. a web gateway would be obligated to send out a `0` `Last-Modified` header, and so on...

What do you think?

UNIXFS.md
ribasushi: In order to minimize amount of different ways encoding the same thing.
Since we need to validate the upper side of the range anyway, might as well validate the lower part too.
I am not opposed dropping this requirement but it seems odd to allow ambiguity.

UNIXFS.md
ribasushi: Fixing...

UNIXFS.md
```diff
@@ -90,7 +96,9 @@ UnixFS currently supports two optional metadata fields:
   - The remaining 20 bits are reserved for future use, and are subject to change. Spec implementations **MUST** handle bits they do not expect as follows:
     - For future-proofing the (de)serialization layer must preserve the entire uint32 value during clone/copy operations, modifying only bit values that have a well defined meaning: `clonedValue = ( modifiedBits & 07777 ) | ( originalValue & 0xFFFFF000 )`
     - Implementations of this spec must proactively mask off bits without a defined meaning in the implemented version of the spec: `interpretedValue = originalValue & 07777`
-* `mtime` -- The modification time in seconds since the epoch. This defaults to the unix epoch if unspecified
+* `mtime` -- A two-element structure ( `EpochSeconds`, `EpochNanoseconds` ) representing the modification time in seconds relative to the unix epoch `1970-01-01T00:00:00Z`. In contexts where an mtime is mandatory ( e.g. FUSE interfaces ) implementations must treat an unspecified mtime as `0`.
```
alanshaw: Minor suggestion, but do we have to include “Epoch” in these names?

alanshaw: Also, EpochNanoseconds is _not_ the number of nanoseconds since the unix epoch as the name may suggest

UNIXFS.md
```diff
@@ -192,6 +200,21 @@ This scheme would see metadata stored in an external database.
...
+#### EpochSeconds
+
+The integer portion of the epoch is represented on the wire using a varint encoding. While this is inefficient for
+negative values, it avoids introducing zig-zag encoding. Negative epoch values will be exceedingly rare, and there
+could very well be value in having such cases stand out, while at the same keeping the "usual" positive values easy
+to eyeball. The varint representing the time of writing this text is 5 bytes long. It will remain so until
+October 26, 3058 ( 34,359,738,367 )
+
+#### EpichNanoseconds
+Since fractional values will very often be > 2^28 nanoseconds, that part is represented as a 4-byte `fixed32`,
```
alanshaw: Not sure I understand the maths behind why they’ll often be this big?

alanshaw:
Please can we have some description in the PR description to explain why this is required?

achingbrain:
> Not sure I understand the maths behind why they’ll often be this big?

2^28 is roughly 1/4 of the total range of nanosecond values so assuming a uniform distribution they'll be bigger than that 3/4 of the time.

UNIXFS.md
achingbrain: I think that's fine.  At least, I can't think of a reason that makes it not fine, given that [we need a default value](https://github.com/ipfs/specs/pull/232#discussion_r360469535) and we're happy with seeing 1970 everywhere when we `ipfs files ls -l`.

UNIXFS.md
achingbrain: Maybe something like this?

```
message EpochTime {
	required int64 Seconds = 1;
	optional fixed32 NanosecondFragment = 2;
}
```

UNIXFS.md
Stebalien: The nanosecond part is essentially a random number between 0 and 1e9.

* 1/2 the time, you'll need 29 bits (`ceil(logtwo(1e9/2))`) to represent this. That's 4 bytes when using fixed32 and 5 bytes when using a varint.
* 99% of the time, you'll need at least 23 bits to represent it (`logtwo(1e9/100)`). That's 4 bytes when using a varint.

UNIXFS.md
alanshaw: +1 - I'd be ok with this suggestion.

ribasushi: Hm... in order to be explicit which `Epoch` are we talking about, and to not sound dorky, how about:
```
message UnixTime {
	required int64 Seconds = 1;
	optional fixed32 FractionalNanoSeconds = 2;
}
```

?

UNIXFS.md
```diff
@@ -192,6 +200,21 @@ This scheme would see metadata stored in an external database.
...
+
+#### EpochSeconds
+
+The integer portion of the epoch is represented on the wire using a varint encoding. While this is inefficient for
+negative values, it avoids introducing zig-zag encoding. Negative epoch values will be exceedingly rare, and there
+could very well be value in having such cases stand out, while at the same keeping the "usual" positive values easy
+to eyeball. The varint representing the time of writing this text is 5 bytes long. It will remain so until
+October 26, 3058 ( 34,359,738,367 )
+
+#### EpichNanoseconds
```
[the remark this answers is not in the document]
ribasushi: I stand ashamed beside my spelling fu :/
Will fix when we settle on the fieldnames

UNIXFS.md
ribasushi: @alanshaw sorry for the delay answering, but yes: what they said ;)

UNIXFS.md
achingbrain: To make this more concrete, I think a web gateway sending a `0` `Last-Modified` header is ok.

UNIXFS.md
achingbrain: LGTM

UNIXFS.md
ribasushi: :(

I will relent if this is a wider consensus, but then it makes me sad as this part of the spec becomes unusable to me as well. In my use case having a gateway render "this is the mtime" vs "mtime was not supplied" is really crucial.

UNIXFS.md
ribasushi: @alanshaw can I get a +1 on the naming proposed at https://github.com/ipfs/specs/pull/236#discussion_r363317171 so we can augment/close this PR?

alanshaw: haha, much better 🚢🚢🚢

(I was debating whether it was worth bringing up the specific epoch issue but I had decided I it didn't matter too much 🤣)

ribasushi:
Ok so... this tortured PR is now blocked on last two issues:

- Range of the fractional nanosecond value. I read this as protocol-level spec, I am advocating for "0 nanoseconds" to be an invalid value. @achingbrain comes from implementation point of view, and correctly points out that a user should not care whether a value is 0 or not ( https://github.com/ipfs/js-ipfs-unixfs/pull/40#discussion_r363389380 )

  - As resolution I propose adding verbiage to the spec that spells out what is valid on the wire vs what is valid on the API side. I have not added to this PR, as there seems to be aversion to a wordy doc

- The question over whether "no mtime" automatically implies `0` for all intents and purposes, or whether implementations ( e.g. a web gateway ) *may* elect to render `actual 0` and "no data" in distinct ways. @achingbrain advocates for `0`/`1970-01-01` to be mandated by the spec, including web-gateways being obligated to always supply a `Last-Modified: ` header. I find this unworkable for the archival use-cases I have in mind, and am strongly for leaving this at the implementation's discretion ( https://github.com/ipfs/specs/pull/236#discussion_r360957112 )

  - Resolution is difficult, as the outcome is zero-sum. I guess requesting for @alanshaw and @Stebalien to break the tie ;)

Stebalien:
> As resolution I propose adding verbiage to the spec that spells out what is valid on the wire vs what is valid on the API side. I have not added to this PR, as there seems to be aversion to a wordy doc

I agree that we should not _encode_ 0 nanoseconds on the wire.

> The question over whether "no mtime" automatically implies 0 for all intents and purposes, or whether implementations ( e.g. a web gateway ) may elect to render actual 0 and "no data" in distinct ways. 

I'd let implementations decide what to do here. However, we should specify _what_ they can do. They can either:

* Preferred: indicate that the time is unset.
* Otherwise: return the epoch. Unfortunately, some APIs make it difficult to indicate that no mtime is set.

That is, they can't just return _any_ random time. They should also document what they do.

Stebalien:
Merging per @ribasushi's suggestion as we can address the final two points in a followup PR.

achingbrain:
On mtime:

>  @achingbrain advocates for 0/1970-01-01 to be mandated by the spec

The reason for this is primarily [this comment](https://github.com/ipfs/specs/pull/232#discussion_r360469535) and that prior metadata PRs have [specified a default value](https://github.com/ipfs/specs/pull/223/files#diff-fd29a4d2fe4d67e820006ccae3a7abcfR72).

I am more than happy to make this field optional, and solid use cases for this definitely strengthen that argument.  It also means the mtime field in the protobuf definition doesn't need extra exposition around it which is good too.

ribasushi:
@achingbrain @Stebalien see #237 for the next attempt at expressing this.

PR #253, opened 2020-12-04
Remove todos by linking to related specs

momack2:


willscott (approved):
Thanks for pulling outdated docs up to current!

PR #265, opened 2022-01-31
Define less ambiguous UnixFS spec in TS types

Gozala:
I would like to include less ambiguous UnixFS specification coded as TS interface definition so that:

1. It can capture more invariant than possible via provided protobuf schema.
2. It is clear which fields are required in which node types.
3. It is more clear how file shards, chunks & roots are represented (in terms of necessary fields and link types)

You may notice several `@TODO` comments as despite my best efforts I still have gaps in my understanding. I would love if you could help me fill those by providing inline comments and/or suggestions.

Despite my best effort, I may still be misunderstanding and incorrectly encoding / describing things, if so I would really appreciate comments and/or suggestions to correct those errors.

I would like to iterate on this document and eventually land this for everyone who may try to get a better understanding of the UnixFS format.

### DIFF
--- a/unixfs.ts
+++ b/unixfs.ts
@@ -1,0 +1,489 @@
+/**
+ * Logical representation of a file chunk.
+ * 
+ * TODO: Clarify when this represenation is used as opposed to `FileChunk`.
+ */
+export interface RawChunk extends PBNode {
+  Data: ByteView<{
+    Type: DataType.Raw,
+    /**
+     * Raw bytes of the content
+     */
+    Data: Bytes
+  }>
+  /**
+   * Raw nodes MUST not have any links, yet empty `Links` list is expected.
+   * At the type level it is expressed as `never[]` which guarantees that
+   * no instatiation other than empty will satisfy this constraint
+   */
+  Links: never[]
+}
+
+/**
+ * Logical representation of a file chunk. When large file is added to IPFS
+ * it gets chunked into smaller pieces (according to specified `--chunker`)
+ * and each chunk is encoded into this representation (and linked from file
+ * DAG).
+ * 
+ * Note: While technically it is possible to add `mode` and `mtime` to
+ * `FileChunk` node it is logical nonsense and therefor to be ignore.
+ */
+export interface FileChunk extends PBNode {
+  Data: ByteView<{
+    Type: DataType.File,
+    /**
+     * Raw bytes corresponding to this file chunk
+     */
+    Data: Bytes
+    /**
+     * Number of bytes in Data field
+     */
+    filesize: uint64
+  }>
+  /**
+   * Raw nodes MUST not have any links, yet empty `Links` list is expected.
+   * At the type level it is expressed as `never[]` which guarantees that
+   * no instatiation other than empty will satisfy this constraint
+   */
+  Links: never[]
+}
+
+export type FileLeaf =
+  /**
+   * May link to raw block (Not to confuse with UnixFS Raw nodes). Happens
+   * when `--raw-leaves` option is used.
+   */
+  | RawNode
+  /**
+   * @TODO - I have no idea when this happens please help!
+   */
+  | RawChunk
+  /**
+   * Node links to actual chunks when it's a level above leaf nodes.
+   */
+  | FileChunk
+
+/**
+ * Logical representation of a file shard. When large files are chunked
+ * slices that span multiple chunks may be represented via file shards.
+ * 
+ * Please note that some file layouts may create shallow DAGs where file root
+ * node links to `FileChunk|Raw` nodes and in other layouts DAGs may be
+ * several levels deep, in those cases file root link to `FileShard` nodes
+ * that link to either other shards or leaf `FileChunk|Raw` nodes.
+ * 
+ * Note: While technically it is possible to add `mode` and `mtime` to
+ * `FileShard` nodes it is logical nonsense and therefor to be ignore.
+ */
+export interface FileShard extends PBNode {
+  Data: ByteView<{
+    Type: DataType.File,
+    /**
+     * The total number of bytes of a file slize represented by this shard.
+     */
+    filesize: uint64
+    /**
+     * List of `filesize`s for each linked node (in exact same order).
+     */
+    blocksizes: uint64[]
+  }>
+  /**
+   * Links to the file slices this shard is comprised of.
+   * Note: That this is heterogeneous list as e.g. in trickle DAG layout
+   * shards may link to both leaves and other shards.
+   */
+  Links: FileLink<FileLeaf|FileShard>[]
+}
+
+/**
+ * Logical representation of a file that fits a single block. Note is
+ * semantically different from a `FileChunk`, even though structurally it is
+ * compatible. Unlike `FileChunk` it may contain `mode`, `mtime` file metadata.
+ */
+export interface SimpleFileLayout extends PBNode {
+  Data: ByteView<{
+    Type: DataType.File,
+    /**
+     * Raw bytes of the file content
+     */
+    Data: Bytes
+    /**
+     * Number of bytes in above `Data` field
+     */
+    filesize: uint64,
+
+    /**
+     * If omitted to be interprented as default `0644`. It is RECOMMENDED
+     * to omit if mode matches default.
+     */
+    mode?: Mode
+    mtime?: UnixTime
+  }>
+
+  /**
+   * MUST not have any links, yet empty `Links` list is expected.
+   * At the type level it is expressed as `never[]` which guarantees that
+   * no instatiation other than empty will satisfy this constraint
+   */
+  Links: never[]
+}
+
+/**
+ * Logical represenatation of a file that consists of multiple blocks. Note is
+ * semantically different from a `FileShard`, even though structurally it is
+ * compatible. Unlike `FileShard` it may contain `mode`, `mtime` file metadata.
+ */
+
+export interface AdvancedFileLayout extends PBNode {
+  Data: ByteView<{
+    Type: DataType.File,
+    // Total number of bytes in the file (not the graph structure).
+    filesize: uint64,
+
+    /**
+     * List of `filesize`s for each linked node (in exact same order).
+     */
+    blocksizes: uint64[]
+
+    /**
+     * If omitted to be interprented as default `0644`. It is RECOMMENDED
+     * to omit if mode matches default.
+     */
+    mode?: Mode
+    mtime?: UnixTime
+  }>
+  /**
+   * Links to the file slices this file is comprised of.
+   * Note: That this is heterogeneous list as e.g. in trickle DAG layout
+   * shards may link to both leaves and other shards.
+   */
+  Links: FileLink<RawNode|FileShard>[]
+}
+
+/**
+ * In IPFS large files are chucked into several blocks for a more effective
+ * replication. Such files in UnixFS are represented via `AdvancedFileLayout`.
+ * And files that fit into a single block are represented via `SimpleFileLayout`.
+ * 
+ * Please note: In some configurations files that fit a single block are not
+ * even encoded as UnixFS but rather as a `RawNode` blocks. However this type
+ * describes UinxFS File representation and not IPFS file representation which
+ * is why `RawNode` variant is not part of it.
+ */
+export type FileLayout =
+  | SimpleFileLayout
+  | AdvancedFileLayout
+
+
+/**
+ * Logacal representation of a directory that fits single block.
+ */
+export interface FlatDirectoryLayout extends PBNode {
+  Data: ByteView<{
+    Type: DataType.Directory
+    /**
+     * Directories MUST have `filesize` set to `0`.
+     */
+    filesize: 0
+    /**
+     * If omitted to be interpreted as default `0755`. It is RECOMMENDED
+     * to omit if mode matches default.
+     */
+    mode?: Mode
+    mtime?: UnixTime
+  }>
+  /**
+   * Links are directory entries.
+   */
+  Links: DirectoryLink<FileLayout|DirectoryLayout>[]
+}
+
+
+/**
+ * Logical representation of directory encoded in multiple blocks (usually when
+ * it contains large number of entries). Such directories are represented via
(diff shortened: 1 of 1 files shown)


Gozala:
Got no permissions to request review so instead I'm tagging people by names @aschmahmann @achingbrain @rvagg @alanshaw @ribasushi

achingbrain:
I'm not sure introducing typescript into the spec is a good idea?  We already have prose and protobuf as definition languages, adding a third just increases the probability of contradictions between them all.

It seems like the problem this is aiming to solve is that lots of fields can be set, most of them are optional and it's not clear which ones should be set and when, and what to do when they are or are not - can we not just make the clarifications by updating the body of the text & adding comments to the protobuf parts?

unixfs.ts
```diff
@@ -0,0 +1,489 @@
...
+/**
+ * Logical representation of directory encoded in multiple blocks (usually when
+ * it contains large number of entries). Such directories are represented via
+ * Hash Array Map Tries (HAMT).
+ * 
+ * @see https://en.wikipedia.org/wiki/Hash_array_mapped_trie
+ */
+export interface AdvancedDirectoryLayout {
+  Data: ByteView<{
+    // TODO: Is root still a directory type or is it going to be HAMT instead ?
```
Gozala, line 211:
```suggestion
```

unixfs.ts
```diff
@@ -0,0 +1,489 @@
...
+    Type: DataType.Directory,
```
Gozala, line 212:
```suggestion
    Type: DataType.HAMTShard,
```

unixfs.ts
```diff
@@ -0,0 +1,481 @@
...
+export type FileNode =
+  | FileChunk
+  | FileLayout
+
+/**
+ * All nodes that have `directory` type.
+ */
+export type DirectoryNode =
+  | FlatDirectoryLayout
+  | AdvancedDirectoryLayout
```
alanshaw, line 277: Currently the same as `DirectoryLayout`?

unixfs.ts
```diff
@@ -0,0 +1,481 @@
...
+     * Number of bytes in Data field
+     */
+    filesize: uint64
+  }>
+  /**
+   * Raw nodes MUST not have any links, yet empty `Links` list is expected.
+   * At the type level it is expressed as `never[]` which guarantees that
+   * no instatiation other than empty will satisfy this constraint
+   */
+  Links: never[]
```
alanshaw, line 48: Is this correct here?

unixfs.ts
```diff
@@ -0,0 +1,481 @@
+/**
+ * Logical representation of a file chunk.
+ * 
+ * TODO: Clarify when this represenation is used as opposed to `FileChunk`.
```
alanshaw, line 4: This is the `leafType` option of the importer and I believe unused unless you specify it. If you specify `raw` for `leafType` you'll get these type of nodes at the leaves of the DAG.

https://github.com/ipfs/js-ipfs-unixfs/tree/master/packages/ipfs-unixfs-importer#api

Gozala: I'm aware of this, although I don't see corresponding option in `ipfs` CLI and generally don't understand pros/cons between choosing file vs raw for chunks.

Is one of them legacy and mostly for backwards compat ? Or is there legitimate reason to have both and one should / would choose one over the other ?

unixfs.ts
Gozala: > Is this correct here?

Comment about types is accurate, however I'm not sure if the claim itself that it should be empty is accurate. According to dag-pb spec `Links` should be present even if empty as per:
https://ipld.io/specs/codecs/dag-pb/spec/#constraints

> Links:
>
>    - must be present, even if empty; the binary form makes no distinction between an empty array and an omitted value, in the Data Model we always instantiate an array.
>    - elements must be sorted in ascending order by their Name values, which are compared by bytes rather than as strings.
>        - Names must be unique or be omitted.

If raw blocks can actually have links, would be great to point that out and also elaborate and provide an example when / what to do with those.

Gozala:
> I'm not sure introducing typescript into the spec is a good idea? We already have prose and protobuf as definition languages, adding a third just increases the probability of contradictions between them all.

I think it's a lot easier to resolve these contradictions canonically here than hope that those will not arise in the implementations. If there was a language neutral way to specify this I would definitely prefer that over TS, but I don't believe one exists and I would much rather have something that:

1. I can run program against to verify.
2. Translate to my language.

My motivation for writing this is precisely protobufs inability to disambiguate, even after going through the both JS and Go implementations I'm not certain I've captured it all accurately. I don't think it's lack of my abilities at fault, I think things are just too vaguely specified.

unixfs.ts
Gozala: oops, thanks for the catch

unixfs.ts
```diff
@@ -0,0 +1,481 @@
...
+   * TODO: Have not came across this one would be nice to either mark
+   * or entype it's represenation deprecated
+   */
+  Metadata = 3,
+  /**
+   * TODO: Have not came across this one either, I'm not sure how it supposed
+   * to be represented. If not used in practice maybe it sholud be marked
+   * deprecated.
+   */
+  Symlink = 4,
```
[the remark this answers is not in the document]
Gozala: Thanks @alanshaw! After digging through code and tests a bit more I got an impression that Symlinks just encode path to the target in a data. I'm still not sure if `blockSizes` is just js-unixfs normalization artifact or if they are expected to be encoded in the dag. I also came across the following test

https://github.com/ipfs/js-ipfs-unixfs/blob/master/packages/ipfs-unixfs/test/unixfs-format.spec.js#L439-L446

Which refers to @daviddias's comment
https://github.com/ipfs-inactive/js-ipfs-unixfs-engine/issues/3#issuecomment-182440079

In the closed issue which points to dead links.


So does anyone know what `fliesize` in `Symlink` should report ?

unixfs.ts
```diff
@@ -0,0 +1,481 @@
...
+
+/**
+ * Less loosely defined PB Link that requires `Name` field.
+ */
+export interface NamedLink<Data> extends PBLink<Data> {
+  Hash: CID<Data>
+  /**
+   * Directory link SHOULD specify size of the entry.
+   */
+  Tsize: uint64
```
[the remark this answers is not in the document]
Gozala: I thought it should not, should it from what I can tell all the Dir links seem to report sizes and as far as I can tell that is size of the block they point at not the size of the content.

unixfs.ts
```diff
@@ -0,0 +1,484 @@
...
+
+/**
+ * @TODO
+ */
+export type MetadataNode = never
+
+/**
+ * @TODO
+ */
+export interface Symlink extends PBNode {
```
Gozala, line 274: Is this accurate ? Given following links I have little confidence 

https://github.com/ipfs/js-ipfs-unixfs/blob/master/packages/ipfs-unixfs/test/unixfs-format.spec.js#L439-L446
https://github.com/ipfs-inactive/js-ipfs-unixfs-engine/issues/3#issuecomment-182440079

unixfs.ts
```diff
@@ -0,0 +1,488 @@
...
+  Data: ByteView<{
+    Type: DataType.Symlink,
+    /**
+     * UTF-8 encoded path to the symlink target.
+     */
+    Data: ByteView<string>
+    /**
+     * Number of bytes in Data field
+     */
+    filesize: uint64
```
Gozala, line 288: Can symlinks have mtime / mode ? They also [seem to have blocksizes](https://explore.ipld.io/#/explore/QmUGzmqt6pcUnghcu68UtkaR8i2r9BGuKQFUNbR1yVTsH8/dave.txt) 

Are they supposed to or is that artifact of normalization no js-ipfs-unixfs side ?

image

unixfs.ts
```diff
@@ -0,0 +1,488 @@
...
+ */
+export type DirectoryLayout =
+  | FlatDirectoryLayout
+  | AdvancedDirectoryLayout
+
+
+/**
+ * @TODO
+ */
+export type MetadataNode = never
```
Gozala, line 269: Can anyone please help clarify what these nodes should contain and what supposed to link to them ? Spec seems to talk about [all the things that were considered and ultimately rejected](https://github.com/ipfs/specs/blob/master/UNIXFS.md#metadata-1), but it seems to fails to mention what the consensus is.

Is `Metadata` deprecated legacy in favor of `mtime`, `mode` or ?

unixfs.ts
rvagg: the reason for this is to do with the ambiguity of present and not-present in the translation between PB and the data model—you just can't distinguish; so the spec makes it just one way, always present, so there is only one data model form for any given PB form, and not multiple data model forms that can be turned into the same PB bytes.

optionality here doesn't save PB bytes btw, it's the same either way, this is just about 1:1 between in-memory representation and over-the-wire bytes.

rvagg:
As an FYI, @Jorropo identified an interesting quirk in go-ipfs's (more likely go-unixfs's) current leaf building which will leave an empty `Name` where it's not necessary: https://github.com/ipfs/go-ipfs/issues/8691

It's more of a hash consistency concern than anything, we have plenty of data in the wild with blank `Name` and we actively test all the dag-pb variations identified in the dag-pb spec (see the fixtures starting with "dagpb" @ https://github.com/ipld/codec-fixtures/tree/master/fixtures). It would be nice to bias toward _not_ doing this where possible, so not locking in such an expectation in any of our specs.

rvagg (approved):
This seems mostly fine to me, I'm not a big fan of locking in `Tsize` or `Name` in the dag-pb forms, but I can see what you're trying to do is build types for cases where they really should exist, leaving the base `PBNode` and `PBLink` matching the dag-pb spec.
I'm not as well versed on the UnixFS portions, although I'd be willing to bet there's some surprises wrt optionality in particular if you went digging through existing data and historical code! But sometimes the best way to do that is lock down a spec and see who/what yelps (that's certainly been my experience!).

unixfs.ts
```diff
@@ -0,0 +1,643 @@
...
+ * 0x55 multicodec) which are on a way to surpass this representation.
+ *
+ * Please note that in reality there is only one `file` node with many optional
+ * fields, however different combination of fields corresponds to a different
+ * semntaics and we represent each by a different type.
+ * 
+ * Also note that some file nodes may also have `mode` and `mtime` fields,
+ * which we represent via `SimpleFileLayout` type however that is not completely
+ * accurate, e.g because one could take two `SimpleFileLayout`s and represent
+ * their concatination via `AdvancedDirectoryLayout` by linking to them. In such
```
Gozala, line 61:
```suggestion
 * their concatination via `AdvancedFileLayout` by linking to them. In such
```

unixfs.ts
```diff
@@ -0,0 +1,643 @@
...
+
+
+/**
+ * Less loosely defined PB Link which requires TSize and does not
+ * require `Name`.
+ */
+export interface AnonymousLink<Data> extends PBLink<Data> {
+  Hash: CID<Data>
+  /**
+   * UnixFS links MUST specify size of the linked block.
```
Gozala, line 519:
```suggestion
   * UnixFS links SHOULD specify size of the linked subgraph
   * (cumulative size of linked block and the blocks under it)
```

unixfs.ts
```diff
@@ -0,0 +1,643 @@
...
+export interface NamedLink<Data> extends PBLink<Data> {
+  Hash: CID<Data>
+  /**
+   * Named links (e.g. directory links) MUST specify link name. In case of
+   * directories those are interpreted as entry names.
+   */
+  Name: string
+
+  /**
+   * UnixFS links MUST specify size of the linked block.
```
Gozala, line 544:
```suggestion
   * UnixFS links SHOULD specify size of the linked subgraph
   * (cumulative size of linked block and the blocks under it)
```

Gozala:
Few things I'd discovered in the realm of unspecified dags

```
ipfs ls /ipfs/Qme1Cyu7ujqn3dRkRGmeTLpHJgbGHFjKmud48XK5W8qA6h
Error: only murmur3 supported as hash function


ipfs dag get /ipfs/Qme1Cyu7ujqn3dRkRGmeTLpHJgbGHFjKmud48XK5W8qA6h 
{"Data":{"/":{"bytes":"CAU"}},"Links":[]}

# this is what went into that { Type: Data.DataType.HAMTShard }

ipfs ls /ipfs/QmS73FCYEH2NaSypVDE8EzQsjXoGTa8Wxq1bC22LfYnbiR                                                                ✱
Error: hamt size should be a power of two

# This is what went into that { Type: Data.DataType.HAMTShard, hashType: 0x22 }

ipfs ls /ipfs/Qma5kEnM5fEKTXrFC5zXYRy5QG3hcMWopoFS7ijhxx19qc
{"Data":{"/":{"bytes":"CAUoIjCAAg"}},"Links":[]}

# This is what went into that { Type: Data.DataType.HAMTShard, hashType: 0x22, fanout: 256 }
# looks like leaving out bitfield is fine but not other two fields
```

Gozala:
Here is another fun example, illustrating that filesize can't be trusted and probably should not be even encoded as it could be derived from Data + blocksizes.

```
echo '{ "Data": "CAISC2hlbGxvIHdvcmxkGAU=", "Links": [] }' | ipfs object put --datafieldenc base6
ipfs files stat /ipfs/QmPivGwShRuUC9gB17RakWKhixsTU5PMvxBZjXfXxo4KcU                                                        ✱
QmPivGwShRuUC9gB17RakWKhixsTU5PMvxBZjXfXxo4KcU
Size: 5
CumulativeSize: 19
ChildBlocks: 0
Type: file
ipfs cat /ipfs/QmPivGwShRuUC9gB17RakWKhixsTU5PMvxBZjXfXxo4KcU | wc -c
11
ipfs cat /ipfs/QmPivGwShRuUC9gB17RakWKhixsTU5PMvxBZjXfXxo4KcU                                                               ✱
hello world
```

rvagg:
Maybe we should make this a TODO in GitHub rather than hiding it in code: https://github.com/ipfs/go-unixfs/blob/5e0e095d575b5b5fbf2fea386ea11ab737e4395e/unixfs.go#L260-L265

PR #269, opened 2022-03-04
Bitswap 1.2.0

aschmahmann:
The current Bitswap spec is a bit unwieldly and out of date and generally seemed in need of some cleaning up. The spec here was a rewrite but some notable changes include:
- Added `/ipfs/bitswap/1.2.0`
- Clarified the protocol identifiers used to identify the protocol versions
- Clarified the specified protocol message format per-protocol version
- Specified the maximum message size (4MiB)
- Specified the minimum compliant maximum block size (2MiB)
- Specified that messages are varint-delimited protobufs

### DIFF
--- a/BITSWAP.md
+++ b/BITSWAP.md
@@ -60,23 +60,23 @@ A single Bitswap message may contain any of the following content:
 
 #### Wire Format
 
-The wire format for Bitswap is simply a stream of Bitswap messages. The following protobuf describes the form of these messages.
+The wire format for Bitswap is simply a stream of Bitswap messages. The following protobuf describes the form of these messages. Note: all protobufs are described using proto3 syntax.
 
 ```
 message Message {
   message Wantlist {
     message Entry {
-      optional bytes block = 1; // the block key, i.e. a CIDv0
-      optional int32 priority = 2; // the priority (normalized). default to 1
-      optional bool cancel = 3;  // whether this revokes an entry
+      bytes block = 1; // the block key, i.e. a CIDv0
+      int32 priority = 2; // the priority (normalized). default to 1
+      bool cancel = 3;  // whether this revokes an entry
     }
 
     repeated Entry entries = 1; // a list of wantlist entries
-    optional bool full = 2;     // whether this is the full wantlist. default to false
+    bool full = 2;     // whether this is the full wantlist. default to false
   }
 
   Wantlist wantlist = 1;
-  optional repeated bytes blocks = 2;
+  repeated bytes blocks = 2;
 ```
 
 ### Protocol Format
--- a/BITSWAP.md
+++ b/BITSWAP.md
@@ -1,12 +1,12 @@
 # ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) Bitswap
 
-**Authors(s)**:
+**Author(s)**:
 - Adin Schmahmann
 - David Dias
 - Jeromy Johnson
 - Juan Benet
 
-**Maintainers(s)**:
+**Maintainer(s)**:
 
 * * *
 
--- a/BITSWAP.md
+++ b/BITSWAP.md
@@ -43,7 +43,7 @@ There are multiple Bitswap versions and more may evolve over time. We give brief
 
 ## Block Sizes
 
-Bitswap implementations must support sending and receiving individual blocks of sizes less than or equal to 2MiB
+Bitswap implementations must support sending and receiving individual blocks of sizes less than or equal to 2MiB. Handling blocks larger than 2MiB is not recommended so as to keep compatibility with implementations which only support up to 2MiB.
 
 ## Bitswap 1.0.0
 
--- a/BITSWAP.md
+++ b/BITSWAP.md
@@ -27,7 +27,7 @@ Bitswap is a data exchange protocol for sending and receiving content addressed 
 
 # Introduction
 
-Bitswap is a message-based protocol, as opposed to response-reply. All messages contain wantlists, and/or blocks. Upon receiving a wantlist, a Bitswap server should eventually send wanted blocks if it has them. Upon receiving blocks, the client should send a `Cancel` notification to peers that have asked for the data, signifying that the client no longer wants the block.
+Bitswap is a message-based protocol, as opposed to request-response. All messages contain wantlists, and/or blocks. Upon receiving a wantlist, a Bitswap server should eventually process and respond to the requester with either information about the block or the block itself. Upon receiving blocks, the client should send a `Cancel` notification to peers that have asked for the data, signifying that the client no longer wants the block.
 
 Bitswap aims to be a simple protocol, so that implementations can balance aspects such as throughput, latency, fairness, memory usage, etc. for their specific requirements.
 
@@ -48,7 +48,8 @@ Given that a client C wants to fetch data from some server S:
 1. C sends a message to S for the blocks it wants, via a stream `s_want`
     1. C may either send a complete wantlist, or an update to an outstanding wantlist
     2. C may reuse this stream to send new wants
-2. S sends back blocks on a stream `s_receive`. S may reuse this stream to send back subsequent responses 
+2. S sends back blocks on a stream `s_receive`. S may reuse this stream to send back subsequent responses.
+  1. S should respect the relative priority of wantlist requests from C, with wants that have higher `priority` values being responded to first.
 3. When C no longer needs a block it previously asked for, it should send a `Cancel` message for that block to all peers from which it has not received a response about that block
 
 ### Bitswap Message
@@ -89,7 +90,7 @@ All protocol messages must be less than or equal to 4MiB in size
 
 ### Block Sizes
 
-Bitswap implementations must support sending and receiving individual blocks of sizes up to 2MiB
+Bitswap implementations must support sending and receiving individual blocks of sizes less than or equal to 2MiB
 
 ## Bitswap 1.1.0
 
@@ -105,13 +106,6 @@ It is otherwise identical to 1.0.0
 
 ```
 message Message {
-
-  message Wantlist {
-    enum WantType {
-      Block = 0;
-      Have = 1;
-    }
-
     message Entry {
 			bytes block = 1; // CID of the block
 			int32 priority = 2;	// the priority (normalized). default to 1
@@ -151,13 +145,13 @@ Given that a client C wants to fetch data from some server S:
     1. If C sends S a Have request for data S has (and is willing to give to C) it should respond with a Have, although it may instead respond with the block itself (e.g. if the block is very small)
     2. If C sends S a Have request for data S does not have (or has but is not willing to give to C) and C has requested for DontHave responses then S should respond with DontHave
     3. S may choose to include the number of bytes that are pending to be sent to C in the response message
+    4. S should respect the relative priority of wantlist requests from C, with wants that have higher `priority` values being responded to first.
 3. When C no longer needs a block it previously asked for it should send a Cancel message for that request to any peers that have not already responded about that particular block. It should particularly send Cancel messages for Block requests (as opposed to Have requests) that have not yet been answered.
 
 ### Wire Format
 
 ```
 message Message {
-
   message Wantlist {
     enum WantType {
       Block = 0;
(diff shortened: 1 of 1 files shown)


BITSWAP.md
```diff
@@ -11,56 +12,51 @@
...
+Bitswap is a relatively simple protocol, where complexity lives in the particular implementations to optimize for their particular balance of aspects such as throughput, latency, fairness, memory usage, etc.
 
-While Bitswap is a relatively simple protocol, a time- and memory- performant implementation requires that many details be carefully thought out. We aim to document these details here so that future implementers may build upon the knowledge gathered over the several iterations of Bitswap.
+# Bitswap Protocol Versions
 
-# Subsystems
+There are multiple Bitswap versions and more may evolve over time. We give brief overviews as to the changes behind each protocol version.
 
-There are two primary flows that Bitswap manages: requesting blocks from and serving blocks to peers. Block requests are primarily mediated by the want-manager, which tells our peers whenever we want a new block. Serving blocks is primarily handled by the decision engine, which decides how resources should be allocated among our peers.
+- `/ipfs/bitswap/1.0.0` - Initial version
```
aschmahmann, line 38: I've also seen `/ipfs/bitswap` and `/bitswap/1.0.0` around digging through the GitHub history. I'm not sure if these are the same as `/ipfs/bitswap/1.0.0` or different earlier versions.

Perhaps someone with sufficient history can clarify, or I can do more GitHub archaeology to learn more.

BITSWAP.md
```diff
@@ -79,122 +75,126 @@ message Message {
...
 
-## Message Queue
+All protocol messages must be less than or equal to 4MiB in size
 
-Each active peer has an associated message queue. The message queue holds the next message to be sent to that peer. The message queues receive updates from two other subsystems:
+### Block Sizes
 
-1.  Wantlist manager: When a `CID` is added or removed from our wantlist, we must update our peers – these wantlist updates are sent to all relevant peers’ message queues.
-2.  Decision engine: When we have a block that a peer wants and the decision engine decides to send the block, we propagate the block to that peer’s message queue.
+Bitswap implementations must support sending and receiving individual blocks of sizes up to 2MiB
```
aschmahmann, line 92: IIUC technically today's Bitswap implementations can support somewhere in the slightly less than 4MiB range of bytes for a given block.

Saying messages must be <=4MiB and Bitswap implementations must support blocks <=2MiB in size is technically correct. However, I wonder if we should be giving any more guidance here about what implementations should do with blocks in the 2-4MiB range.

BITSWAP.md
```diff
@@ -79,122 +75,126 @@ message Message {
...
-  - protocol sender opens streams to partners to send messages
-  - separate out a decision engine that selects which blocks to send to which partners, and at what time. (this is a bit tricky, but it’s super easy to make as a whole if the engine is separated out)
+  message Block {
+    bytes prefix = 1; // CID prefix (all of the CID components except for the digest of the multihash)
+    bytes data = 2;
+  }
 
-Sender:
+  Wantlist wantlist = 1;
+  repeated Block payload = 3;
```
aschmahmann, line 131: While it appears to be supported in go-bitswap I removed `repeated bytes blocks = 2;` from versions of the protocol >1.0.0 as that field is deprecated and superseded by another field (i.e. `payload`). Given that we use libp2p and multistream to negotiate protocol versions there seems to be no reason why a `/ipfs/bitswap/1.1.0` node should ever receive a message with the `blocks` field since a node needing the `blocks` field should have negotiated for `/ipfs/bitswap/1.0.0`.

However, given that this behavior is likely implemented across the various implementations today I'm happy to put it back in the spec if people think it should be there.

BITSWAP.md
```diff
@@ -79,122 +75,126 @@ message Message {
...
+      bool cancel = 3; // whether this revokes an entry
+      WantType wantType = 4; // Note: defaults to enum 0, ie Block
+      bool sendDontHave = 5; // Note: defaults to false
+		}
+
+    repeated Entry entries = 1;	// a list of wantlist entries
+    bool full = 2; // whether this is the full wantlist. default to false
+  }
+  message Block {
+    bytes prefix = 1; // CID prefix (all of the CID components except for the digest of the multihash)
```
aschmahmann, line 179: The previous comment here seemed to indicate that this had to be a CIDv1 when it could be a CIDv0 as well.

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
 # ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) Bitswap
 
-**Authors(s)**:
+**Author(s)**:
+- Adin Schmahmann
```
aschmahmann, line 4: Just guessed this was alphabetical order by first name 🤷.

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
...
 - David Dias
 - Jeromy Johnson
 - Juan Benet
 
-**Maintainers(s)**:
+**Maintainer(s)**:
```
aschmahmann, line 9: I'm not sure what this is supposed to be used for/if it really makes sense to keep around. Perhaps this is an interest group that I should be adding more people in the IPFS community who are interested in the Bitswap specs and improving them or being particularly involved in protocol upgrade proposals.

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
...
 
-## Message Queue
+All protocol messages must be less than or equal to 4MiB in size
 
-Each active peer has an associated message queue. The message queue holds the next message to be sent to that peer. The message queues receive updates from two other subsystems:
+### Block Sizes
 
-1.  Wantlist manager: When a `CID` is added or removed from our wantlist, we must update our peers – these wantlist updates are sent to all relevant peers’ message queues.
-2.  Decision engine: When we have a block that a peer wants and the decision engine decides to send the block, we propagate the block to that peer’s message queue.
+Bitswap implementations must support sending and receiving individual blocks of sizes up to 2MiB
```
Jorropo, line 92: I have already seen other IPLD implementers saying that the limit is actually 2MiB-1byte, (in order to fit in a 21bits uint ?).

However I couldn't find any data to back this up, current implementations accepts more anyway to account for the fact that very few chunkers are capable to precisely matching this, and might be slightly over that limit.

It should be clearer if that is `2MiB` inclusive or exclusive.

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
...
-When a message queue that already contains a Bitswap message receives another, the new message should be coalesced with the original to reduce the overhead of sending separate packets.
-
-## Bitswap Sessions
-
-Bitswap sessions are an attempt to optimize the block requests sent to other Bitswap clients. When requesting a graph of blocks from the network, we send a wantlist update containing the graph’s root block to all of our peers. Then, for each peer who sends the root block back, we add that peer to the graph’s *active set*. We then send all requests for other nodes in the graph only to the peers in the active set. The idea is that peers who have the root node of a graph are likely to have its children as well, while those who do not have the root are unlikely to have its children.
+```
+message Message {
 
-**TODO**: Everything below must either be updated/integrated above, or removed
+  message Wantlist {
```
aschmahmann, line 109: This seems to have some of the 1.2 fields. It should be fixed to correctly represent 1.1

BigLep:
@aschmahmann : can we merge this?  This seems like a quick thing to close out before you head out.

aschmahmann:
@BigLep yep, will get this cleaned up and ready for merge today

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
...
-  - `Block`: A binary blob.
-  - `Message`: A Bitswap message.
-  - `Entry`: A wantlist entry that may be included in a Message when adding/removing a particular `CID` from our wantlist. Contains:
-      - `CID` referring to a particular block.
-      - `Priority` relative priority with which the user wants `CID` (relevant only if `Cancel` is not true.
-      - `Cancel` is a boolean representing whether this `Entry` is meant to remove `CID` from our wantlist.
-  - `Ledger`: A record of the aggregate data exchanged between two peers. Each peer stores one `Ledger` for each of their peers.
+1. C opens a stream `s_want` to S and sends a message for the blocks it wants
+    1. C may either send a complete wantlist, or an update to an outstanding wantlist
+    2. C may reuse this stream to send new wants
```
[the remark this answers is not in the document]
aschmahmann: Maybe, I just want to make sure that people don't think they need to (or that they cannot) reuse the stream to send wants rather than opening a stream, sending a message, and the closing the stream each time.

BITSWAP.md
```diff
@@ -1,200 +1,200 @@
...
 
 ```
 message Message {
   message Wantlist {
     message Entry {
-      optional bytes block = 1; // the block key
-      optional int32 priority = 2; // the priority (normalized). default to 1
-      optional bool cancel = 3;  // whether this revokes an entry
+      bytes block = 1; // the block key, i.e. a CIDv0
+      int32 priority = 2; // the priority (normalized). default to 1
```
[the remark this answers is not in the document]
aschmahmann: Added some text in https://github.com/ipfs/specs/pull/269/commits/c9b1451cdfbfc937424edb520518ec0c2e17b1fe

BITSWAP.md
aschmahmann: Added more clarity in https://github.com/ipfs/specs/pull/269/commits/c9b1451cdfbfc937424edb520518ec0c2e17b1fe

BITSWAP.md
aschmahmann: fixed

guseggert (approved):

BITSWAP.md
```diff
@@ -1,202 +1,198 @@
...
-  - `Peer`: Another Bitswap instance that we are connected to.
-  - `Block`: A binary blob.
-  - `Message`: A Bitswap message.
-  - `Entry`: A wantlist entry that may be included in a Message when adding/removing a particular `CID` from our wantlist. Contains:
-      - `CID` referring to a particular block.
-      - `Priority` relative priority with which the user wants `CID` (relevant only if `Cancel` is not true).
-      - `Cancel` is a boolean representing whether this `Entry` is meant to remove `CID` from our wantlist.
-  - `Ledger`: A record of the aggregate data exchanged between two peers. Each peer stores one `Ledger` for each of their peers.
+1. C sends a message to S for the blocks it wants, via a stream `s_want`
+    1. C may either send a complete wantlist, or an update to an outstanding wantlist
```
aschmahmann, line 55: Not sure if there's a better way to do the formatting here. Using two spaces causes a linting error and four might be overkill. Do I need to manually use letters here?

guseggert: Four spaces sounds fine to me :man_shrugging:

BITSWAP.md
```diff
@@ -1,202 +1,198 @@
...
-There are two primary flows that Bitswap manages: requesting blocks from and serving blocks to peers. Block requests are primarily mediated by the want-manager, which tells our peers whenever we want a new block. Serving blocks is primarily handled by the decision engine, which decides how resources should be allocated among our peers.
+- `/ipfs/bitswap/1.0.0` - Initial version
+- `/ipfs/bitswap/1.1.0` - Support CIDv1
+- `/ipfs/bitswap/1.2.0` - Support Wantlist Have's and Have/DontHave responses
 
-The subsystems involved in these flows are detailed in the following subsections.
+## Block Sizes
 
-**TODO**: Update graphic
+Bitswap implementations must support sending and receiving individual blocks of sizes less than or equal to 2MiB
```
aschmahmann, line 46:
```suggestion
Bitswap implementations must support sending and receiving individual blocks of sizes less than or equal to 2MiB. Handling blocks larger than 2MiB is not recommended so as to keep compatibility with implementations which only support up to 2MiB.
```

@guseggert WDYT is adding this line overkill given that it's essentially implied by the "must"?

guseggert: I don't think it's necessarily implied by the "must", you can support 1 GiB blocks and still conform with the "must", so it's a useful note. Unless you change the wording ot be something like "must only support".  But I think it's good to leave a little wiggle room here, so I like this new phrasing.

aschmahmann: SGTM. Note 1GiB couldn't be supported, you can only go up to a bit under 4MiB even if you wanted to due to the Bitswap message size limit.

guseggert: Ha true, was just using an extreme example. SGTM too.

Issue #260, opened 2022-01-05
Bitswap auth


ianopolous:
I'd like to propose an optional extension to bitswap for requiring authorisation to retrieve a block. We've already implemented this and integrated it in Peergos. 

The proposal is to add optional auth byte[] to wants, blocks and block presence. Our updated protobuf is here - 
https://github.com/Peergos/go-bitswap-auth/blob/master/message/pb/message.proto

The repo where we've implemented this is: https://github.com/peergos/go-bitswap-auth
Trying to directly upstream our code there probably doesn't make sense because of the changes we've made to some external interfaces, but if it does then we can consider relicencing it from AGPL in that situation. 

We use this in Peergos (via [ipfs-nucleus](https://github.com/peergos/ipfs-nucleus/)) to implement a post-quantum capability based auth scheme using S3 V4 signatures. You can read more about our design here: https://github.com/Peergos/Peergos/issues/862

PR #270, opened 2022-03-07
Bitswap 1.3.0 - Tokens (and auth) support

aschmahmann:
This is a spec proposal continuing from #260 and other issues related to support for authentication, payments, etc. in Bitswap. This builds off of #269 which helps clean up the Bitswap spec a bit and bring it up to date.

The main thrust of the proposal is support for adding a set of tokens to requests and responses to allow metadata. These tokens can convey various application specific requirements or interactions 

A bonus component added to the proposal is adding a `BlockTooBig` type of `BlockPresence` so that proper error messages can be communicated for that use case rather than either causing a stream reset or something more confusing like returning a Have, DontHave, or not responding at all.

This kind of change has been needed for a while, and while this may not resolve everything it seems like it will help us to have a concrete proposal to critique and improve.

cc some folks who have been asking about this for a while: @Stebalien @ianopolous @csuwildcat @obo20 @carsonfarmer @bmann

---

I'll give some background and pre-empt some questions here. If any of this is interesting enough to put in the spec we can do so there.

Unfortunately, GH is pretty bad at threaded discussions not tied to a particular piece of code so if you have a comment about something not explicitly in the code just grab some empty line in the PR and start a thread there.

## Background

In Bitswap versions 1.2.0 and earlier it is not feasible for a server to decide on whether a peer is authorized to receive a particular block of data other than by discriminating on the peer itself. Consider if Alice wants to request a block B from Bob he will serve it to her if she is on his allowlist, and otherwise will not. If Alice wants to gain authorization to download B she can use some external protocol `/authproto` to negotiate her authorization for B with Bob.

This precludes a number of use cases including:
- Multiple auth mechanisms per peer
    - Multitenancy: If Alice was a multitenancy client that was trying to download data from Bob on behalf of Charlie and Dan she doesn't have a good way to know which of Charlie or Dan are supposed to have access to B which means she could be handing over B to Dan even though he shouldn't have been given it in the first place. This might be very useful for semi-trusted intermediaries that are able to download data via Bitswap on behalf of users in more restrictive environments.
    - Per Retrieval Payments: Similar issues arise in the case of Alice and Bob trying to negotiate payments for the retrieval of chunks of two DAGs D1 and D2 both of which contain the block B. When Bob sends Alice B if payments occur after receipt then Alice doesn't know if Bob is expecting to be paid the D1 or the D2 payment rate for B. While probably Bob should just charge whichever is lesser, if the payment scheme is more complicated this might be non-obvious or generally determinable
- Using auth as a fallback: i.e. trying to fetch data without auth before falling back to using auth
    - If Bob has B and Alice asks for it but is not yet authorized there is no way for Bob to communicate the difference between "I don't have the data" and "I have the data but you need to authorization" if he so chooses. This is particularly problematic in the case where the authorization is open (e.g. anyone who wants to pay for the data, anyone who solves a CAPTCHA, etc.) since then the options are to not believe any DONT_HAVEs and waste resources negotiating authorization or for Alice not be able to download data from any peer she didn't know in advance she needed to authenticate with
    
## Alternatives/FAQs

- Why not remove the `AuthRequired` token and instead put auth requirements in the content routing system?
    - Requires going to the content routing system frequently (potentially for every block) to figure out if authorization is required which defeats the important Bitswap optimization of skipping content routing by asking connected peers if they can send us blocks
    - Requires content routing systems to support the auth mechanisms which is a bigger ask than if it could be done at the exchange layer
- Why support arbitrary token types?
    -  Proposals like #260 that do not differentiate different token types make it difficult for the application layer to know what to do with the tokens it received. It may have to do some testing on the tokens to figure out what they are for. It may also lead to ambiguity when a token could be used for two separate authentication schemes
    - IMO a generally useful piece of multiformats is that for the low cost of a few bytes you can figure out what your data is, that seems like something to take advantage of here
- Why not dictate exactly what tokens can/should be used for?
    - Mostly because application developers will do whatever they want here anyway unless it explicitly breaks the protocol. We should give guidance, but any MUSTs here are likely pipe-dreams even if we wanted them
- Why is there a level of indirection around tokens and token indices?
    - Basically to save bytes since we're not using something like gzip on the messages. It is expected that certain scenarios will have many blocks reuse the same tokens and we can save a lot of bytes there, while not being too costly in the one-token-per-block case

### DIFF
--- a/BITSWAP.md
+++ b/BITSWAP.md
@@ -38,6 +38,7 @@ There are multiple Bitswap versions and more may evolve over time. We give brief
 - `/ipfs/bitswap/1.0.0` - Initial version
 - `/ipfs/bitswap/1.1.0` - Support CIDv1
 - `/ipfs/bitswap/1.2.0` - Support Wantlist Have's and Have/DontHave responses
+- `/ipfs/bitswap/1.3.0` - Support adding tokens in Bitswap requests/responses and BlockTooBig error message
 
 ## Bitswap 1.0.0
 
@@ -193,6 +194,94 @@ message Message {
   repeated Block payload = 3;
   repeated BlockPresence blockPresences = 4;
   int32 pendingBytes = 5;
+}
+```
+
+## Bitswap 1.3.0
+
+Bitswap 1.3.0 extends the Bitswap 1.2.0 protocol with the following changes:
+1. Having a list of tokens that may be sent with the message and referenced within the message
+2. Allowing each entry in the wantlist to contain a set of tokens
+3. Allowing responses (both BlockPresences and Blocks) to contain a set of tokens
+4. Adding an `AuthRequired` BlockPresence
+5. Adding a `BlockTooBig` BlockPresence
+
+### Interaction Pattern
+
+Given that a client C wants to fetch data from some server S:
+
+1. C opens a stream `s_want` to S and sends a message for the blocks it wants
+    1. C may either send a complete wantlist, or an update to an outstanding wantlist
+    2. C may reuse this stream to send new wants
+    3. For each of the items in the wantlist C may ask if S has the block (i.e. a Have request) or for S to send the block (i.e. a block request). C may also ask S to send back a DontHave message in the event it doesn't have the block
+    4. For each of the items in the wantlist C may append any `tokens` they want. Recommended tokens include those that would convince S to give C the blocks they want.
+2. S responds back on a stream `s_receive`. S may reuse this stream to send back subsequent responses
+    1. If C sends S a Have request for data S has (and is willing to give to C) it should respond with a Have, although it may instead respond with the block itself (e.g. if the block is very small)
+        1. For each of the items that S sends back Blocks or BlockPresences for they may append any `tokens` they want
+    2. If C sends S a Have request for data S has but is not currently willing to give to C, S may respond with an `AuthRequired` BlockPresence
+      1. For each of the items that S sends back Blocks or BlockPresences for they may append any `tokens` they want. Recommended tokens include those that would inform C how they could convince S to give them access to the blocks they want.
+    2. If C sends S a Have request for data S does not have (or has but is not willing to tell C it has) and C has requested for DontHave responses then S should respond with DontHave
+    3. S may choose to include the number of bytes that are pending to be sent to C in the response message
+    4. If C asked for a block that S has but it is bigger than the maximum block size it should return `BlockTooBig`
+3. When C no longer needs a block it previously asked for it should send a Cancel message for that request to any peers that have not already responded about that particular block. It should particularly send Cancel messages for Block requests (as opposed to Have requests) that have not yet been answered.
+
+### Tokens
+
+The major change in this protocol version is the introduction of the ability to pass tokens along with requests and responses. A token is defined as `<multicode><data>` where the multicode is an identifier in the multicodec table, and the data is token-specific data associated with that code.
+
+Users who require additional codes for their new token formats should do one of:
+- Register their code in the table
+- For non-deployed testing purposes only - use a code in the application reserved range of the code table
+   - Note: if codes in the application reserved range will not be reservable in the code table which means that the code may conflict with another one used in the ecosystem which could cause application problems on collision. It is high recommended to not use these codes outside of testing or early development.
+
+To save space within the message the list of tokens used within the message are declared within the top level message and all other references to tokens are instead to the indices within the token list in the top level message.
+
+### Wire Format
+
+```
+message Message {
+
+  message Wantlist {
+    enum WantType {
+      Block = 0;
+      Have = 1;
+    }
+
+    message Entry {
+      bytes block = 1; // CID of the block
+      int32 priority = 2;	// the priority (normalized). default to 1
+      bool cancel = 3; // whether this revokes an entry
+      WantType wantType = 4; // Note: defaults to enum 0, ie Block
+      bool sendDontHave = 5; // Note: defaults to false
+      repeated tokens int32 = 6; // the indices of the tokens in the token list
+		}
+
+    repeated Entry entries = 1;	// a list of wantlist entries
+    bool full = 2; // whether this is the full wantlist. default to false
+  }
+  message Block {
+    bytes prefix = 1; // CID prefix (all of the CID components except for the digest of the multihash)
+    bytes data = 2;
+    repeated tokens int32 = 3; // the indices of the tokens in the token list
+  }
+
+  enum BlockPresenceType {
+    Have = 0;
+    DontHave = 1;
+    AuthRequired = 2;
+    BlockTooBig = 3;
+  }
+  message BlockPresence {
+    bytes cid = 1;
+    BlockPresenceType type = 2;
+    repeated tokens int32 = 3; // the indices of the tokens in the token list
+  }
+
+  Wantlist wantlist = 1;
+  repeated Block payload = 3;
+  repeated BlockPresence blockPresences = 4;
+  int32 pendingBytes = 5;
+  repeated bytes tokens = 6; // Each token is of the form <multicode><token-data> where the multicode identifies what the data is for
 }
 ```
 


ianopolous:
Hi @aschmahmann Thanks for making a start on this! 

I would argue that a byte[] is the most general you can get. In fact, for our usage in Peergos these byte[] are cbor. So we get the benefit of types, upgradability etc. but at the application layer and without the centralisation point of multicodecs. And this also doesn't force everyone to pay the overhead of the multicodec. This means applications can innovate faster and without requiring anyone's permission. 

The other change I would suggest is to allow per cid auth tokens inline, as we do. Otherwise you're forcing someone like us to pay the extra overhead of the indirection (No auth tokens are ever shared between blocks for us, and arguably that would be a vulnerability (e.g. S3 doesn't allow this)). Probably there is a use case for both, especially if you include non capability-based schemes. It should be easy to support both inline tokens and indirect references. 

Do you have a concrete use case in mind for allowing multiple auth tokens for a single cid in the same request? Our usage in Peergos is maximally fine grained and doesn't need this. For example, Amazon S3 doesn't need this and they clearly support almost all use cases in the world right now.

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
 }
 ```
 
+## Bitswap 1.3.0
+
+Bitswap 1.3.0 extends the Bitswap 1.2.0 protocol with the following changes:
+1. Having a list of tokens that may be sent with the message and referenced within the message
+2. Allowing each entry in the wantlist to contain a set of tokens
+3. Allowing responses (both BlockPresences and Blocks) to contain a set of tokens
+4. Adding an `AuthRequired` BlockPresence
```
dignifiedquire, line 206: given that tokens are pretty generic, why not make this error more generic?

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
...
+### Tokens
+
+The major change in this protocol version is the introduction of the ability to pass tokens along with requests and responses. A token is defined as `<multicode><data>` where the multicode is an identifier in the multicodec table, and the data is token-specific data associated with that code.
+
+Users who require additional codes for their new token formats should do one of:
+- Register their code in the table
+- For non-deployed testing purposes only - use a code in the application reserved range of the code table
+   - Note: if codes in the application reserved range will not be reservable in the code table which means that the code may conflict with another one used in the ecosystem which could cause application problems on collision. It is high recommended to not use these codes outside of testing or early development.
+
+To save space within the message the list of tokens used within the message are declared within the top level message and all other references to tokens are instead to the indices within the token list in the top level message.
```
aschmahmann, line 237: From @ianopolous:

> The other change I would suggest is to allow per cid auth tokens inline, as we do. Otherwise you're forcing someone like us to pay the extra overhead of the indirection (No auth tokens are ever shared between blocks for us, and arguably that would be a vulnerability (e.g. S3 doesn't allow this)). Probably there is a use case for both, especially if you include non capability-based schemes. It should be easy to support both inline tokens and indirect references.

We could add another field at each entry `repeated tokens bytes` in the Entry, Block and BlockPresence messages. It adds more complexity to save some bytes, I wonder if it's really that helpful. Another byte saving approach which would might save more bytes would be to store another list of the token-codes used in the token list. e.g. if your token code is > 3 bytes and you put it into the token list and then put your per-block tokens into the token list that would probably save you more bytes than this. Also, it seems like if we wanted to support strings/arbitrary bytes as custom "codes" then the above might save a whole bunch of data if we defined the codes as `<code-number>[optional varint + data that goes with the code]<data>` and stored the first two pieces in the second token list.

Both of these are about trying to shave a few bytes in exchange for some complexity. The initial proposal tried to cut down on the complexity unless the duplicated bytes were really excessive, finding which (if any) other complexity we want to add here though seems like a reasonable thing to discuss.

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
 }
 ```
 
+## Bitswap 1.3.0
+
+Bitswap 1.3.0 extends the Bitswap 1.2.0 protocol with the following changes:
+1. Having a list of tokens that may be sent with the message and referenced within the message
+2. Allowing each entry in the wantlist to contain a set of tokens
```
aschmahmann, line 204: From @ianopolous:

> Do you have a concrete use case in mind for allowing multiple auth tokens for a single cid in the same request? Our usage in Peergos is maximally fine grained and doesn't need this. For example, Amazon S3 doesn't need this and they clearly support almost all use cases in the world right now.

A couple things come to mind:
- Requiring two tokens from different systems: e.g. a Peergos token for authentication and some token related to paying for retrieval of that data
- "Suggested" ACLs: The server wants to tell the client which token caused me to give them the block (e.g. for multitenancy auth or payment purposes), but I also want to tell the client some general ACL I'd like them to respect

Maybe it's overkill, but we're making it easier for people to do auth (e.g. they could've just used peerID auth and some offline channel to coordinate authentication and payments) and we should probably try to make systems composable in-band. Thoughts?

aschmahmann (commented):
Thanks for the thoughts so far, I gave some feedback but definitely room for improvement. I'm trying to use threads to make the conversation a bit more manageable, we'll see how it goes 😄.

BITSWAP.md
dignifiedquire: Maybe just `TokenError`?

BITSWAP.md
ianopolous: Logically the simplest proposal is to have an auth string per cid, but you rightly point out that that has a large overhead when an auth string is repeated. If we're bumping the bitswap version number then maybe it's worth considering just gzipping the message?

BITSWAP.md
ianopolous: Separating the auth token from a payment token is an interesting idea. Has anyone done anything like this (is your angle composability with filecoin?)? I would normally imagine you would just pay for access to the content. I.e. pay to get an auth token/key. But I'm just hypothesising here. I guess what you're saying is different because you're not paying for access, you're paying for delivery, which should be independent. 

Note that in our multi-tenant case, the client already knows which token was authed to get a block without needing this.

dignifiedquire: Use cases I have talked to folks about: 
- ACLs without payments
- ACLs with payments
- request limiter, e.g. this token allows you to retrieve 100 times

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
...
+  enum BlockPresenceType {
+    Have = 0;
+    DontHave = 1;
+    AuthRequired = 2;
+    BlockTooBig = 3;
+  }
+  message BlockPresence {
+    bytes cid = 1;
+    BlockPresenceType type = 2;
+    repeated tokens int32 = 3; // the indices of the tokens in the token list
```
ianopolous, line 277: Are there any implications on the extensibility of the protobuf with a repeated field? My guess is the field numbers make this still extensible?

BITSWAP.md
ianopolous: To be clear, one absolutely critical property of any usage of this is that auth tokens MUST be tied to the sender's nodeId and this must be verified before releasing a block. Otherwise you are trivially vulnerable to replay attacks because you broadcast the tokens to the DHT. I think it would be worthwhile to flesh out a few of these use cases to make sure it is possible to do securely. E.g. a naive approach of using this to do an ACL by having tokens that the recipient looks up in a db to check that a token is allowed for a cid is insecure. To make it secure the db would also need to store the nodeId's allowed to use a given token, but then it wouldn't need this extension at all or any auth tokens, just use the nodeid of the requestor in normal bitswap.

ianopolous:
One other nice to have is to maintain backwards compatibility with our prior art coming up with a secure capability based auth scheme for Peergos and the associated bitswap extension. We have just extended the protobuf for bitswap 1.2, and this is making a new protocol id 1.3. However at the moment all versions of bitswap share the same protobuf. This is a nice simplification to not have to maintain code around N different versions of the protobuf. I think the simplest thing is just to avoid the field indices we've used in https://github.com/ipfs/specs/issues/260 ? Or even better, extend our protobuf which only adds optional auth strings. This means users can choose between the indirect list of auth tokens, or an inline one.

BITSWAP.md
ianopolous: The more I think about trying to use this to do ACLs (as opposed to capabilities) the more I think it is impossible to do securely because of the broadcast nature of bitswap, except in the trivial case where the ACL only depends on the nodeId. In that case you don't need this PR at all and you should just use a custom allow function as suggested [here](https://github.com/ipfs/go-ipfs/issues/7871) and as we've implemented [here](https://github.com/Peergos/go-bitswap-auth/blob/master/auth/blockstore.go#L42).

Fundamentally, it comes down to ACLs depending on identity and the only identity between ipfs nodes is the node id. So the ACL will only be able to depend on the node id, in which case why not just use the node id which is already authenticated and unfakeable without this PR. 

There is one way around this though, and that is single-use (not N>1) tokens. If a token is enforced to be only used once, then clearly a replay attack can't occur. That sounds like a massive footgun to me and also a scalabilty and centralisation problem as well.

ianopolous: Scratch my last statement about single use tokens, because whilst true, there is also a race condition between the originator and anyone else in the DHT they send the token to. So yes that is a footgun.

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
...
+      repeated tokens int32 = 6; // the indices of the tokens in the token list
+		}
+
+    repeated Entry entries = 1;	// a list of wantlist entries
+    bool full = 2; // whether this is the full wantlist. default to false
+  }
+  message Block {
+    bytes prefix = 1; // CID prefix (all of the CID components except for the digest of the multihash)
+    bytes data = 2;
+    repeated tokens int32 = 3; // the indices of the tokens in the token list
```
aschmahmann, line 265: from @ianopolous 
>  I think the simplest thing is just to avoid the field indices we've used

How important is this for you? There are two ways to handle this:
1. Skip field numbers (i.e. there is no field 3 defined)
2. We can use whatever numbers we want and since it's a different protocol version `/ipfs/bitswap/1.2.0` Block field 3 can be different from `/ipfs/bitswap/1.3.0` Block field 3.
    - This requires some custom code on your end at least until people upgrade to the new version. Not sure how annoying this is for you all though (e.g. if nodes upgrade quickly then it's probably not so bad since it only has to hang around for a little while)

dignifiedquire:
> I think gzipping would be nice, although it'd be nicer if it could be done at the libp2p layer

strong agree, we should push for adding compression in libp2p, not try to add it manually to protocols

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
...
+    2. If C sends S a Have request for data S does not have (or has but is not willing to tell C it has) and C has requested for DontHave responses then S should respond with DontHave
+    3. S may choose to include the number of bytes that are pending to be sent to C in the response message
+    4. If C asked for a block that S has but it is bigger than the maximum block size it should return `BlockTooBig`
+3. When C no longer needs a block it previously asked for it should send a Cancel message for that request to any peers that have not already responded about that particular block. It should particularly send Cancel messages for Block requests (as opposed to Have requests) that have not yet been answered.
+
+### Tokens
+
+The major change in this protocol version is the introduction of the ability to pass tokens along with requests and responses. A token is defined as `<multicode><data>` where the multicode is an identifier in the multicodec table, and the data is token-specific data associated with that code.
```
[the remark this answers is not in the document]
Stebalien: Basically, if you _don't_ use a multicodec prefix, you either need to:

1. Get everyone to agree on the same encoding scheme (cbor) and the same layout (e.g., some field to signal the "type"). 
2. Do a bunch of duck-typing. I.e., try to decode as cbor, try to fit into the expected type, then pass to the next token handler, etc.

The first case just won't happen. That's significantly _harder_ than getting everyone to prefix their tokens with a unique (per-protocol), fixed, prefix-free code.

In terms of the second case, yes, you can do this. But that leads to ambiguity that can be very difficult to reason about. In that case, I can't just register multiple token handlers, I need to think about how these different handlers might interact.

This is exactly multicodecs exist. It makes it really easy to remove ambiguity without having to get everyone to agree on a complex format, structure, etc.

BITSWAP.md
Stebalien: It can be done at the libp2p layer.

BITSWAP.md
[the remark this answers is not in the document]
aschmahmann: @dignifiedquire https://github.com/ipfs/specs/pull/270#issuecomment-1112463028

> so I am strongly in favor against requiring cbor.

IIUC no one wants to require cbor here. The discussion is around whether we should prefix with a multicode or use a duck-typing approach.

I was countering the argument in https://github.com/ipfs/specs/pull/270#discussion_r820902852

> cbor is exactly as capable as multicodecs at signalling types

by saying that it's only as capable if everyone agrees to it (as Steven mentioned above) and in that case a varint prefix seems preferable to requiring cbor inside of the protobufs 😄.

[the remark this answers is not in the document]
ianopolous: I see what you're both saying, but this ties a lot of high level application stuff, and a bunch of totally unrelated codecs that will never be used here, into the low level wire format. Thus making the bitswap protocol less of a thin waist protocol and increasingly a large waist with a bunch of constants from random protocols in it. I won't argue it anymore though, it's a stylistic choice basically. Both can function exactly the same performance and code wise (either agreeing on codec, or agreeing on a format and type signalling).

BITSWAP.md
ianopolous: Bitswap doesn't have a mechanism to only send wants to "the correct" peers. From a security perspective Wants are broadcast to the world. (If I want to receive an auth token for a given cid then I just need to pretend to the DHT that I have that cid and then the requests will come in with valid tokens). This is super important, and will trivially break security for most things people will try and do with it. 

It is possible to do securely *with* the broadcast assumption, which is what we've [figured out](https://peergos.org/posts/bats) in Peergos. A simplified analogy would be that if a secret S allows access to cid C. Then if node A is requesting C then the token sent with bitswap could be some form of signature using S of (A, C). Then the recipient can verify that the signature is valid, and the request came from A. Then there is no possibility for replay attacks because the signature is tied to the sender's node id. 

So I highly recommend speccing out some of these proposed usages and checking they are not totally broken security-wise.

BITSWAP.md
ianopolous: I'm just trying to think what our upgrade path with be like. If we skip the field numbers (1) then we can just trivially merge the protobuf and keep bitswap's single protobuf definition across all versions. That seems very desirable.

BITSWAP.md
aschmahmann: > Bitswap doesn't have a mechanism to only send wants to "the correct" peers

I don't understand this. Bitswap is a message sending protocol between two peers. 

> From a security perspective Wants are broadcast to the world.

This is only the case in go-bitswap, the decision as to which peers to send messages to is up to the implementation. Is there somewhere in the spec that suggests that wants are globally broadcast?

Security aside implementations may not want to broadcast to everyone simply as a way of cutting down on spamming and bandwidth usage. For example, even in the non-auth case I could see an implementation making the argument to only ask peers it's connected to over a LAN or where they were recommended by the content routing system.

> So I highly recommend speccing out some of these proposed usages and checking they are not totally broken security-wise.

A couple that are more suitable for open networks have been described (e.g. bats and payments), but doing simple bearer token style auth seems very easy as well. It just means that writing some CLI command like `fetch-cid <cid> <bearer-token>` is meaningless and instead you'd need to specify `fetch-cid <cid> <multiaddr> [bearer-token]` or `fetch-cid <cid> [tuple of mulitaddr and bearer token]`.

I'll try and reach out to some of the folks that have requested this functionality in the past for comments since AFAIK none of the people who have commented on this PR are planning on implementing support for something closer to a bearer-token style which wouldn't be appropriate for broadcast.

However, I'm confused as to what this has to do with the spec proposal. Even bats + payments require a spec change that looks like this though, right?

BITSWAP.md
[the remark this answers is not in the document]
Stebalien: I think there's still a pretty significant misunderstanding here.

> I see what you're both saying, but this ties a lot of high level application stuff, and a bunch of totally unrelated codecs that will never be used here, into the low level wire format.

Bitswap already uses CIDs and multihashes, which are built on multicodecs.

> Thus making the bitswap protocol less of a thin waist protocol and increasingly a large waist with a bunch of constants from random protocols in it.

Bitswap doesn't care about specific multicodecs, just that the token format is `<multicodec> ++ data`. It just cares that you prefix your token with a varint-encoded "thing".

> it's a stylistic choice basically. Both can function exactly the same performance and code wise (either agreeing on codec, or agreeing on a format and type signalling).

It's a pretty important design decision.

1. Just bytes. This is definitely the simplest, but it doesn't compose. I want to be able to have a single IPFS node integrate fetch blocks from multiple networks with multiple token types.
2. Typed
     1. CBOR + type signaling
           1. Mandates CBOR (complex, not everyone will want to use it).
           2. Needs a new "table of token types".
     2. Multicodec
          1. Just enough format to unambiguously distinguish between different types. Applications can use whatever they want for the actual token format (JWT, etc.).
          2. Re-uses the table of types we already have.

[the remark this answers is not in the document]
ianopolous: Agreed. Minor point - it would almost certainly be broken security to send a JWT as the token here for the reasons I've mentioned elsewhere here.

matheus23:
Hi :wave: I'm part of the fission team. This is an interesting proposal for us, since it'd technically make using [UCANs](https://ucan.xyz/) possible with bitswap directly.

Thoughts:
* Multiple tokens is a :+1: from me. Technically you could combine tokens into your own custom list format inside the byte array, but in combination with the multicodec proposal I think this way has some advantages, e.g. if tokens get forwarded into the DHT, different nodes may want to pick up different subsets of tokens, and we'd want them to agree on the format for "finding the subset".
* Multicodecs for tokens is also a :+1:. Multicodecs are the most minimal signalling protocol I know and they're used throughout the IPFS stack. I think the biggest value of this is forcing applications to use *some* signalling in the most minimal way. The worst possible situation would be allowing any bytes and someone ending up using a token where all byte sequences have a valid meaning.
* I'm undecided on whether to remove the "custom compression" and just go gzip on another level. I'm generally in favor of simplifying protocols, however in this case I think the complexity is really low and it's going to be a very common case of re-using tokens for multiple CIDs, if not the very default case. The repeated bytes will have to be represented at *some* point in memory, even if we're streaming them into gzip, the stream will need to repeat the token. Just wanted to note this, it may still be premature optimization, as I said, I'm not quite decided on that yet.

In any case, I can see this making UCAN use-cases possible with bitswap which is exciting for fission.

ianopolous:
@matheus23 Just a warning that using a UCAN in the token field here would also be totally insecure (in normal ipfs bitswap usage) as per the discussion above unless the audience of the ucan is the node id of the requesting ipfs node AND this is verified on the receiving end against the actual requestor, in addition to the usual ucan checks.

BITSWAP.md
ianopolous: > This is only the case in go-bitswap
To my knowledge, go-bitswap is the main implementation of bitswap, which is used by go-ipfs the main implementation of ipfs. Are you suggesting that existing ipfs apis wouldn't use this at all? If only new apis like you suggest will use this then that's much easier to avoid footguns.

matheus23:
> Just a warning that using a UCAN in the token field here would also be totally insecure (in normal ipfs bitswap usage) as per the discussion above unless the audience of the ucan is the node id of the requesting ipfs node AND this is verified on the receiving end against the actual requestor, in addition to the usual ucan checks.

Thanks for making that clear. We're aware of that. :+1: 
Sorry, I'm trying to read between the lines here... Would you expect my stance to change given that information?

aschmahmann:
Status Update:

Long overdue status update from the implementers sync (https://pl-strflt.notion.site/IPFS-Implementers-Sync-2022-05-12-b338b20567904923b3772c58f1c9186c) here is that this spec PR needs to be updated but should otherwise be reasonably solid. IIUC the remaining points were 1) smaller fixes to text in the spec 2) changing the protobuf numbers to skip the ones used by Peergos.

---

Ideally I'd like us to have two users of this work to help prove out the spec change's utility. Peergos is one use case that's also not blocked since they currently have an alternative mechanism riding on top of Bitswap 1.2. I was previously leaning on some of the folks working on payments to provide that second user story, but the folks working on [Boost](https://github.com/filecoin-project/boost) have delayed that work.

@matheus23 is this something Fission is interested in working on/using? From the Bitswap implementations I've looked at the server side changes to support tokens are fairly minimal, but the client side changes (as Ian has mentioned) require more work to fit the use case and consider how clients figure out which peers to send a given token to.

expede:
> Otherwise you are trivially vulnerable to replay attacks because you broadcast the tokens to the DHT

We've done some work with UCAN to avoid replays, too 👍 The most general solution is to enforce a nonce, so that replays are impossible (under normal assumptions: the protocol is adhered to, etc).

> unless the audience of the ucan is the node id of the requesting ipfs node AND this is verified on the receiving end against the actual requestor,

@ianopolous I think some of the roles are swapped, but yes that's how UCANs _must_ be used in all circumstances. The issuers and audience need to form a chain all the way back to the original node, including the node that is fulfilling the request.

For what it's worth, putting a UCAN right in a bitswap header probably doesn't make sense due to its size. As alluded to earlier, in point-to-point sessions, it's possible to do a handshake with a "blank"/self-targetted UCAN, and then send the triple `aud, ucan_cid, nonce` to shrink the size of the payload. We've also started playing around with the idea of a hash OTP, which seems to work well.

expede:
> cc some folks who have been asking about this for a while

Maybe to clarify a bit: the Fission team definitely welcomes this as a privacy measure, and I can see this potentially enabling more generic protocols for interacting with service providers, creating ad hoc private networks, etc. Given the often long lead times for changes in libp2p to get into projects like IPFS, working on this now seems like a good idea 💯 

That said, Fission doesn't have a burning need for these features today. I like the idea conceptually, but we probably won't go out of our way to use it until it's more baked in.

I also wonder if it would be possible to move this kind of a spec up a level from Bitswap? We're starting to explore alternatives to Bitswap, and having a generic authenticated session in libp2p is attractive.

ianopolous:
> > Otherwise you are trivially vulnerable to replay attacks because you broadcast the tokens to the DHT
> 
> We've done some work with UCAN to avoid replays, too +1 The most general solution is to enforce token uniqueness (via a nonce), so that replays are impossible (under normal assumptions: the protocol is adhered to, etc).

@expede Unfortunately, this is still broken by what I mentioned above - it reduces the replay to a (trivially exploitable) race condition. Nonces don't make it secure.

> > unless the audience of the ucan is the node id of the requesting ipfs node AND this is verified on the receiving end against the actual requestor,
> 
> @ianopolous I think some of the roles are swapped, but yes that's how UCANs _must_ be used in all circumstances. The issuers and audience need to form a chain all the way back to the original node, including the node that is fulfilling the request.

Based on my reading of the ucan spec the terms are correct. What I'm saying is in addition to a being a normal well formed ucan.

aschmahmann:
> That said, Fission doesn't have a burning need for these features today. I like the idea conceptually, but we probably won't go out of our way to use it until it's more baked in.

Could you clarify what you mean by "more baked in"? I want to double check that you're not waiting on go-ipfs (now [kubo](https://github.com/ipfs/go-ipfs/issues/8959#issuecomment-1156963447)) to implement something like `ipfs get bafyfoobar --token=<some-token>` since that's not something that's planned to be implemented for a combination of the UX and security reasons that have been mentioned above. However, in the case of other implementations and applications, supporting tokens like this makes things like downloading data authenticated with tokens much easier (https://github.com/ipfs/specs/pull/270#discussion_r857955469).

> I also wonder if it would be possible to move this kind of a spec up a level from Bitswap? We're starting to explore alternatives to Bitswap, and having a generic authenticated session in libp2p is attractive.

Having a generic authenticated session in libp2p seems like an interesting idea, although I don't think that solves our problems here since now if there are multiple possible tokens to use to download a given block of data (or multiple tokens that must be used together) life becomes much more complicated.

If you have a proposal in mind that seems like a great IPIP (https://github.com/ipfs/specs/pull/289), or a libp2p spec proposal. As you may have seen, the suggestion for a compression libp2p protocol that other protocols could layer on top of has come up which seems like it would need similar mechanisms, and is an area I think would be pretty cool for people to explore. However, it both seems like more work than this proposal and so far seems like it'd be less efficient as well so all things being equal I think this proposal as-is currently makes more sense.

expede:
> I want to double check that you're not waiting on go-ipfs (now https://github.com/ipfs/go-ipfs/issues/8959#issuecomment-1156963447) to implement something like ipfs get bafyfoobar --token=<some-token> since that's not something that's planned to be implemented for a combination of the UX and security reasons that have been mentioned above. However, in the case of other implementations and applications, supporting tokens like this makes things like downloading data authenticated with tokens much easier (https://github.com/ipfs/specs/pull/270#discussion_r857955469).

@aschmahmann yup, that's kind of the idea. I'm just downplaying my own comments as possibly not being relevant since if it doesn't land in Kubo or Iroh, we likely won't use this feature in the near-to-medium term, but I like the general concept. Boris was pinged on the issue, so we wanted to make sure that Fission weighed in!

ianopolous:
> @ianopolous you're going to have to walk me through that race condition. They can't connect as a node in the DID chain. This is an extremely limited number of peers, all of who are granted access by transitive closure. It ties them to the nodeId in the chain.

@expede We can go through it in person at the IPFS workshop. I suspect you have an additional assumption or two on top of the ucan spec that I'm not assuming.

BITSWAP.md
```diff
@@ -196,6 +197,94 @@ message Message {
...
+    2. If C sends S a Have request for data S has but is not currently willing to give to C, S may respond with an `AuthRequired` BlockPresence
+      1. For each of the items that S sends back Blocks or BlockPresences for they may append any `tokens` they want. Recommended tokens include those that would inform C how they could convince S to give them access to the blocks they want.
+    2. If C sends S a Have request for data S does not have (or has but is not willing to tell C it has) and C has requested for DontHave responses then S should respond with DontHave
+    3. S may choose to include the number of bytes that are pending to be sent to C in the response message
+    4. If C asked for a block that S has but it is bigger than the maximum block size it should return `BlockTooBig`
+3. When C no longer needs a block it previously asked for it should send a Cancel message for that request to any peers that have not already responded about that particular block. It should particularly send Cancel messages for Block requests (as opposed to Have requests) that have not yet been answered.
+
+### Tokens
```
guseggert, line 228: IIUC, "tokens" are not auth-specific, and are really just arbitrary key-value metadata pairs which are useful for passing auth tokens, among other things. If that's the case, would it make sense to call this something other than "token"? Generally "token" is used in auth contexts, and so its use here implies that it's intended only for auth, which isn't really the case.

willscott: I would say that while it can be more general, token isn't a particularly bad name for this.

willscott:
It would be great to get a summary of what needs to happen to move this from draft to ready for review, since this has been sitting for a few months now and we know there are several implementation interested in making use of this mechanism.

BITSWAP.md
aschmahmann: Done in https://github.com/ipfs/specs/pull/270/commits/661ab5bf195a6d7094851dbec85545cc8052e9bb

BITSWAP.md
aschmahmann: Numbers skipped in https://github.com/ipfs/specs/pull/270/commits/acc0fe0f10016b5b397704293d15f72db25bae17

aschmahmann:
> It would be great to get a summary of what needs to happen to move this from draft to ready for review

See https://github.com/ipfs/specs/pull/270#issuecomment-1156612945.

I've fixed up the small changes in the spec, but we're still lacking multiple groups planning to implement and utilize this behavior so that we can make sure that the abstraction here is sufficiently usable for folks. Maybe it's missing something or will be a pain to work with on the implementation, especially on the client side.

As has been noted multiple times in this PR depending on what types of tokens you're planning on passing along things implementations might need to deviate from how they work today, particularly on the client side. IMO merging a spec change to a core protocol while it's not being used is not responsible.

> and we know there are several implementation interested in making use of this mechanism.

While there have been multiple groups asking for this functionality in the past, AFAIK Peergos is the only group that has expressed an interest in actually building out the pieces they need to use this and IIUC they're currently fine using their fork of Bitswap 1.2.0 which does enough of the job for them but doesn't really support any other use cases.

Some of the other groups that initially wanted this (e.g. for wanting to support paid retrieval of data served over Bitswap, limiting serving data using some permissioning schemes similar to bearer tokens or UCANs, etc.) have since decided that their need is not yet high enough for them to want to prioritized implementing the changes. If this has changed and we have  an additional group or two that want to implement and utilize these changes then it'd be great to see how the implementations go so we can either merge the spec as is, or make any changes that end up being needed for implementers.

PR #272, opened 2022-03-25
Spec proposal for Reframe

aschmahmann:
This is a proposal for an extensible protocol that we can use for endeavors such as delegated routing.

A few things to note for people less familiar with this project
1. Ask questions 😄, ideally the spec background can properly motivate its existence and how it works so we should fill in the blanks if things are missing
2. It is expected that this protocol will be in flux until we have some example client + server implementations operational, as is generally the case when trying to get feedback on protocols before they're already running in production and it's too late to change them
3. It is expected for this protocol to evolve to encapsulate more methods and use cases. Some glaringly obvious omissions for even covering the base cases of a delegated routing protocol are: Lack of methods for handling peer records, lack of a function for advertising provider records. For now we're just trying to tackle this a bit at a time based on what is needed.

There is also a bunch of information in here that emerged with some helpful conversations with folks that may not be properly reflected in this PR, so definitely ask clarifying questions, concerns, etc.

----

I also want to note that while the pressing concern here is mostly around the `FindProviders` method the IPNS methods seemed generally to be easy wins that are both useful and let us describe both `Put` and `Get` style methods.

### DIFF
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -1,0 +1,287 @@
+# ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) Reframe
+
+**Author(s)**:
+- Adin Schmahmann
+- Petar Maymounkov
+
+**Maintainer(s)**:
+
+* * *
+
+**Abstract**
+
+The Reframe protocol is designed for request-response messages that is sufficiently generic and extensible to evolve over time as new needs for it arise. This includes separately defining the transport and message serialization mechanisms from the actual method types and semantics.
+
+The initial use case motivating the protocol's design is to help peers discover various routing hints that enable them to find the content or service they are actually looking for. The extensibility required in this case is that various routing systems could share the protocol so that applications can work on them generically and that recursive routing systems could choose to use the protocol themselves to maximize interoperability.
+
+# Organization of this document
+
+- [Introduction](#introduction)
+- [Spec](#spec)
+- [Intended Upgrade Paths](#intended-upgrade-paths)
+- [Implementations](#implementations)
+
+# Introduction
+
+The Reframe protocol is a request-response based protocol. Upon receiving a request a Reframe server should respond to clients with information they have pertaining to the request. It's possible that a Reframe server may not have all of the information necessary for a client to actually get the data they need, but even partial resolution is acceptable. For example, if a peer is looking to download some block of data a Reframe server may only know about some peers that have the data but not their network addresses. That's ok and the client can always choose alternative mechanisms to discover the missing information. 
+
+# Spec
+
+To build an implementation of the protocol it is required to both define a transport for the messages and the method types supported by the given implementation. Over time both the method and transport types may grow and also be included in this specification.
+
+## Interaction Pattern
+
+Given that a client C wants to request information from some server S:
+
+1. C opens sends a request to S that is a single message
+2. S will respond to C with either a single message or a group of messages
+
+## Transports
+
+### HTTP + DAG-JSON
+
+All messages MUST be encoded as DAG-JSON.
+
+Requests MUST be sent as HTTP POST requests to the endpoint where the message data is sent in the body with chunked encoding.
+
+Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time.
+
+If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of concatenated DAG-JSON messages (i.e. `{Response1}{Response2}...`).
+
+Requests and responses SHOULD come with `version=1` as a field in the Content-Type header.
+
+Note: This version header is what allows the transport to more easily evolve over time (e.g. if it was desired to change the transport to support other encodings than DAG-JSON, utilize headers differently, move the request data from the body, etc.). Not including the version number is may lead to incompatibility with future versions of the transport.
+
+## Protocol Message Overview
+
+We can represent each message as an IPLD Schema to denote its abstract representation independent of the serialization scheme. 
+
+To help visualize example messages and illustrate implementation of the first concrete transport (HTTP + DAG-JSON) example messages will be given in DAG-JSON.
+
+The payload sent as a request is a single message. The response is a repeated set of messages until the transport signals that the response is completed (e.g. closing a stream, connection, etc.)
+
+Reception, on the server or client side, of:
+1. New or unknown fields in any request or response message are ignored (rather than causing an error)
+2. Missing fields or unknown union cases cause a terminal error
+
+### Known Methods
+
+The known Request types are the following and are described below:
+
+```ipldsch
+type Request union {
+    | "IdentifyRequest" IdentifyRequest
+    | "FindProvidersRequest" FindProvidersRequest
+    | "GetIPNSRequest" GetIPNSRequest
+    | "PutIPNSRequest" PutIPNSRequest
+}
+```
+
+The known Response types are the following and are described below:
+
+```ipldsch
+type Response union {
+    | "IdentifyResponse" IdentifyResponse
+    | "FindProvidersResponse" FindProvidersResponse
+    | "GetIPNSResponse" GetIPNSResponse
+    | "PutIPNSResponse" PutIPNSResponse
+    | "Error" Error
+}
+```
+
+Note: Each Request type has a corresponding Response type.
+Every message except the Error type should end in Request/Response.
+
+#### Error
+
+The Error message type should be used in Responses to indicate that an error has occurred.
+
+```ipldsch
+    type Error struct {
+        Method String
+    }
+```
+
+#### Identify
+
+A message for discovering which messages a server supports. May be used by applications to optimize which servers get sent which types of requests.
+
+```ipldsch
+    type IdentifyRequest struct {}
+
+    type IdentifyResponse struct {
+        Methods [String]
+    }
+```
+
+`IdentifyResponse` should return the set of supported methods aside from the "Identify" method.
+
+##### DAG-JSON Examples
+
+Request:
+```
+{"IdentifyRequest" : {}}
+```
+
+Response:
+```
+{"IdentifyResponse" : {
+    "Methods" : ["FindProviders", "GetIPNS"]
+}}
+```
+
+#### FindProviders
+
+A message for finding nodes that have an interest in a given key. Some common examples include finding which peers have advertised that they have a given CID, or which peers are interested in a given pubsub topic.
+
+```ipldsch
+    type FindProvidersRequest struct 
+        Key &Any
+    }
+```
+
+Note: While the Key is a CID it is highly recommended that server implementations treat these FindProvidersRequests as if they were for the multihash. This allows consumers that perform lookups by CID but with a different IPLD codec than one that is known to the server to still get back valid responses.
+
+```ipldsch
+    type FindProvidersResponse struct {
+        Providers [Provider]
+    }
+
+    type Provider struct {
+     Node Node
+     Proto optional [TransferProtocol]
+     }
+
+    # Node describes a node (identity, network address, authentication, whatever else applies)
+    # We expect different types of nodes, e.g. peer, miner, public IP, etc.
+    type Node union {
+     | Peer "peer"
+    } representation keyed
+
+    type Peer struct {
+        ID Bytes # The binary representation of a libp2p peerID https://github.com/libp2p/specs/blob/f433ad595224cf33d916c166d1738f11aadfa9f7/peer-ids/peer-ids.md
+        Multiaddresses optional [Bytes] # Each element in the list is the binary representation of a complete multiaddr without a peerID suffix
+                                        # https://github.com/multiformats/multiaddr/blob/d02c681171629912dd977f4f5d4b3b72dcac507f/README.md
+                                        # https://github.com/libp2p/specs/blob/98e0ca0e54fd83f7b8c86d7e9677e1e430da5810/addressing/README.md#multiaddr-in-libp2p
+    }
+
+    # Note: This is not quite valid IPLD Schema because the "structured" represenation does not exist yet, nor does the SingletonInt mechanism
+    # However, we can use this to describe that we have a union keyed by numbers rather than strings.
+    type TransferProtocol union {
+        | Bitswap
+        | GraphSync-FILv1
+    } representation structured 
+    
+    type Bitswap struct {
+        ID SingletonInt(0x0900) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L133
+    }
+    
+    type GraphSync-FILv1 struct {
+        ID SingletonInt(0x0910) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L134
+        DealID Bytes # TODO: Is there a spec for what goes here?
+    }
+```
+
+##### DAG-JSON Examples
+
+Request:
+```
+{"FindProviders" : {
+    "Key" : {"/":{"bytes":"AXIUBPnagss"}}
+}}
+```
+
+Response:
+```
+{"FindProviders" : {
+    "Key" : {"/":{"bytes":"AXIUBPnagss"}},
+    "Providers" : [
+        {"Node":{"Peer":{"ID":{
+            "/":{"bytes":"EncodedPeerID"}},
+            "Multiaddresses":[{"/":{"bytes":"EncodedAddr1"}}, {"/":{"bytes":"EncodedAddr2"}}]}}},
+        {"Node":{"Peer":{"ID":{
+            "/":{"bytes":"EncodedPeerID2"}},
+            "Multiaddresses":[{"/":{"bytes":"EncodedAddr1"}}, {"/":{"bytes":"EncodedAddr2"}}]}}},
+                "Proto" : [{
+                    "ID": 2320, # the integer of the graphsync-filv1 code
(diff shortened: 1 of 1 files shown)


REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+
+### HTTP + DAG-JSON
+
+All messages MUST be encoded as DAG-JSON.
+
+Requests MUST be sent as HTTP POST requests to the endpoint where the message data is sent in the body with chunked encoding.
+
+Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time.
+
+If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of concatenated DAG-JSON messages (i.e. `{Response1}{Response2}...`).
```
willscott, line 49: can these each be on a new line?

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+## Interaction Pattern
+
+Given that a client C wants to request information from some server S:
+
+1. C opens sends a request to S that is a single message
+2. S will respond to C with either a single message or a group of messages
+
+## Transports
+
+### HTTP + DAG-JSON
```
willscott, line 41: is the goal that this protocol should be possible to implement a web client for?
If so, the custom `Content-Type` header extension may be harder to pull out in some browser situations vs a separate `x-version` style header. may be worth looking at what's going to be most compatible?

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+
+All messages MUST be encoded as DAG-JSON.
+
+Requests MUST be sent as HTTP POST requests to the endpoint where the message data is sent in the body with chunked encoding.
```
willscott, line 45: does this disallow `HTTP/2` transport upgrades, which subsume chunked encoding with it's own transport framing?

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+        | GraphSync-FILv1
+    } representation structured 
+    
+    type Bitswap struct {
+        ID SingletonInt(0x0900) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L133
+    }
+    
+    type GraphSync-FILv1 struct {
+        ID SingletonInt(0x0910) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L134
+        DealID Bytes # TODO: Is there a spec for what goes here?
```
willscott, line 184: https://github.com/filecoin-project/index-provider/blob/main/metadata/graphsync_filecoinv1.ipldsch

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+    }
```
aschmahmann, line 182: Did I correctly describe this function, both by name and the fields you need? When should someone use this field, what protocols should they be supporting at minimum?

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+        ID Bytes # The binary representation of a libp2p peerID https://github.com/libp2p/specs/blob/f433ad595224cf33d916c166d1738f11aadfa9f7/peer-ids/peer-ids.md
+        Multiaddresses optional [Bytes] # Each element in the list is the binary representation of a complete multiaddr without a peerID suffix
+                                        # https://github.com/multiformats/multiaddr/blob/d02c681171629912dd977f4f5d4b3b72dcac507f/README.md
+                                        # https://github.com/libp2p/specs/blob/98e0ca0e54fd83f7b8c86d7e9677e1e430da5810/addressing/README.md#multiaddr-in-libp2p
+    }
+    
+    # An alternative mechanism might be to have a union keyed by the protocol identifier that has more detailed
+    # information inside the bytes. 
+
+    # Note: This is not quite valid IPLD Schema because the "structured" represenation does not exist yet, nor does the SingletonInt mechanism
```
aschmahmann, line 168: This was done in order to accommodate the requests to use the integer protocol identifiers even though IPLD Schemas currently not supporting having unions keyed this way.

@mvdan @warpfork does this seem like a reasonable representation to add to IPLD Schema? We can talk in this thread or make a separate issue in ipld/ipld about this.

IIUC @petar and @warpfork have talked a bit about singletons in IPLD Schemas already.

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+    # However, we can use this to describe that we have a union keyed by numbers rather than strings.
```
aschmahmann, line 169: Some alternative representations could have been:

A: 
```
type TransferProtocol struct {
    ID Bytes
    Hint optional Bytes # Or Any?
}
```

and make more of the information about what fields are needed by different protocols live in documentation/not in the schema.

B: 

One of the existing union representations using the integer codes as strings although that seems pretty weird given it's not a string

```
type TransferProtocol union {
     | Bitswap "2304"
     | GraphSync-FILV1 "2320"
} representation keyed
```

C: Use an existing union representation but using protocol names rather than numbers

I don't feel super strongly about the way we encode this and was mostly trying to take requests from other parties into account here. If there are strong preferences to go with one of the above approaches or something else here's the place to make them known.

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+Response:
+```
+{"PutIPNSResponse : {}"}
+```
+
+# Method Upgrade Paths
+
+It is acknowledged that the initial methods and mechanisms of this protocol will likely need to change over time and that we should prepare for how to do so without the need to wholesale replace this protocol with an alternative.
+
+2. If it is desired to add optional parameters to a given method (in either a request or response) new fields can be added. It is recommended that new fields be submitted to this spec to prevent conflicts and for unsubmitted features to be added using namespaces unlikely to conflict (e.g. `MyProject-Field`).
```
aschmahmann, line 282: I don't know if within IPLD land we have any precedence for this type of namespacing we want to be consistent with.

In libp2p land we use paths (e.g. `/ipfs/bitswap/1.2.0`) and HTTP land has X Headers. Anything  different from `MyProject-...` we want to encourage here @mvdan?

REFRAME.md
```diff
@@ -0,0 +1,287 @@
+# ![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) Reframe
```
aschmahmann, line 1: You have a better name? Here's the place to pitch it, otherwise I'm less concerned about names and more concerned about specs and code and we can move on 🙃.

REFRAME.md
aschmahmann: Any reason why/why not? JSON is a self-delimiting protocol so my understanding of things is that many decoders will happily just keep on parsing once they're done with a single message.

I don't feel strongly either way, trying to understand the context.

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+
+All messages MUST be encoded as DAG-JSON.
+
+Requests MUST be sent as HTTP POST requests to the endpoint where the message data is sent in the body with chunked encoding.
+
+Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time.
+
+If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of concatenated DAG-JSON messages (i.e. `{Response1}{Response2}...`).
+
+Requests and responses SHOULD come with `version=1` as a field in the Content-Type header.
```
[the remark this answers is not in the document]
aschmahmann: From @willscott:

> is the goal that this protocol should be possible to implement a web client for?

Yes.

> If so, the custom Content-Type header extension may be harder to pull out in some browser situations vs a separate x-version style header. may be worth looking at what's going to be most compatible?

Good question, I don't strongly recall which headers made the most sense here for browsers. @lidel probably has thoughts here though.

REFRAME.md
aschmahmann: Continuing below https://github.com/ipfs/specs/pull/272/files#r835590269 (your speedy review beat me to my bulk review with questions 😄)

REFRAME.md
aschmahmann: @lidel @guseggert I think you all have more context in HTTP land than I do so figure you all might have better answers here

REFRAME.md
aschmahmann: Thanks 🙏. Fixed in d497d2c71087d5d8af77c653c5701eeb3aa64541

REFRAME.md
```diff
@@ -0,0 +1,291 @@
...
+    type TransferProtocol union {
+        | Bitswap
+        | GraphSync-FILv1
+    } representation structured 
+    
+    type Bitswap struct {
+        ID SingletonInt(0x0900) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L133
+    }
+    
+    type GraphSync-FILv1 struct {
```
aschmahmann, line 179: When should someone use this field, what protocols should they be supporting at minimum? IIUC there are multiple versions of `/ipfs/graphsync` and `/fil/data-transfer` that exist. Does this cover all of them via protocol negotiation (similar to the Bitswap case) or is this meant to cover a more limited subset?

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+    # Note: This is not quite valid IPLD Schema because the "structured" represenation does not exist yet, nor does the SingletonInt mechanism
+    # However, we can use this to describe that we have a union keyed by numbers rather than strings.
+    type TransferProtocol union {
+        | Bitswap
+        | GraphSync-FILv1
+    } representation structured 
+    
+    type Bitswap struct {
+        ID SingletonInt(0x0900) # https://github.com/multiformats/multicodec/blob/f5dd49f35b26b447aa380e9a491f195fd49d912c/table.csv#L133
+    }
```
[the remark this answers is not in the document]
willscott: that seems reasonable / this would change once we decide on a deprecation policy.
From what I can see of current implementations, they support downgrading back to all previous versions currently. We could imagine defining a `VersionedBitswap` transport that includes indicators of min/max supported version as it's payload.

REFRAME.md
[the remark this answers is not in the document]
guseggert: Agree with @willscott here, chunked encoding should be optional.

Suggested wording:

> Requests MUST be sent as HTTP POST requests. If a client or server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.

(This leaves the server free to not support chunked-encoded reqs...rationale is that, AFAIU, we only really care about making sure that servers can send chunked-encoded responses.)

guseggert: It doesn't disallow transport upgrades per se, b/c the req is still handled as HTTP/1.1 on an upgrade, whereas the response is sent back on a half-closed HTTP/2 stream. But it disallows connection reuse b/c you can't make subsequent reqs using HTTP/1.1 chunked encoding on an HTTP/2 stream. It also disallows https with HTTP/2, because in that case HTTP/2 uses ALPN instead of Upgrade headers for negotiation, so the initial req is sent as HTTP/2. But all this goes away if we don't require HTTP/1.1 chunked-encoded requests, which seems reasonable to me.

REFRAME.md
```diff
@@ -0,0 +1,291 @@
...
+        Multiaddresses optional [Bytes] # Each element in the list is the binary representation of a complete multiaddr without a peerID suffix
+                                        # https://github.com/multiformats/multiaddr/blob/d02c681171629912dd977f4f5d4b3b72dcac507f/README.md
+                                        # https://github.com/libp2p/specs/blob/98e0ca0e54fd83f7b8c86d7e9677e1e430da5810/addressing/README.md#multiaddr-in-libp2p
+    }
+
+    # Note: This is not quite valid IPLD Schema because the "structured" represenation does not exist yet, nor does the SingletonInt mechanism
+    # However, we can use this to describe that we have a union keyed by numbers rather than strings.
+    type TransferProtocol union {
+        | Bitswap
+        | GraphSync-FILv1
```
aschmahmann, line 172: Feature request after talking with @willscott it'd be nice if whichever union type we went with here we could add some fallback where servers could return data for transports they don't quite understand without needing someone to make a spec PR, update the implementations, etc. If clients don't know how to process the result, they can just throw it out.

In the structured case or keyed case it'd be nice to have an Any fallback for union cases we don't quite know (e.g. for strings we don't understand or for singletonInts we don't know about). We don't want to have to fallback to the less structured `type TransferProtocol struct { ID Bytes Hint optional Bytes # Or Any? }` but we can if we need to.

cc @petar @mvdan @warpfork

REFRAME.md
```diff
@@ -0,0 +1,291 @@
...
+```
+{"PutIPNSRequest" : {
+    "ID" : {"/":{"bytes":"AXIUBPnagss"}},
+    "Record" : {"/":{"bytes":"AXIUBPnagss"}}
+}}
+```
+
+Response:
+```
+{"PutIPNSResponse : {}"}
```
aschmahmann, line 279: I think for now this is ok. It might be that in the future we want to return other things like if the server knows about a more recent/better IPNS record (which is what the DHT code does), however that kind of situation also seems like something that'd be a reasonable error (i.e. indicating that the put is too old to be valid and that you can do a Get if you want the latest value).

REFRAME.md
```diff
@@ -0,0 +1,290 @@
...
+}}
+```
+
+#### FindProviders
+
+A message for finding nodes that have an interest in a given key. Some common examples include finding which peers have advertised that they have a given CID, or which peers are interested in a given pubsub topic.
+
+```ipldsch
+    type FindProvidersRequest struct 
+        Key &Any
```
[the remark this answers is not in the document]
warpfork: `&Any` is a fine way to describe the CID case.

I feel like maybe a union is indeed gonna be worth inserting here, though.  Possibly even a keyed one.  If I imagine myself writing destructuring code for handling a message like this, I'd probably want a pretty explicit logical marker for if the thing coming is a content identifier or a pubsub topic (or whatever else).

REFRAME.md
warpfork: B or C both seem fine to me.  A is viable but sad.

As a general rule, the impetus for choosing B or C is that they mean it's perfectly clear how things end up serialized in DAG-JSON (where we're limited to string keys in maps, by the aim of compatibility with JSON itself).

Beyond that, I tend not to care.  Season to taste.  How much entropy / specificity / descriptivism / whatever-you-want-to-call-it one wants to stir into strings is up to the author.  Stringified magic numbers don't offend me any more than other magic strings.

warpfork: Approach A gives the most open ended holder for totally unknown unstructured data, but at the cost of being the least useful.  E.g. there would almost certainly be a second layer of parsing and pattern matching going on inside that.  Doable but probably will be perceived by most developers as unpleasant and complicated.

REFRAME.md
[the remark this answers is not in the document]
lidel: >  I'm not sure about whether this is best as a SHOULD or MUST. It doesn't seem required, but is a footgun waiting around for people who don't use it

I would change it to MUST in both request and response.

One may argue that Requests could stay as "SHOULD", as server can always respond with HTTP 400 Bad Request if message format is invalid,  but there is a risk of subtle bugs where the same message is valid in both versions, and some field has different logical meaning in two different versions.

> the custom Content-Type header extension may be harder to pull out in some browser situations vs a separate x-version style header. may be worth looking at what's going to be most compatible?

IMO better to not add any custom headers.

If we care about ease of access, then there is no difference in getting value from `Content-Type` vs `X-foo`. Both need to be [safelisted via CORS preflight request](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers#example_preflight_request) to be accessible from JS running on a webpage. 

Extracting header parameters is usually supported with utility functions in every language/libraries, e.g. golang stlib: https://go.dev/play/p/G-NfIqt8YuW


I suggest:

```suggestion
Requests and responses MUST come with `version=1` as a _Required Parameter_  in the `Content-Type` HTTP header.

```

REFRAME.md
lidel: dag-json messages separated by `\n` sgtm.

Prior art: RPC endpoints like  /api/v0/refs/local use  http://ndjson.org (which afaik is the same/similar concept).
js-ipfs-http-client is able to [parse such response](https://github.com/ipfs/js-ipfs/blob/9a40109632e5b4837eb77a2f57dbc77fbf1fe099/packages/ipfs-http-server/test/utils/all-ndjson.js) and [yield incoming messages as soon each gets parsed](https://github.com/ipfs/js-ipfs/blob/c51b1609f9ba825b14ad1efd0b99f6af6805c4e0/packages/ipfs-http-client/src/refs/local.js#L22), which is pretty good to have when processing bigger batches.

REFRAME.md
aschmahmann: https://github.com/ipld/ipld/issues/194

REFRAME.md
BigLep: Discussion had during IPFS standup: we're going to do option B

REFRAME.md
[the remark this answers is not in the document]
aschmahmann:
```suggestion
Requests MUST be sent as HTTP POST requests. If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
```

REFRAME.md
aschmahmann: It looks like there might be some implementation details in getting fallbacks merged into the schema schema, however there seems to be general alignment that this type of behavior is sane, so we'll use the syntax from the linked issue to represent the fallbacks required by the protocol.

guseggert (approved):

PR #275, opened 2022-04-08
generalize content-type in reframe spec

petar:
I think the content-type in the REFRAME API spec should not be viewed as a part of the specific REFRAME API, but rather as a part of the RPC networking framework that the REFRAME API uses. I.e. there may be other APIs that use the same networking RPC stack. Consequently, the content-type should identify the networking framework itself rather than make a reference to the reframe API. So I am proposing we rename it from `application/vnd.ipfs.reframe+dag-json; version=1` to `application/vnd.ipfs.http+dag-json; version=1`.

### DIFF
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -40,7 +40,7 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.reframe+dag-json; version=1`
+All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.http+dag-json; version=1`
 
 
 Requests MUST be sent as HTTP POST requests. If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -40,7 +40,7 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.http+dag-json; version=1`
+All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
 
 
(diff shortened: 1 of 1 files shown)


lidel (commented):
_Apologies if this is a naive question (late friday, but afraid I will forget to ask on monday)_

If we want to indicate it is a "generic RPC for IPFS", wouldn't `application/vnd.ipfs.rpc+dag-json; version=1` do the trick?

My thinking: if we have something other than HTTP, let's say WebTransport at some point (https://github.com/ipfs/in-web-browsers/issues/190),  what would be the difference between below two?
-  `application/vnd.ipfs.http+dag-json; version=1`
-  `application/vnd.ipfs.webtransport+dag-json; version=1`

Do we send the same dag-json in both cases, no matter if we use HTTP or WebTransport?
If so, what is the value added by having separate types?

petar:
> _Apologies if this is a naive question (late friday, but afraid I will forget to ask on monday)_
> 
> If we want to indicate it is a "generic RPC for IPFS", wouldn't `application/vnd.ipfs.rpc+dag-json; version=1` do the trick?
> 
> My thinking: if we have something other than HTTP, let's say WebTransport at some point ([ipfs/in-web-browsers#190](https://github.com/ipfs/in-web-browsers/issues/190)), what would be the difference between below two?
> 
> * `application/vnd.ipfs.http+dag-json; version=1`
> * `application/vnd.ipfs.webtransport+dag-json; version=1`
> 
> Do we send the same dag-json in both cases, no matter if we use HTTP or WebTransport? If so, what is the value added by having separate types?

This is a good question. In fact, since the content-type describes the encoding (dag-json) and the semantics (rpc), and not the transport (http or webtransport), the string should indeed probably just be `application/vnd.ipfs.rpc+dag-json; version=1`. Good call!

REFRAME.md
```diff
@@ -40,7 +40,7 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.reframe+dag-json; version=1`
+All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.http+dag-json; version=1`
```
lidel, line 43: @petar  so something like this?
```suggestion
All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
```

lidel (approved):

PR #280, opened 2022-05-13
IPIP-280: App Conventions for HTTP Gateway Detection

### LINKED ISSUE #192: Easy-to-implement gateway-less IPFS URLs
(I am new to IPFS contribution and have no idea which repo/forum I should use for this.  Please move as appropriate.)

**Problem:** I want to point all my tools/configs at content stored in IPFS because IPFS is awesome.  It is not feasible (now, or perhaps ever) to add a full dhtclient node implementation to every tool, particularly across every programming language.

**Non-solution:** I can use HTTP gateway URLs, but then my configs are not portable across sites.  I want to use _my_ gateway at my site (because caching is important for large content), but that gateway host doesn't make sense for other sites.

**Implication:** I think we need to start using "real" (gateway-less) [IPFS URLs](https://github.com/ipfs/go-ipfs/issues/1678) (ipfs:/...) now for portability, and invent a trivial-to-implement "client" that can fetch IPFS content using libraries that are already commonly available in programming languages.  Importantly, it should be feasible to propose adding this trivial-to-implement client to _every_ tool that wants to fetch content.

Note that once we have this, we can evolve the actual client implementations over time without changing the high-level IPFS semantics exposed to users.

**Proposal:** I think the only "trivial" solution available right now is converting IPFS URLs to HTTP(S) gateway URLs, and then reuse existing ubiquitous HTTP(S) libraries.  I propose we define a semi-standard way to "discover" the local IPFS gateway, and then implementations can just do a trivial URL string manipulation to switch ipfs:/path to $gateway/path (with some protection against "../" attacks).

**Concrete proposal:**
1. Declare a semi-standard `IPFS_GATEWAY` environment variable to specify the HTTP(S) URL prefix.
2. Fallback to a `_ipfs-gateway._tcp` DNS-SD TXT lookup for easy site-wide configuration (controversial).
3. Fallback to `https://ipfs.io/` (controversial?).
4. Fetch content over regular HTTP(S) from the gateway.  In particular, we're trusting the discovery result, gateway implementation, and gateway<->client communication to offload content/checksum verification.
5. Start proposing patches to various non-IPFS projects that take advantage of the above to add support for ipfs:/ URLs.
6. Profit.  (Actually, continue to iterate on the internals of the above when/if "more native" approaches become available - eg long-lived golang programs could embed a full dhtclient node)

At the moment I have only considered read-only workloads (this is the _vast_ majority of my personal use-cases), but the above should extend to write as well - assuming the gateway allows uploads.

Thoughts?  I have a specific tool I would like to add this to, and was hoping to get some consensus on the approach before starting on code.

markg85:
This should help other application developers to have an idea of what's needed to add IPFS support to their application.

cc @Stebalien and @autonome Conversations with mainly you two have led to this spec being implemented as is written here in ffmpeg.

- closes #192


markg85:
Ping. What's still required for this document to get merged?

pimterry:
Hi @markg85, two quick comments on this from some related discussion in https://github.com/ipfs/js-ipfs/issues/4010:

* Could we extend this just slightly, to also specify gateway detection for browser JS environments? This is a common use case, which I think could easily closely match what you're suggesting here, by using a global `window.IPFS_GATEWAY` env var to allow browser and browser extensions to inject gateway configuration into JS apps. Similarly, those apps should use that configuration if available, or use the fallback gateway if not.

* This spec covers gateways only, but that's not the only way people may want to integrate IPFS into their applications. There's many interesting use cases where 3rd-party applications might want to interact with the network more generally, by embedding a full IPFS node. When doing so, it would be useful to provide very similar mechanisms to detect a user's preferred network configuration (particularly for web applications - see https://github.com/ipfs/js-ipfs/issues/4010 for lots more context). Of course I don't want to derail this PR, which can happily proceed ignoring all that, but I think this would fit well as an extension to this 'Integration' guide later on. Do you think that makes sense? How about if I open up a PR against your branch here to start discussion on that too, but independently?

INTEGRATION.md
```diff
@@ -0,0 +1,46 @@
...
+
+<img src="img/gateway_decision_tree.drawio.png" width="503" height="716" />
+
+### Gateway from command argument
+An application can opt to support a command line option to provide a gateway. A user should not _need_ to provide this and should therefore be considered optional. However, if a user does provide this option then it should overrule any other gateway detection and be used as the gateway to use. An example implementation that is doing this is ffmpeg with the ffplay utility. It allows the `-gateway` argument which by default is empty but can be set like: `-gateway http://127.0.0.1:8080` and would then be used to handle `ipfs://<cid>` or `ipns://<cid>`.
+
+### Gateway from IPFS_GATEWAY environment variable
+When there is no command line argument, the `IPFS_GATEWAY` environment is next. If it contains a value, it should be used as gateway.
+
+### Gateway file from IPFS_PATH environment variable or home folder ipfs data
```
lidel, line 28: This section needs to be updated:
- Checking in $IPFS_PATH and $HOME/.ipfs is fine but is specific to Kubo and should not be the only place.
- ..but we should play nice with well established conventions and follow platform specific directories and ALSO check platform-specific places, for example
  - on linux, Linux something under `XDG_CONFIG_HOME`  (default: `~/.config`)  – maybe `XDG_CONFIG_HOME/ipfs/gateway` ?
  - on windows → ?
  - on macos → ?

INTEGRATION.md
```diff
@@ -0,0 +1,46 @@
...
+If the `IPFS_PATH` environment variable is defined, it should point to the ipfs data folder. If this environment variable isn't defined then an attempt should be made to see if `$USER/.ipfs` exists and consider that to be the IPFS data folder when it does.
+
+If this turns out to be an existing path the existence of the `gateway` file in that path should be checked for. If gateway file exists then the first line should be considered to be the full url to the local IPFS gateway.
+
+Do note that this gateway file logic relies on the future implementation of having the gateway file be auto-generated. It's specced [here](https://github.com/ipfs/go-ipfs/issues/8847) and is approved to be implemented. Implementers of this spec should act as if that gateway file already exists!
+
+## Validation and fallback gateway
+All of the above describes what should ideally be done. Validation of any kind is up to the application implementing IPFS support. To provide an "always working" feeling, implementers are recommended to use a fallback gateway and they should pick `dweb.link`. That gateway is maintained by Protocol Labs and is allowed to be used for this purpose.
```
lidel, line 36: Needs to be changed. Many people will read this document as their entry point into IPFS and Gateways.
First, important to  state difference between modes and risks:
   - "trusted localhost gateway" that does the deserialization and validation and returns  raw data
   - "untrusted, but verifiable third-party gateway" which does not do deserialization, but allows client to verify the data
Then, 
  - give a conservative approach  ffmpeg did  (only trusted local gateway, third-party must be opt-in)
  - document that only trusted non-localhost gateways should be used, and if cluent can't trust gateway, [it should  retrieve data in verifiable format](https://docs.ipfs.tech/reference/http/gateway/#trustless-verifiable-retrieval) and compare hashes locally.

INTEGRATION.md
```diff
@@ -0,0 +1,46 @@
+# ![](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square) Integration  <!-- omit in toc -->
```
lidel, line 1: Needs
- a header  – see https://github.com/ipfs/specs/blob/main/template.md
- short IPIP in `IPIP/` with motivation why we add this – see https://github.com/ipfs/specs/blob/main/IPIP/0000-template.md

lidel (changes requested):
@markg85  time  to revisit this, now that we have IPIP process. some pointers below :)

markg85:
> Hi @markg85, two quick comments on this from some related discussion in [ipfs/in-web-browsers#197](https://github.com/ipfs/in-web-browsers/issues/197):
> 
> * Could we extend this just slightly, to also specify gateway detection for browser JS environments? This is a common use case, which I think could easily closely match what you're suggesting here, by using a global `window.IPFS_GATEWAY` env var to allow browser and browser extensions to inject gateway configuration into JS apps. Similarly, those apps should use that configuration if available, or use the fallback gateway if not.

I get the mindset here and it would help a lot! Not just with apps using IPFS but also with js libraries (think for example video.js) where we'd like to have IPFS integration in at some point in time too. Having a `window.IPFS_GATEWAY` would surely help!

But.. how?

When IPFS Companion is used, injecting a variable is doable.
But for IPFS integration in brave, that would be a code change on the brave side.

The downside is another point of metric collection to profile people. And for that reason alone i don't think this variable would be accepted. Even if we would all like it and add it to the spec, the browser security minded folks would probably have an issue with it.

@lidel do you have a opinion on this?

> * This spec covers gateways only, but that's not the only way people may want to integrate IPFS into their applications. There's many interesting use cases where 3rd-party applications might want to interact with the network more generally, by embedding a full IPFS node. When doing so, it would be useful to provide very similar mechanisms to detect a user's preferred network configuration (particularly for web applications - see [As a user, there is no mechanism to control the IPFS networking of a webapp I use in-web-browsers#197](https://github.com/ipfs/in-web-browsers/issues/197) for lots more context). Of course I don't want to derail this PR, which can happily proceed ignoring all that, but I think this would fit well as an extension to this 'Integration' guide later on. Do you think that makes sense? How about if I open up a PR against your branch here to start discussion on that too, but independently?

Usually i'd say "yes, go for it!" but in this case, waiting is probably the smarter move. If you'd go for that you'd have a spec wip branch on top of my wip branch. That's a little too much wip. Wait till this spec lands (it's mostly there) before you consider making a spec on top of this one.

markg85:
> @markg85 time to revisit this, now that we have IPIP process. some pointers below :)

Sweet!
Will update it all in the next couple days. Tank you for all the pointers, much appreciated!

INTEGRATION.md
markg85: I dropped this section completely. The rationale is that this spec just isn't covering that part of gateways at all. It "merely" describes how to pick a gateway and how to expose one.

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
...
+
+### Gateway from IPFS_GATEWAY environment variable
+
+When there is no command line argument, the `IPFS_GATEWAY` environment is next. If it contains a value, it will be used as gateway.
+
+### Gateway file
+
+**This feature is only for IPFS implementers, not for IPFS application integrations.**
+
+When the implementation provides a gateway (and it's not disabled through other means) then it should make that known to other local applications. The gateway file serves this purpose. It's a file with only 1 single line containing the full http URL to your gateway. For example, it could contain the line: "http://localhost:8080".
```
lidel, line 94: > only 1 single line

This is problematic – user will overwrite their preferences by starting and stopping local node, and stopping a node will break IPFS resolution, or switch user to some application-specific  gateway  hardcoded in apps.

Mind adjusting this to note the file is a list of gateways?

The common use will be someone running Kubo on localhost, and localhost gateways should be prioritized, but if local gateway is down someone should be able to specify fallbacks.

So if I have `gateway` file like this:
```
https://dweb.link
https://cf-ipfs.com
```
And I start local node, it should not overwrite the preexisting entries, but add  local gateway to the top of the file:

```
http://localhost:8080
https://dweb.link
https://cf-ipfs.com
```

and if the local node is shut down it should remove itself from the file, but keep the manually added entries:
```
https://dweb.link
https://cf-ipfs.com
```


The list could be used as a pool (for racing gateways) or in-order fallbacks (if first one is down, app can retry with second, etc).

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
...
+
+The file conditions:
+    1. is named "gateway"
+    2. **only** exists when your implementation actually starts a gateway
+    3. is removed when your implementation shuts down
+
+For historical and compatibility readons, this file shall be placed in:
+$HOME/.ipfs/ thus the resulting end path is going to be $HOME/.ipfs/gateway
```
lidel, line 102: We need to clean this up – as-is, it won't work with Iroh.  cc @b5 

My take: 
- Specification should allow for running Iroh and Kubo on the same PC. 
- Canonical location should be implementation-agnostic, namely `$XDG_CONFIG_HOME/ipfs/gateway`.
Kubo should switch to that.
  - For backward-compatibility, Kubo would  create symlink from `$IPFS_PATH/gateway` to `$XDG_CONFIG_HOME/ipfs/gateway`

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
...
+
+Future implementers must follow the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) where the gateway file will placed in:
+**$XDG_CONFIG_HOME/ipfs/gateway**
+
+The conditions for this file are the same as those in $HOME/.ipfs/gateway.
+
+In the, admitedly rare, event of running multiple IPFS implementations each hosting their own gateway. First-come-first-serve applies here. The application that created the gateway file owns it and takes care of removing it. Subsequent instances or different application should not touch the file if it's already there.
```
lidel, line 109: > Subsequent instances or different application should not touch the file if it's already there.

How would this work? What locking mechanism you have in mind?
If I have `gateway` manually created to have `https://dweb.link`  and I start local IPFS node, it  will not be able to update the file? 

This feels like bad UX/DX, giving me more confidence that 
- this file should be a list of URLs (one per line)
- starting  new IPFS node on port `8081` should add localhost URL at the top of the file, but after any already existing localhost nodes:
  ```
  http://localhost:8080
  http://localhost:8081
  https://dweb.link
  ```
  - in the example above, `8080` was already running when `8081` started – this ensures consistent UX – node that started first will remain default local gateway until it exits – starting other localhost nodes will not suddenly hijack requests (it would be bad, especially when running tests and forgetting to adjust HOME dir)

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
...
+To simplify IPFS integrations in third party applications it's important to know if said application can use a gateway. This spec defines the rules to find such a gateway and/or how to influence it.
+
+### User benefit
+
+Users as in applications using IPFS. They get a defined way to find out if their host pc runs an IPFS gateway. Without this spec there is no defined way.
+
+### Compatibility
+
+Kubo currently makes `$HOME/.ipfs/gateway`, this should stay for backwards compatibility.
+A new version implementing this spec should create `$XDG_CONFIG_HOME/ipfs/gateway` in exactly the same fashion.
```
lidel, line 140: If we want to truly enable multiple IPFS implementations, Kubo-specific config needs to go.
```suggestion
Kubo currently makes `$HOME/.ipfs/gateway`, this should stay for backwards compatibility.
A new version implementing this spec should create `$XDG_CONFIG_HOME/ipfs/gateway` in exactly the same fashion.
Future version of Kubo will switch to `$XDG_CONFIG_HOME/ipfs/gateway` and create a symlink at `$HOME/.ipfs/gateway` for backward-compatibility.
```

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
+# Gateway Integration
+
+![Specification status is reliable](https://img.shields.io/badge/status-reliable-green.svg?style=flat-square)
```
lidel, line 3:
```suggestion
![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
```

lidel (changes requested):
Quick feedback:

- make this Kubo-agnostic
- document configuration location on MS Windows and macOS
- ensure manually added entries are not removed by starting/stopping local node
- allow for multiple local gateways to co-exist, and not "steal" traffic

INTEGRATION.md
```diff
@@ -0,0 +1,152 @@
...
+
+**This feature is optional and only for applications integrating IPFS support.**
+
+An application can opt to support a command line option to provide a gateway. If a user does provide this option then it should overrule any other gateway detection and be used as the gateway of choice. If implemented, it's recommended to go for either a `--gateway` or `--ipfs-gateway` argument. It depends very much on the application itself as to which option is most sensible.
+
+An example implementation that is doing this is ffmpeg with the ffplay utility. It allows the `-gateway` argument which by default is empty but can be set like: `-gateway http://127.0.0.1:8080` and would then be used to handle `ipfs://<cid>` or `ipns://<cid>`.
+
+### Gateway from IPFS_GATEWAY environment variable
+
+When there is no command line argument, the `IPFS_GATEWAY` environment is next. If it contains a value, it will be used as gateway.
```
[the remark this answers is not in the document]
markg85: I highly prefer to not specify multiple here.
It would:
- be a pain parsing environments variables properly
- break ffmpeg (it expects 1)
- not match the naming (they indicate 1, not plural)

They also should be seen as "escape options" to get it working. The proper way would be through the gateway file.

INTEGRATION.md
```diff
@@ -35,6 +35,39 @@ Integration here has 2 different meanings.
 
 2. One way to handle the protocols is to use the [HTTP gateway](https://docs.ipfs.tech/concepts/ipfs-gateway/). This document describes how to determine which gateway to use.
 
+### $CONFIG folder, $GWFILE and $LAGACY_GWFILE
```
lidel, line 38: nit: (could be subjective) I would not introduce these variables, they distract the reader from the only variable that this IPIP aims to define, which is `IPFS_GATEWAY`. 

```suggestion
### Configuration file with a list of preferred IPFS Gateways
```

If you want to highlight a term, perhaps remove `$CONFIG`, `$GWFILE` and `$LEGACY_GWFILE`, replace them with nouns in bold:

`$GWFILE` → **`gateways` list**
`$LEGACY_GWFILE` → **Kubo's `$IPFS_PATH/gateway` file**

INTEGRATION.md
```diff
@@ -35,6 +35,39 @@ Integration here has 2 different meanings.
...
+
+Throughout this spec you might see `$CONFIG/ipfs/...`. `$CONFIG` is a mere abbreviation for the config folder, the variable doesn't exist and only serves to make this spec more readable. This section defines where `$CONFIG` should point to in a cross-platform manner.
+
+As a reference, [this](https://github.com/cjbassi/platform-dirs-rs#path-list) list of configuration folders is used.
```
lidel, line 42: We want to link to upstream standards directly.

```suggestion
Configuration is placed in folders that follow platform-specific standards:
- Linux/*BSD: [XDG Base Directories] and [XDG User Directories]
- macOS: [Standard Directories]
- Windows: [Known Folder]

[XDG Base Directories]: https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html
[XDG user directories]: https://www.freedesktop.org/wiki/Software/xdg-user-dirs/
[Known Folder]: https://msdn.microsoft.com/en-us/library/windows/desktop/dd378457.aspx
[Standard Directories]: https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html#//apple_ref/doc/uid/TP40010672-CH2-SW6
```

INTEGRATION.md
```diff
@@ -35,6 +35,39 @@ Integration here has 2 different meanings.
...
+| | with variables | full paths |
+| -------- | -------- | -------- |
+| Windows     | %APPDATA%/ipfs     | C:/Users/%USERNAME%/AppData/Roaming/ipfs     |
+| macOS     | $XDG_CONFIG_HOME/ipfs     | ~/Library/Application Support/ipfs     |
+| Linux     | $XDG_CONFIG_HOME/ipfs     | ~/.config/ipfs     |
+
+In the same manner as $CONFIG, $GATEWAY and $LEGACY_GATEWAY are names for this spec only to simplify reading.
+
+`$GWFILE` = `$CONFIG/gateway`
+`$LEGACY_GWFILE` = `~/.ipfs/gateway`
```
lidel, line 53: IPIP is defining a location of a single `gateways` file, keep it simple:

```suggestion
| | with variables | default
| -------- | -------- | -------- |
| Windows     | %APPDATA%/ipfs/gateways     | C:/Users/%USERNAME%/AppData/Roaming/ipfs/gateways     |
| macOS     | $XDG_CONFIG_HOME/ipfs/gateways     | ~/Library/Application Support/ipfs/gateways     |
| Linux (per user)     | $XDG_CONFIG_HOME/ipfs/gateways     | ~/.config/ipfs/gateways     |
| Linux (global)   | /etc/ipfs/gateways     | /etc/ipfs/gateways |
```

Note, I've added `/etc/ipfs/gateways` to account for use cases where one wants to provide config to all users.

INTEGRATION.md
```diff
@@ -35,6 +35,39 @@ Integration here has 2 different meanings.
...
+
+### Gateway file and Legacy gateway file
+#### For KUBO only
```
lidel, line 56: Move this to "Compatibility" section of standalone IPIP in `IPIP` directory.
Kubo's legacy should not pollute the modern specification we want to create here.
We want projects like ffmpeg to follow new spec. Kubo will keep doing `~/.ipfs/gateway` but it should not be something new implementations are even aware of – focus should be on `~/.config/ipfs/gateways` and `/etc/ipfs/gateways` 

Note templates:
- https://github.com/ipfs/specs/blob/main/template.md → living specification
- https://github.com/ipfs/specs/blob/main/IPIP/0000-template.md  → description of introduced changes

(it looks you used IPIP tempalte for living specification – needs clanup, but its mostly moving content between sections – lmk if you need editorial help)

INTEGRATION.md
```diff
@@ -44,34 +77,37 @@ The below decision tree defines the decisions to be made in order to choose the
       A[Has gateway argument] --> |yes| B(gateway from argument);
       A[Has gateway argument] --> |no| C(try IPFS_GATEWAY env);
```
lidel, line 78: "argument" feels focused on command-line use case, but we may have a GUI application that has custom  gateway setting in own preferences, and no argument per se  (like Brave).

Maybe "Has own gateway config" → "Gateway from own config" ?

INTEGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+
+**Maintainer(s)**:
+- [Mark Gaiser](https://github.com/markg85/)
+
+* * *
+
+## Summary
+
+Defines the decision model an IPFS enabled application should use to find a IPFS gateway. Simultaneously defines how an IPFS implementation should expose it's gateway.
+
```
autonome, line 16: s/IPFS gateway/HTTP gateways/ (note the plural)

this change throughout - should be about configuring a gateway list.

INTEGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+For clarity upon reading this document:
+**IPFS implementation**, this is an application implementing (a set of) the IPFS specifications. It would expose a gateway for other applications to use. Examples here are KUBO and Iroh.
```
autonome, line 18: s/KUBO/Kubo/

INTEGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+
+Simultaneously the spec also defines how IPFS implementations should expose their gateway.
+
+## Detailed design
+
+This integration spec defines the recommended way for a third party applications to integrate IPFS support in their application and thereby gaining easy access to resources stored on the IPFS platform.
+
+Integration here has 2 different meanings.
+
+1. The implementing application can handle IPFS resources with the `ipfs` and `ipns` protocols. As an example, the implementing application should be able to handle a url in this format `ipfs://QmbGtJg23skhvFmu9mJiePVByhfzu5rwo74MEkVDYAmF5T` which in this case would give you the big buck bunny video. Likewise a url in the `ipns` should be handled too. Making these protocols usable is left up to the specific application implementing this support.
```
autonome, line 34: s/url/address/ in this section. no locations to be found 'round here!

INTEGRATION.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Motivation
+
+Applications wanting to use IPFS resources are, without this spec, left to invent their own ways of finding a gateway. This spec defines how an application wanting to implement IPFS support can find (at least one) gateway to use.
+
+Simultaneously the spec also defines how IPFS implementations should expose their gateway.
+
+## Detailed design
+
+This integration spec defines the recommended way for a third party applications to integrate IPFS support in their application and thereby gaining easy access to resources stored on the IPFS platform.
```
autonome, line 30: s/stored/available/

ipfs is address-ability but not necessarily storage. short-caching http gateways are a great example of that.

INTEGRATION.md
```diff
@@ -44,34 +77,37 @@ The below decision tree defines the decisions to be made in order to choose the
...
 
 #### Environment variables
 
-The decision tree is influences by a couple environement variables.
+The decision tree is influenced by a couple environement variables.
 
-**`IPFS_RACING_GATEWAY`** When this environemnt variable isn't found (the default), racing gateways should be attempted upon reaching this point in the flow. When the environment variable exists it's value should be used. If it's value is `1` (or `true`) then racing gateways should be attempted. If this value is `0` (or `false`) then racing gateways are off. The control flow proceeds in the `no` branch.
+**`IPFS_RACING_GATEWAY`** When this environemnt variable isn't found (the default), racing gateways will be attempted upon reaching this point in the flow. When the environment variable exists it's value should be used. If it's value is `1` (or `true`) then racing gateways should be attempted. If this value is `0` (or `false`) then racing gateways are off. The control flow proceeds in the `no` branch.
 
 **`IPFS_FALLBACK_GATEWAY`** When this variable doesn't exist `dweb.link` will be used. When the variable does exist it's value will be used instead.
```
lidel, line 112: I don't think this belongs to IPIP defining `gateways` file.
These concerns should be moved to separate one, which defines the consumer of the `gateways` file. 

Perhaps:

- Move this spec to `./integrations/GATEWAYS_FILE.md` for now
- Open separate IPIP and add `./integrations/GATEWAYS_CLIENT.md`

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
+# IPIP 0280: Gateways file
+- Start Date: 2022-12-08
+- Related Issues:
+  - https://github.com/ipfs/kubo/issues/8847
+
+## Summary
+
+Defines how IPFS implementations must expose their gateway. This only affects the file-based method of exposiing a gateway.
```
meandavejustice, line 8: typo: exposiing

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+
+Applications wanting to use IPFS resources are, without this spec, left to invent their own ways of finding a gateway. This spec defines how an application wanting to implement IPFS support can find a list of gateways.
+
+Users should be aware that this is just a list of gateways. It could contain a local gateway but doesn't have to. It's up to the user to do filtering on that list to find an actual local gateway.
+
+## Detailed design
+
+This integration spec defines the recommended way for IPFS implementations to expose the gateway they host for IPFS integrations (think applications wanting to support IPFS) to find and use said gateways.
+
+An IPFS implementation must expose the gateway it serves in a file called `gateways` (more on where this file is located in the section below). The gateway must be exposed as a single line in the following format:
```
meandavejustice, line 20: > more on where this file is located in the section below
Should be an anchor jumping link to `#Gateways file placement`

#Gateways file placement

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+As a reference, [this](https://github.com/cjbassi/platform-dirs-rs#path-list) list of configuration folders is used.
+
+| | with variables | full paths |
+| -------- | -------- | -------- |
+| (user) Windows     | %APPDATA%/ipfs     | C:/Users/%USERNAME%/AppData/Roaming/ipfs     |
+| (global) Windows     | %PROGRAMDATA%/ipfs     | C:/ProgramData/ipfs     |
+| (user) macOS     | $XDG_CONFIG_HOME/ipfs     | ~/Library/Application Support/ipfs     |
+| (global) macOS     | N/A     | /Library/Application Support/ipfs     |
+| (user) Linux     | $XDG_CONFIG_HOME/ipfs     | ~/.config/ipfs     |
+| (global) Linux     | N/A     | /etc/ipfs     |
```
meandavejustice, line 39: Nit: missing final ` | ` on last line of table. May cause rendering issues in future

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+
+The `gateways` file must be placed in the folder appropiate for the platform the IPFS implementation instance is running on. For linux that would be `~/.config/ipfs/gateways` or `$XDG_CONFIG_HOME/ipfs/gateways`
```
meandavejustice, line 41: typo: `appropiate`

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+
+The file will be created when it doesn't exist.
+It will never be delated! This means the file will exist and be empty when an IPFS implementation removes it's own gateway from that file and if that gateway was the only line in the file.
```
meandavejustice, line 44: typo: delated

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+
+Creating and updating the `gateways` file only applies to the `(user)` prefixed file. The `(global)` prefixed file will not be written to by the IPFS implementation!
+
+#### Local file
+The local file is for read-write access for the IPFS implementation.
+
+#### Global file
```
meandavejustice, line 51: @markg85 Where does global gateways file come from? Is this an optional file created by the end user?

integrations/GATEWAYS_FILE.md
```diff
@@ -0,0 +1,91 @@
...
+
+To simplify IPFS integrations in third party applications it's important to know if said application can use a gateway. This spec defines the rules to find such a gateway and/or how to influence it.
+
+### User benefit
+
+Users, applications using IPFS, get a defined way to find gateways to use. Without this spec there is no defined way.
+
+### Compatibility
+
+This is only applicable to Kubo! Other IPFS implementations are to ignore this
```
meandavejustice, line 68: This should be styled differently, either maybe italicized or prefixed with "Note:" This was confusing to me, I first read it that the whole IPIP-280 was to be ignored outside of Kubo (which obviously it is not)

dirkf:
For Windows, it's not clear that the Roaming config (`%APPDATA%`) is suitable. You may not want your gateway setting on machine A (eg, with no local proxy) to be synced to machine B (eg, with a local proxy that has set a global gateway file). Perhaps use `%LOCALAPPDATA%`, or let it take precedence?

dirkf:
Shouldn't the spec should say what text format is used? Is the filename in the file limited to ASCII, or otherwise is the file UTF-8, or in the encoding for the system locale, the system file encoding, etc?

dirkf:
What happens if a gateway adds its URL to the global `gateways` and crashes without removing it (say, power loss)? Should there a clean-up system and/or some way to separate admin-specified defaults at the end of the file from dynamic entries at the start of the file?

src/http-gateways/gateway-detection.md
```diff
@@ -0,0 +1,102 @@
...
+   - Linux
+     1. `$HOME/.config/ipfs/gateway` (user)
+     2. `/etc/ipfs/gateway` (global)
+
+When `gateway` file is present, the file contents MUST be interpreted as line
+separated (`\n` or `\r\n`) `text/plain` file.
+
+The first line of `gateway` file MUST be a valid `http://` or `https://` URL.
+
+Implementations that only need one URL SHOULD use the first one and ignore the rest of the file.
```
lidel, line 75: @markg85 I wrote this bit naively, and  now I realized that having more than one URL will cause issues in curl ([code](https://github.com/markg85/curl/blob/22e2eaa278b7be853506bc3e81d349736a147140/src/tool_operate.c#L745-L757)). 

Any concerns with simplifying the spec and stating that only one URL is allowed in the file?

lidel (commented):
@markg85  I've rebased and split IPIP into spec and IPIP memo, integrated feedback from @dirkf  and others, and  turned this into a spec we would be sending to applications to simplify integrations like ones in ffmpeg, curl etc. 

lmk if I've removed something that is still relevant, or if you feel we should include something new.

src/http-gateways/gateway-detection.md
```diff
@@ -0,0 +1,102 @@
...
+   - Windows
+     1. `%LOCALAPPDATA%/ipfs/gateway` (local user)
+     2. `%APPDATA%/ipfs/gateway` (roaming user)
+     3. `%PROGRAMDATA%/ipfs/gateway` (global)
+   - macOS
+     1. `$HOME/Library/Application Support/ipfs/gateway` (user)
+     2. `/Library/Application Support/ipfs/gateway` (global)
+   - Linux
+     1. `$HOME/.config/ipfs/gateway` (user)
+     2. `/etc/ipfs/gateway` (global)
```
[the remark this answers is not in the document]
markg85: I'm not feeling strong about those either.
They don't hurt or complicate things. The only reason i'd lean towards keeping it is because they clarify the windows part, mac/linux is effectively handled by points 1-4 as well.

So if it doesn't hurt, keep it. Else remove it.

src/http-gateways/gateway-detection.md
markg85: Ohh ouch!

For this spec. I'd keep it at a first line requirement but not limited to 1 line. Aka, it's fine as-is. 

Curl needs some attention to catch the multiline scenario regardless of this spec. As just any other data after the first line is gonna break it.

lidel: Ok, my understanding is this problem was solved in https://github.com/curl/curl/commit/859e88f6533f9e1f890f6f39f34178871c48869e#diff-1fa338b6a8d4b9a9ffd2243527ac4951c2576cb32dc28440f36f4f59e60643e2R785, the `gateway` file can now have more than one line, curl ignores everything after the first one.

2color:
@lidel Do you still have any interest in getting this ratified?

markg85:
Ohhh, this one feels like decades ago! (just 3 years)

For what it's worth, I'm still, to this day and on a near day-to-day basis, using what is written in this spec for my own media server.

On one hand I`d say do whatever you folks think is right for this IPIP.
On the other, the description here is what's implemented in curl and ffmpeg so ratifying this would make sense from that point of view.

lidel (approved):
Thanks for ping. This spec documents something that's already working in the wild (curl, ffmpeg, kubo's `$IPFS_PATH/gateway`), which is valuable, and ratification feels like a formality:


| Project | Reads `IPFS_GATEWAY` | Reads `gateway` file | Writes `gateway` file |
|---------|---------------------|---------------------|----------------------|
| Kubo | - | - | Yes (since v0.15) |
| [curl](https://curl.se/docs/ipfs.html) | Yes | Yes (`~/.ipfs/gateway`) | - |
| [ffmpeg](https://ffmpeg.org/ffmpeg-protocols.html#ipfs) | Yes | Yes (`$IPFS_PATH/gateway` or `~/.ipfs/gateway`) | - |


I've pushed some cosmetic changes, but would like second set of eyes on this before we merge (@markg85  / @2color ?)

IIUC this spec documents existing behavior that's already shipping in curl and ffmpeg, and having it written down helps future implementers (especially when LLMs are reading and likely prioritizing specification documents for best-practices when code generating).

markg85:
Hi @lidel,

Thank you for your updates, that's much appreciated!

I just went over the text and it looks oke to me. I can't give an approve as I'm the author (not allowed in the ui) but consider this my stamp of approval :heavy_check_mark:

lidel:
Thank you. Let's keep this open over holidays in case someone wants to provide late feedback. Ok to merge in January when we are back from a break.

PR #283, opened 2022-06-03
Add HTTP Gateway Specs

lidel:
This PR adds initial HTTP gateway specs under `./http-gateways` directory. 

## :eyes:  PREVIEW 

- https://github.com/ipfs/specs/blob/feat/gateway-specs/http-gateways/README.md
  - most of the content is in https://github.com/ipfs/specs/blob/feat/gateway-specs/http-gateways/PATH_GATEWAY.md

## :bulb: Why we need this

The goal of this work is to document the _current_ behavior (based on implementation in [go-ipfs 0.13-rc1](https://github.com/ipfs/go-ipfs/releases/tag/v0.13.0-rc1)).

This is  a prerequisite that will allow us to **switch future gateway improvements to be specs-driven**:

- IPFS stewards and community members will be able to propose improvements as a diff against the specs
  - See the plan to introduce [lightweight RFC process](https://github.com/ipfs/specs/issues/286)
- Different implementations will be able to adopt changes without the need to follow and reimplement whatever go-ipfs does

## :mag_right:  Needs more eyes!

I did my best to document the most important aspects of the current state, but there is a lot to cover.  If you see something is missing, or needs rephrasing (i am not a native speaker, and bad writer in general), please comment on this PR. 

All feedback will be apprecieted

* :point_right: is something unclear or should be expanded? please comment inline on the specific section
* :point_right: is something missing entirely? please comment below
* :bulb: got an idea how to re-organize content to present this better? please comment below too!

## Next

- [x] submit PR draft
- [x] sleep
- [x] add CODEOWNER 
- [x] ping potential stakeholders for feedback
- [ ] fill gaps identified during review 
  - [x] add `Content-Range` in range responses
  - [x] subdomain gw: add `?uri=` for `registerProtocolHandler`
  - [ ] path: add CORS section to Appendix
  - [x] subdomain: add TLS notes about wildcards
  - [x] dir-index-html: note  that dir listing should only fetch root node and if Size column is needed for UnixFS, it should be Tsize from [logical format](https://ipld.io/specs/codecs/dag-pb/spec/#logical-format) + hint more details for child items in a dirr could be lazy-loaded  via `?format=dag-json` and `?format=raw` 
- [ ] (meantime) move `TODO` blocks into separate PRs, prototype and document the light RFC process
- [ ] merge after 2-3 weeks of addressing feedback

### DIFF
--- a/http-gateways/DNSLINK_GATEWAY.md
+++ b/http-gateways/DNSLINK_GATEWAY.md
@@ -55,15 +55,33 @@ Same as GET, but does not return any payload.
 
 ### `Host` (request header)
 
+Defines the [DNSLink](https://docs.ipfs.io/concepts/glossary/#dnslink) name
+to RECURSIVELY resolve into an immutable `/ipfs/{cid}/` prefix that should
+be prepended to the `path` before the final IPFS content path resolution
+is performed.
 
-Defines the DNSLink name to resolve into `/ipfs/{cid}/` prefix that should be
-prepended to the `path` before the final IPFS content path resolution is
-performed.
+Implementations MUST ensure DNSLink resolution is safe and correct:
+- each DNSLink may include an additional path segment, which MUST be preserved
+- each DNSLink may point at other DNSLink, which means there MUST be a hard
+  recursion limit (e.g. 32) and HTTP 400 Bad Request error MUST be returned
+  when the limit is reached.
 
-Example: if client sent HTTP GET request for `/sub-path` path and  `Host:
-example.com` header, and DNS at `_dnslink.example.com` has TXT record with
-value  `dnslink=/ipfs/cid1`, then the final content path is
-`/ipfs/cid1/sub-path`
+**Example: resolving an advanced DNSLink chain**
+
+To illustrate, given DNSLink records:
+
+- `_dnslink.a.example.com` TXT record: `dnslink=/ipns/b.example.net/path-b`
+- `_dnslink.b.example.net` TXT record: `dnslink=/ipfs/bafy…qy3k/path-c`
+
+HTTP client sends `GET /path-a` request with  `Host: a.example.com` header
+which recursively resolves all DNSLinks and produces the final immutable
+content path:
+
+1. `Host` header + `/path-a` → `/ipns/a.example.net/path-a`
+2. Resolving DNSlink at `a.example.net` replaces `/ipns/a.example.net` with `/ipns/b.example.net/path-b`
+3. Resolving DNSlink at `b.example.net` replaces `/ipns/b.example.net` with `/ipfs/bafy…qy3k/path-c`
+4. The immutable content path is `/ipfs/bafy…qy3k/path-c/path-b/path-a`
+
 
 # Appendix: notes for implementers
 
--- a/http-gateways/DNSLINK_GATEWAY.md
+++ b/http-gateways/DNSLINK_GATEWAY.md
@@ -1,0 +1,76 @@
+# DNSLink Gateway Specification
+
+![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+**Authors**:
+
+- Marcin Rataj ([@lidel](https://github.com/lidel))
+
+----
+
+**Abstract**
+
+DNSLink Gateway is an extension of
+[PATH_GATEWAY.md](./PATH_GATEWAY.md)
+that enables hosting a specific content path under a specific DNS name.
+
+This document describes the delta between [PATH_GATEWAY.md](./PATH_GATEWAY.md) and this gateway type.
+
+In short:
+
+- HTTP request includes a valid DNSLink name in `Host` header
+- gateway resolves DNSLink to an immutable content root identified by a CID
+- HTTP response includes the data for the CID
+- No third-party CIDs can be loaded
+
+# Table of Contents
+
+- [DNSLink Gateway Specification](#dnslink-gateway-specification)
+- [Table of Contents](#table-of-contents)
+- [HTTP API](#http-api)
+  - [`GET /[{path}][?{params}]`](#get-pathparams)
+  - [`HEAD /[{path}][?{params}]`](#head-pathparams)
+- [HTTP Request](#http-request)
+  - [Request headers](#request-headers)
+    - [`Host` (request header)](#host-request-header)
+- [Appendix: notes for implementers](#appendix-notes-for-implementers)
+  - [Leveraging DNS for content routing](#leveraging-dns-for-content-routing)
+
+# HTTP API
+
+## `GET /[{path}][?{params}]`
+
+Downloads data at specified path under the content path for DNSLink name provided in `Host` header. 
+
+- `path` – optional path to a file or a directory under the content root sent in `Host` HTTP header
+    - Example: if `Host: example.com` then the content path to resolve is `/ipns/example.com/{path}`
+
+## `HEAD /[{path}][?{params}]`
+
+Same as GET, but does not return any payload.
+
+# HTTP Request
+
+## Request headers
+
+### `Host` (request header)
+
+
+Defines the DNSLink name to resolve into `/ipfs/{cid}/` prefix that should be
+prepended to the `path` before the final IPFS content path resolution is
+performed.
+
+Example: if client sent HTTP GET request for `/sub-path` path and  `Host:
+example.com` header, and DNS at `_dnslink.example.com` has TXT record with
+value  `dnslink=/ipfs/cid1`, then the final content path is
+`/ipfs/cid1/sub-path`
+
+# Appendix: notes for implementers
+
+## Leveraging DNS for content routing
+
+- It is a good idea to publish
+  [DNSAddr](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md)
+  TXT records with known content providers for the data behind a DNSLink. IPFS
+  clients will be able to detect DNSAddr and preconnect to known content
+  providers, removing the need for expensive DHT lookup.
--- a/http-gateways/DNSLINK_GATEWAY.md
+++ b/http-gateways/DNSLINK_GATEWAY.md
@@ -18,7 +18,7 @@ This document describes the delta between [PATH_GATEWAY.md](./PATH_GATEWAY.md) a
 
 In short:
 
-- HTTP request includes a valid DNSLink name in `Host` header
+- HTTP request includes a valid [DNSLink](https://dnslink.dev/) name in `Host` header
 - gateway resolves DNSLink to an immutable content root identified by a CID
 - HTTP response includes the data for the CID
 - No third-party CIDs can be loaded
--- a/http-gateways/DNSLINK_GATEWAY.md
+++ b/http-gateways/DNSLINK_GATEWAY.md
@@ -19,9 +19,9 @@ This document describes the delta between [PATH_GATEWAY.md](./PATH_GATEWAY.md) a
 In short:
 
 - HTTP request includes a valid [DNSLink](https://dnslink.dev/) name in `Host` header
+- gateway decides if DNSlink name is allowed
 - gateway resolves DNSLink to an immutable content root identified by a CID
 - HTTP response includes the data for the CID
-- No third-party CIDs can be loaded
 
 # Table of Contents
 
--- a/http-gateways/DNSLINK_GATEWAY.md
+++ b/http-gateways/DNSLINK_GATEWAY.md
@@ -33,6 +33,7 @@ In short:
 - [HTTP Request](#http-request)
   - [Request headers](#request-headers)
     - [`Host` (request header)](#host-request-header)
+- [HTTP Response](#http-response)
 - [Appendix: notes for implementers](#appendix-notes-for-implementers)
   - [Leveraging DNS for content routing](#leveraging-dns-for-content-routing)
 
@@ -50,6 +51,8 @@ Downloads data at specified path under the content path for DNSLink name provide
 Same as GET, but does not return any payload.
 
 # HTTP Request
+
+Below MUST be implemented **in addition** to the [HTTP Request section from `PATH_GATEWAY.md`](./PATH_GATEWAY.md#http-request).
 
 ## Request headers
 
@@ -82,6 +85,9 @@ content path:
 3. Resolving DNSlink at `b.example.net` replaces `/ipns/b.example.net` with `/ipfs/bafy…qy3k/path-c`
 4. The immutable content path is `/ipfs/bafy…qy3k/path-c/path-b/path-a`
 
+# HTTP Response
+
+Same as [HTTP Response section in `PATH_GATEWAY.md`](./PATH_GATEWAY.md#http-response).
 
 # Appendix: notes for implementers
 
--- a/http-gateways/PATH_GATEWAY.md
+++ b/http-gateways/PATH_GATEWAY.md
@@ -626,6 +626,11 @@ the DAG it represents does not include content path remainder.
 - Advanced caching strategies can be built using additional information in
   `X-Ipfs-Path` and `X-Ipfs-Roots` headers.
 
+- Implement support for requests sent with
+  [`Cache-Control: only-if-cached`](#cache-control-request-header).
+  It allows IPFS-aware HTTP clients to probe and prioritize gateways that
+  already have the data cached, significantly improving retrieval speeds.
+
 ## Denylists
 
 Optional, but encouraged.
--- a/http-gateways/PATH_GATEWAY.md
+++ b/http-gateways/PATH_GATEWAY.md
@@ -1,0 +1,530 @@
+# Path Gateway Specification
+
+![](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square)
+
+**Authors**:
+
+- Marcin Rataj ([@lidel](https://github.com/lidel))
+
+----
+
+**Abstract**
+
+The most versatile form of IPFS Gateway is a Path Gateway.
+
+It exposes namespaces like  `/ipfs/` and `/ipns/`  under HTTP server root and
+provides basic primitives for integrating IPFS resources within existing HTTP
+stack.
+
+**Note:** additional Web Gateways aimed for website hosting and web browsers
+extend the below spec and are defined in
+[SUBDOMAIN_GATEWAY.md](./SUBDOMAIN_GATEWAY.md) and
+[DNSLINK_GATEWAY.md](./DNSLINK_GATEWAY.md). There is also a minimal
+[TRUSTLESS_GATEWAY.md](./TRUSTLESS_GATEWAY.md) specification for use cases
+where client prefers to perform all validation locally.
+
+# Table of Contents
+
+- [Path Gateway Specification](#path-gateway-specification)
+- [Table of Contents](#table-of-contents)
+- [HTTP API](#http-api)
+  - [`GET /ipfs/{cid}[/{path}][?{params}]`](#get-ipfscidpathparams)
+  - [`HEAD /ipfs/{cid}[/{path}][?{params}]`](#head-ipfscidpathparams)
+- [HTTP Request](#http-request)
+  - [Request Headers](#request-headers)
+    - [`If-None-Match` (request header)](#if-none-match-request-header)
+    - [`Cache-Control` (request header)](#cache-control-request-header)
+    - [`Accept` (request header)](#accept-request-header)
+    - [`Range` (request header)](#range-request-header)
+    - [`Service-Worker` (request header)](#service-worker-request-header)
+  - [Request Query Parameters](#request-query-parameters)
+    - [`filename` (request query parameter)](#filename-request-query-parameter)
+    - [`download` (request query parameter)](#download-request-query-parameter)
+    - [`format` (request query parameter)](#format-request-query-parameter)
+- [HTTP Response](#http-response)
+  - [Response Status Codes](#response-status-codes)
+    - [`200` OK](#200-ok)
+    - [`206` Partial Content](#206-partial-content)
+    - [`301` Moved Permanently](#301-moved-permanently)
+    - [`400` Bad Request](#400-bad-request)
+    - [`404` Not Found](#404-not-found)
+    - [`410` Gone](#410-gone)
+    - [`429` Too Many Requests](#429-too-many-requests)
+    - [`451` Unavailable For Legal Reasons](#451-unavailable-for-legal-reasons)
+    - [`500` Internal Server Error](#500-internal-server-error)
+    - [`504` Gateway Timeout](#504-gateway-timeout)
+  - [Response Headers](#response-headers)
+    - [`Etag` (response header)](#etag-response-header)
+    - [`Cache-Control` (response header)](#cache-control-response-header)
+    - [`Last-Modified` (response header)](#last-modified-response-header)
+    - [`Content-Type` (response header)](#content-type-response-header)
+    - [`Content-Disposition` (response header)](#content-disposition-response-header)
+    - [`Content-Length` (response header)](#content-length-response-header)
+    - [`Accept-Ranges` (response header)](#accept-ranges-response-header)
+    - [`Location` (response header)](#location-response-header)
+    - [`X-Ipfs-Path` (response header)](#x-ipfs-path-response-header)
+    - [`X-Ipfs-Roots` (response header)](#x-ipfs-roots-response-header)
+  - [Response Payload](#response-payload)
+- [Appendix: notes for implementers](#appendix-notes-for-implementers)
+  - [Content resolution](#content-resolution)
+    - [Finding the content root](#finding-the-content-root)
+    - [Traversing remaining path](#traversing-remaining-path)
+  - [Best practices for HTTP caching](#best-practices-for-http-caching)
+
+# HTTP API
+
+Path Gateway provides HTTP interface for requesting content-addressed data at
+specified content path.
+
+## `GET /ipfs/{cid}[/{path}][?{params}]`
+
+Downloads data at specified content path.
+
+- `cid` – a valid content identifier  ([CID](https://docs.ipfs.io/concepts/glossary#cid))
+- `path` – optional path remainer pointing at a file or a directory under the `cid` content root
+- `params` – optional query parameters that adjust response behavior
+
+## `HEAD /ipfs/{cid}[/{path}][?{params}]`
+
+Same as GET, but does not return any payload.
+
+# HTTP Request
+
+## Request Headers
+
+All request headers are optional.
+
+### `If-None-Match` (request header)
+
+Used for HTTP caching.
+
+Enables advanced cache control based on `Etag`,
+allowing client and server to skip data transfer if previously downloaded
+payload did not change.
+
+The Gateway MUST compare Etag values sent in `If-None-Match` with `Etag` that
+would be sent with response. Positive match MUST return HTTP status code 304
+(Not Modified), without any payload.
+
+### `Cache-Control` (request header)
+
+Used for HTTP caching.
+
+Client can send `Cache-Control: only-if-cached` to request data only if the
+gateway already has the data (e.g. in local datastore) and can return it
+immediately.
+
+If data is not cached locally, and the response requires an expensive remote
+fetch, a 504 (Gateway Timeout) status code should be returned.
+
+See [RFC7234#only-if-cached](https://datatracker.ietf.org/doc/html/rfc7234#section-5.2.1.7)
+<!-- TODO: https://github.com/ipfs/go-ipfs/issues/8783 -->
+
+### `Accept` (request header)
+
+Can be used for requesting specific response format
+
+For example:
+
+- [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) – disables IPLD/IPFS deserialization, requests a verifiable raw block to be returned
+- [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) – disables IPLD/IPFS deserialization, requests a verifiable CAR stream to be returned
+<!-- TODO: https://github.com/ipfs/go-ipfs/issues/8823
+- application/vnd.ipld.dag-json OR application/json – requests IPLD Data Model representation serialized into [DAG-JSON format](https://ipld.io/docs/codecs/known/dag-json/)
+- application/vnd.ipld.dag-cbor OR application/cbor - requests IPLD Data Model representation serialized into [DAG-CBOR format](https://ipld.io/docs/codecs/known/dag-cbor/)
+-->
+
+### `Range` (request header)
+
+`Range` can be used for requesting specific byte ranges of UnixFS files and raw
+blocks.
+
+Gateway implementations SHOULD be smart enough to require only the minimal DAG subset
+necessary for handling the range request.
+
+
+NOTE: for more advanced use cases such as partial DAG/CAR streaming, or
+non-UnixFS data structures,  see the  `selector` query parameter described
+below.
+
+### `Service-Worker` (request header)
+
+Mentioned here for security reasons and should be implemented with care.
+
+This header is sent by web browser attempting to register a service worker
+script for a specific scope. Allowing too broad scope can allow a single
+content root to take control over gateway endpoint. It is important for
+implementations to handle this correctly.
+
+Service Worker should only be allowed under specific to content roots under
+`/ipfs/{cid}/` and `/ipns/{name}/` (IMPORTANT: note the trailing slash).
+
+Gateway should refuse attempts to register a service worker for entire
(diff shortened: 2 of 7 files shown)


http-gateways/DNSLINK_GATEWAY.md
```diff
@@ -0,0 +1,76 @@
...
+# HTTP Request
+
+## Request headers
+
+### `Host` (request header)
+
+
+Defines the DNSLink name to resolve into `/ipfs/{cid}/` prefix that should be
+prepended to the `path` before the final IPFS content path resolution is
+performed.
```
olizilla, line 61: Should we define that:
 
- a dnslink that points to a `/ipns` path must be resolved to an immutable path first 
- a dnslink with a path component should be resolved to it's target CID

...path before appending the request path

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,541 @@
...
+Path Gateway decides what content to serve by taking the path from the URL
+requested and splitting it into two parts: the *CID*  and the *remainder* of
+the path.
+
+The *CID* provides the starting point, often called *content root*. The
+*remainder* of the path, if present,  will be used as instructions to traverse
+IPLD data, starting from that data which the CID identified.
+
+**Note:** Other types of gateway may allow for passing CID by other means, such
+as `Host` header, removing the need for path splitting. (See [
```
olizilla, line 522: dangling `(See [`

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,164 @@
...
+enables website hosting compatible with web browsers relative pathing and
+security model of the web. Below should be read as a delta on top of that spec.
+
+Summary:
+
+- data is requested by CID placed in `Host` header
+    - URL paths are not prefixed with `/ipfs/{cid}` or `/ipns/{foo}`
+- retrieve data from IPFS in a way that is compatible with URL-based addressing
+    - URL’s path `/` points at the content root identified by the CID
+- each CID is granted a unique [Origin sandbox](https://en.wikipedia.org/wiki/Same-origin_policy)
```
olizilla, line 23:
```suggestion
- Requests carry the CID as a sub-domain in the `Host` header rather than as a URL path prefix
  - e.g. `{cid}.ipfs.example.org` instead of `example.org/ipfs/{cid}`
- The root CID is used to define the [Resource Origin](https://en.wikipedia.org/wiki/Same-origin_policy), aligning it with the web's security model.
  - Files in a DAG may request other files within the same DAG as part of the same Origin Sandbox.
- Data is retrieved from IPFS in a way that is compatible with URL-based addressing
    - URL’s path `/` points at the content root identified by the CID

```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,164 @@
...
+form if necessary. For example:
+- `https://dweb.link/ipfs/QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR`
+  returns HTTP 301 redirect to the same CID but in case-insensitive base32:
+  - `Location: https://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.ipfs.dweb.link/`
+- `https://dweb.link/ipns/en.wikipedia-on-ipfs.org` returns HTTP 301 redirect
+  to subdomain with DNSLink name correctly inlined:
+  - `Location: https://en-wikipedia--on--ipfs-org.ipns.dweb.link/`
+
+# Appendix: notes for implementers
+
```
olizilla, line 149:
```suggestion
## DNS label limits

DNS labels, must be case-insensitive, and up to a maximum of 63 characters
[per label](https://datatracker.ietf.org/doc/html/rfc2181#section-11).
Representing CIDs within these limits can be challenging.

Base32 encoding is used for CIDs to ensure case-insensitve, URL safe 
characters are used.

Base36 is used for ED25519 libp2p keys to get the string representation to
safely fit with the 63 character limit.

How to represent CIDs with a string representation greater than 63
characters, such as those for `sha2-512` hashes, remains an
[open question](https://github.com/ipfs/go-ipfs/issues/7318).

```

olizilla (approved):
🥇 great work! A foundation for the future of IPFS <-> HTTP bridging 🌈✨🌐

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,541 @@
...
+
+### Traversing remaining path
+
+UnixFS pathing over files and directories is the implicit default used for
+resolving content paths that start with `/ipfs/` and `/ipns/`. It allows for
+traversal based on link names,  which provides a better user experience than
+low level logical pathing from IPLD:
+
```
vasco-santos, line 530: Should we recommend similar behaviour for case where given path is not valid?

http-gateways/README.md
```diff
@@ -0,0 +1,42 @@
...
+* [TRUSTLESS_GATEWAY.md](./TRUSTLESS_GATEWAY.md)
+
+## Web
+
+Special types of gateway which leverage `Host` header in addition to URL `pathname`.
+
+Designed for website hosting and improved interoperability with web browsers
+and [origin-based security
+model](https://en.wikipedia.org/wiki/Same-origin_policy).
+
```
vasco-santos, line 40: There are some layers on top of HTTP Gateways that even without being part of the spec, should probably be mentioned in a follow up section here as Implementation, given they affect gateway functionality and should be taken into account for gateway implementors. Namely, deny lists. What do you think?

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,164 @@
...
+**Abstract**
+
+Subdomain Gateway is an extension of  [PATH_GATEWAY.md](./PATH_GATEWAY.md) that
+enables website hosting compatible with web browsers relative pathing and
+security model of the web. Below should be read as a delta on top of that spec.
+
+Summary:
+
+- data is requested by CID placed in `Host` header
```
vasco-santos, line 19: Should we mention v1? we mention it in Host header details, but maybe a bit hidden

vasco-santos (commented):
This is great @lidel . Thanks for putting it together, it is super useful for everyone writing IPFS Gateways or building on top

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,541 @@
...
+### `404` Not Found
+
+Error to indicate that request was formally correct, but traversal of the
+requested content path was not possible due to a invalid or missing DAG node.
+
+### `410` Gone
+
+Error to indicate that request was formally correct, but this specific Gateway
+refuses to return requested data and is unable to provide reason why.
```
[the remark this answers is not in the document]
lanzafame: Should we be indicating what cids have not been returned because of denylists?

http-gateways/DNSLINK_GATEWAY.md
lidel: I've added  this + note about recursion limit + an example in https://github.com/ipfs/specs/commit/7be06110cf4a377adeda57a16c9b270a6287bb3e – lmk if that covers all the bases

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,541 @@
...
+
+Downloads data at specified content path.
+
+- `cid` – a valid content identifier  ([CID](https://docs.ipfs.io/concepts/glossary#cid))
+- `path` – optional path remainer pointing at a file or a directory under the `cid` content root
+- `params` – optional query parameters that adjust response behavior
+
+## `HEAD /ipfs/{cid}[/{path}][?{params}]`
+
+Same as GET, but does not return any payload.
```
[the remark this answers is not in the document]
lidel: Correct, we need to fetch at least the root blocks to generate some HTTP headers.
But there is also a separate need for inexpensive probe that does no I/O.

I've documented both in https://github.com/ipfs/specs/commit/6a0e2fc6a048b69f6f137de8f77e3540e662defe – lmk if this clears this up :pray:

http-gateways/PATH_GATEWAY.md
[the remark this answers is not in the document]
lidel: I think that is CAN and not MUST, so perhaps:

```suggestion
refuses to return requested data.

Particularly useful for implementing deny lists, in order to not serve malicious content.
The name of deny list and unique identifier of blocked entries can be provided in the response body.
```

http-gateways/PATH_GATEWAY.md
lidel:
```suggestion
as `Host` header, changing the rules behind path splitting.
(See [SUBDOMAIN_GATEWAY.md](./SUBDOMAIN_GATEWAY.md)
and [DNSLINK_GATEWAY.md](./DNSLINK_GATEWAY.md)). 
```

http-gateways/SUBDOMAIN_GATEWAY.md
lidel: Good idea, added in https://github.com/ipfs/specs/pull/283/commits/a414411c44d552e6a2d35266029083e3fd479adf

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,164 @@
...
+
+## `HEAD /[{path}][?{params}]`
+
+Same as GET, but does not return any payload.
+
+# HTTP Request
+
+## Request Query Parameters
+
+### `uri` (request query parameter)
```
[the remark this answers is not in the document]
lidel: Good idea, done in https://github.com/ipfs/specs/pull/283/commits/af0363e8703e0f9b041249c41fc050be942bab41

http-gateways/PATH_GATEWAY.md
lidel: @vasco-santos mind elaborating?

vasco-santos: Sorry for not being clear. For instance, if I provide `/ipfs/bafy...foo/this/path/does/not/exist`, this path won't exist. My suggestion here would be if we should recommend a response as part of the spec so that each gateway send same response back.

I would love to have a better response then what we get as https://dweb.link/ipfs/bafkreidyeivj7adnnac6ljvzj2e3rd5xdw3revw4da7mx2ckrstapoupoq/test

http-gateways/DNSLINK_GATEWAY.md
```diff
@@ -0,0 +1,94 @@
...
+which recursively resolves all DNSLinks and produces the final immutable
+content path:
+
+1. `Host` header + `/path-a` → `/ipns/a.example.net/path-a`
+2. Resolving DNSlink at `a.example.net` replaces `/ipns/a.example.net` with `/ipns/b.example.net/path-b`
+3. Resolving DNSlink at `b.example.net` replaces `/ipns/b.example.net` with `/ipfs/bafy…qy3k/path-c`
+4. The immutable content path is `/ipfs/bafy…qy3k/path-c/path-b/path-a`
+
+
+# Appendix: notes for implementers
```
thibmeu, line 86: An additional notes about DNSLink on the domain directly

```text
While DNSLink specification allows for the DNS entry to be made on `a.example.com` directly, it is common to place this record under a dedicated `_dnslink` entry. Most IPFS implementation resolve this subdomain first, reducing the number of queries to be made by the gateway.
```

http-gateways/README.md
```diff
@@ -0,0 +1,42 @@
...
+* [SUBDOMAIN_GATEWAY.md](./SUBDOMAIN_GATEWAY.md)
+* [DNSLINK_GATEWAY.md](./DNSLINK_GATEWAY.md)
```
thibmeu, line 42: should implementation be mentioned? This is the case on other specs.

```
# Implementations

+ [go-ipfs](httpsI//github.com/ipfs/go-ipfs)
+ [js-ipfs](https://github.com/ipfs/js-ipfs/tree/master/packages/ipfs-http-gateway/)
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,196 @@
...
+
+- Marcin Rataj ([@lidel](https://github.com/lidel))
+
+----
+
+**Abstract**
+
+Subdomain Gateway is an extension of  [PATH_GATEWAY.md](./PATH_GATEWAY.md) that
+enables website hosting compatible with web browsers relative pathing and
+security model of the web. Below should be read as a delta on top of that spec.
```
thibmeu, line 15: > compatible with web browsers relative pathing and security model of the web

This is the case for path gateway as well.

The main addition of subdomain gateway is to provide CID and IPNS names with their own environment following cross-origin isolation policy developed by web browsers

Possibly
```suggestion
Subdomain Gateway is an extension of  [PATH_GATEWAY.md](./PATH_GATEWAY.md) that
enables website hosting isolated per CID/name, while remaining compatible with web browsers
relative pathing and security model of the web
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,196 @@
...
+
+Summary:
+
+- Requests carry the CID as a sub-domain in the `Host` header rather than as a URL path prefix
+  - Case-insensitive [CIDv1](https://docs.ipfs.io/concepts/glossary/#cid-v1) encoding is used in sub-domain (see [DNS label limits](#dns-label-limits))
+  - e.g. `{cidv1}.ipfs.example.org` instead of `example.org/ipfs/{cid}`
+- The root CID is used to define the [Resource Origin](https://en.wikipedia.org/wiki/Same-origin_policy), aligning it with the web's security model.
+  - Files in a DAG may request other files within the same DAG as part of the same Origin Sandbox.
```
thibmeu, line 23:
```suggestion
  - Files in a DAG defined by the root CID may request other files within the same DAG as part of the same Origin Sandbox.
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,196 @@
...
+    - `Host: {inlined-dnslink-name}.ipns.example.net`
+        - DNSLink names include `.` which means they  MUST be inlined into a single DNS label to provide unique origin and work with wildcard TLS certificates.
+            - DNSLink label encoding:
+                - Every `-` is replaced with `--`
+                - Every `.` is replaced with `-`
+            - DNSLink label decoding
+                - Every standalone `-` is replaced with `.`
+                - Every remaining `--` is replaced with `-`
+        - Example:
+            - `/ipns/en.wikipedia-on-ipfs.org` → `Host: en-wikipedia--on--ipfs-org.ipns.dweb.link`
```
thibmeu, line 97:
```suggestion
            - `example.net/ipns/en.wikipedia-on-ipfs.org` → `Host: en-wikipedia--on--ipfs-org.ipns.example.net`
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,196 @@
...
+
+Subdomain Gateway is an extension of  [PATH_GATEWAY.md](./PATH_GATEWAY.md) that
+enables website hosting compatible with web browsers relative pathing and
+security model of the web. Below should be read as a delta on top of that spec.
+
+Summary:
+
+- Requests carry the CID as a sub-domain in the `Host` header rather than as a URL path prefix
+  - Case-insensitive [CIDv1](https://docs.ipfs.io/concepts/glossary/#cid-v1) encoding is used in sub-domain (see [DNS label limits](#dns-label-limits))
+  - e.g. `{cidv1}.ipfs.example.org` instead of `example.org/ipfs/{cid}`
```
thibmeu, line 21: nits (can be ignored): there are mention of example.org and example.net. Not sure which one should be used.

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,196 @@
...
+## Response Headers
+
+### `Location` (response header)
+
+Returned (with HTTP Status Code 301) when `Host` header does not follow the
+subdomain naming convention, but the requested URL path happens to be a valid
+`/ipfs/{cid}` or `/ipfs/{name}` content path.
+
+This additional normalization allows subdomain gateway to be used as a drop-in
+replacement compatible with regular path gateways.
```
thibmeu, line 119:
```
> Note
> During the migration from a path gateway to a subdomain gateway, even though the Location header is present, some clients may check for HTTP 200, and consider other responses as invalid. It is up to the gateway operator to clearly communicate when such a transition is to happen.
```

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,589 @@
...
+
+HTTP client can send `HEAD` request with
+[`Cache-Control: only-if-cached`](#cache-control-request-header)
+to disable IPFS data transfer and inexpensively probe if the gateway has the data cached.
+
+Implementation MUST ensure that handling `only-if-cached` `HEAD` response is
+fast and does not generate any additional I/O such as IPFS data transfer. This
+allows light clients to probe and prioritize gateways which already
+have the data.
+
```
thibmeu, line 111:
```suggestion

## `/ipns/{name}[/{path}][?{params}]`

Implementation must resolve the name to a root content, then serve is as a request to /ipfs/{resolved-content}.
`name` may refer to
+ [IPNS name](https://docs.ipfs.io/concepts/ipns/)
+ [DNSLink](https://dnslink.io)
```

http-gateways/DNSLINK_GATEWAY.md
lidel: Publishing DNSLink on the main domain  is a [deprecated, legacy way](https://dnslink.dev/#why-use-dnslink-domain-instead-of-domain). 
New implementations should only check the subdomain.  
(afaik go-ipfs will try both in parallel, if not, that is a bug).

thibmeu (approved):

Arqu:
@lidel Bit late to the party, but thank you so much for putting this together. I had the pleasure of doing a code sift to "extract" this knowledge for n0 as much as possible. 

I've been through the docs twice and it pretty much aligns with all my notes. I only got a minor bit to add on the [headers](https://github.com/ipfs/specs/blob/feat/gateway-specs/http-gateways/PATH_GATEWAY.md#x-ipfs-roots-response-header) section. Specifically that there are other headers (some deprecated eg `x-ipfs-gateway-prefix`) like `x-content-type-options` which is used, but not really covered here, though very optional. We also use `x-trace-id` as it's a commonplace header to link a trace ID to a given request. 

I'd simply mention that per spec, other `x-` headers are welcome but not part of the requirements.

I'm highlighting this mostly because there are "gotchas" and codified knowledge that will not be part of this spec, yet exists in code.

http-gateways/PATH_GATEWAY.md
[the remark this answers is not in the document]
lanzafame: This is perfect. Thanks @lidel

vasco-santos (approved):

lidel:
@Arqu thank you for proofreading and comparing notes, very encouraging to know the spec aligns with reality. 
We should document useful `X-*` headers, ok to mark them as _optional_. 

I've added `X-Content-Type-Options` in https://github.com/ipfs/specs/pull/283/commits/04111e61e98eed5cebb2d0878e7dff2ae5ed89ee, as it serves a security purpose (protecting operators of gateways that only serve Block/CAR from being used for website hosting and/or directly hotlinked and abused on other websites). 

Adding `X-Trace-Id` sounds like a good recommendation too. Implementations like [go-ipfs have internal tracing](https://github.com/ipfs/go-ipfs/blob/master/docs/environment-variables.md#tracing), and we could add the value from request to span  attributes – any docs/code I could read around this?

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,622 @@
...
+### `Content-Type` (response header)
+
+Returned with custom response formats such as `application/vnd.ipld.car` or
+`application/vnd.ipld.raw`. CAR must be returned with explicit version.
+Example: `Content-Type: application/vnd.ipld.car; version=1`
+
+When no explicit response format is provided with the request, and the
+requested data itself has no built-in content type metadata, implementations
+are free to perform content type sniffing based on filename and magic bytes to
```
BigLep, line 408: nit: link to "filename" parameter?

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,622 @@
...
+
+The remainder is an optional `filename` parameter that will be prefilled in the
+“Save As” dialog.
+
+NOTE: when the `filename` includes non-ASCII characters, the header must
+include both ASCII and UTF-8 representations for compatibility with legacy user
+agents and existing web browsers. 
+
+To illustrate, `?filename=testтест.pdf` should produce:
+`Content-Disposition inline; filename="test____.jpg"; filename*=UTF-8''test%D1%82%D0%B5%D1%81%D1%82.jpg`
```
BigLep, line 439: doublechecking that the "*" in "filename*=UTF-8" is intentional.

http-gateways/TRUSTLESS_GATEWAY.md
```diff
@@ -0,0 +1,71 @@
...
+
+### `Accept` (request header)
+
+This HTTP header is required when running in a strict, trustless mode.
+
+Gateway is free to return HTTP 400 Bad Request when running in strict trustless
+mode and  `Accept` header is missing
+
+Below response types MUST to be supported:
+- [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) – requests a singe, verifiable raw block to be returned
```
BigLep, line 62:
```suggestion
- [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) – requests a single, verifiable raw block to be returned
```

http-gateways/TRUSTLESS_GATEWAY.md
```diff
@@ -0,0 +1,71 @@
...
+- [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) – disables IPLD/IPFS deserialization, requests a verifiable CAR stream to be returned
+
+# Response
```
BigLep, line 65: Do we not want to require any of the caching headers?

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+Defines the root that should be prepended to the `path` before IPFS content
+path resolution is performed.
+
+The value in `Host` header must be a valid FQDN with at least three DNS labels:
+a case-insensitive content root identifier followed by `ipfs` or `ipns`
+namespace, and finally the domain name used by the gateway.
+
+Converting `Host` into a content path depends on the nature of requested resource:
+
+- For content at  `/ipfs/{cid}`
```
BigLep, line 82:
```suggestion
- For content at  `/ipfs/{cid}`:
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+    - `Host: {cid-mbase32}.ipfs.example.net`
+        - Example: `Host: bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi.ipfs.dweb.link`
+- For content at `/ipns/{libp2p-key}`
```
BigLep, line 85:
```suggestion
- For content at `/ipns/{libp2p-key}`:
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+    - `Host: {libp2p-key-mbase36}.ipns.example.net`
+        - Example: `Host: k2k4r8jl0yz8qjgqbmc2cdu5hkqek5rj6flgnlkyywynci20j0iuyfuj.ipns.dweb.link`
+        - Note: Base36 must be used to ensure CIDv1 with ED25519 fits in a single DNS label (63 characters).
+- For content at `/ipns/{dnslink-name}`
```
BigLep, line 89:
```suggestion
- For content at `/ipns/{dnslink-name}`:
```

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+        - DNSLink names include `.` which means they  MUST be inlined into a single DNS label to provide unique origin and work with wildcard TLS certificates.
+            - DNSLink label encoding:
+                - Every `-` is replaced with `--`
+                - Every `.` is replaced with `-`
+            - DNSLink label decoding
+                - Every standalone `-` is replaced with `.`
+                - Every remaining `--` is replaced with `-`
+        - Example:
+            - `example.net/ipns/en.wikipedia-on-ipfs.org` → `Host: en-wikipedia--on--ipfs-org.ipns.example.net`
+- For everything else (missing `Host`, or not following the above convention)
```
BigLep, line 100: Is this line complete?  What do we do in the "everything else" case?

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+
+- Subdomain gateways provide unique origin per content root, however the
+  origins still share the parent domain name used by the gateway. To fully
+  isolate websites from each other:
+    - The gateway operator should add a wildcard entry to
+      [https://publicsuffix.org](https://publicsuffix.org/)  (PSL).
+        - Example: `dweb.link` gateway [is listed on PSL](https://publicsuffix.org/list/public_suffix_list.dat) as `*.dweb.link`
+    - Web browsers with IPFS support should detect subdomain gateway (URL
+      pattern `https://{content-root-id}.ip[f|n]s.example.net`) and dynamically
+      add it to PSL.
```
BigLep, line 177: maybe provide a link to PSL so readers know what is meant.

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,208 @@
...
+
+See [URI router](#uri-router) section for usage and implementation details.
+
+# HTTP Response
+
+## Response Headers
+
+### `Location` (response header)
+
+Returned with HTTP Status Code 301 (Moved Permanently) when `Host` header does
```
BigLep, line 116: Isn't it also returned when the uri parameter is passed (so as to redirect to the right place)?

http-gateways/DNSLINK_GATEWAY.md
```diff
@@ -0,0 +1,94 @@
...
+that enables hosting a specific content path under a specific DNS name.
+
+This document describes the delta between [PATH_GATEWAY.md](./PATH_GATEWAY.md) and this gateway type.
+
+In short:
+
+- HTTP request includes a valid [DNSLink](https://dnslink.dev/) name in `Host` header
+- gateway resolves DNSLink to an immutable content root identified by a CID
+- HTTP response includes the data for the CID
+- No third-party CIDs can be loaded
```
BigLep, line 24: At least up to this point reading this document, I don't know what is meant by "third-party CID".

lidel: simplified in ce193e6

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,622 @@
...
+
+<!-- TODO: https://github.com/ipfs/go-ipfs/issues/8783 -->
+
+### `Accept` (request header)
+
+Can be used for requesting specific response format
+
+For example:
+
+- [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) – disables IPLD/IPFS deserialization, requests a verifiable raw block to be returned
```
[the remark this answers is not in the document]
lidel: yes, added in 7e65a76

http-gateways/PATH_GATEWAY.md
lidel: @vasco-santos added "Handling traversal errors" section in e3637e5 – lmk what additional info should be added in that section

http-gateways/SUBDOMAIN_GATEWAY.md
lidel: I think i forgot to mark this as TODO and never got to this.
Finished this in  25497a4

http-gateways/TRUSTLESS_GATEWAY.md
lidel: Yes, I clarified that in 4b57971 (`PATH_GATEWAY.md` is the base which every gateway type should implement, other types are delta agaist `PATH_GATEWAY.md`)

http-gateways/PATH_GATEWAY.md
```diff
@@ -563,6 +564,13 @@ Optional, present in certain response types:
   non-executable binary response types are not used in `<script>` and `<style>`
   HTML tags.
 
+### `X-Trace-Id` (response header)
+
+Optional. Implementations are free to use this header to return a globally
+unique identifier to help in debugging errors and performance issues.
+
+A good practice is to always return it with HTTP error [status codes](#response-status-codes) >=`400`.
+
```
lidel, line 573: @Arqu  fyi added this to https://github.com/ipfs/specs/pull/283  – having a suggestion of 'best practice' in official specs will simplify debugging across ecosystem

http-gateways/SUBDOMAIN_GATEWAY.md
```diff
@@ -0,0 +1,271 @@
...
+        - Note: Base36 must be used to ensure CIDv1 with ED25519 fits in a single DNS label (63 characters).
+- For content at `/ipns/{dnslink-name}`:
+    - `Host: {inlined-dnslink-name}.ipns.example.net`
+        - DNSLink names include `.` which means they  MUST be inlined into a single DNS label to provide unique origin and work with wildcard TLS certificates.
+            - DNSLink label encoding:
+                - Every `-` is replaced with `--`
+                - Every `.` is replaced with `-`
+            - DNSLink label decoding
+                - Every standalone `-` is replaced with `.`
+                - Every remaining `--` is replaced with `-`
```
RangerMauve, line 103: How does this account for `---`?

http-gateways/PATH_GATEWAY.md
```diff
@@ -0,0 +1,665 @@
...
+### `Accept-Ranges` (response header)
+
+Optional, returned to explicitly indicate if gateway supports partial HTTP
+[`Range`](#range-request-header) requests for a specific resource.
+
+For example, `Accept-Ranges: none` should be returned with
+`application/vnd.ipld.car` responses if the block order in CAR stream is not
+deterministic.
+
+### `Location` (response header)
```
RangerMauve, line 498: Would it make sense to also have a [Content-Location](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location) header in responses which would contain the canonical `ipfs://{cid}/` URL for the resource?

RangerMauve: Related to this (more details) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Location

lidel: `Content-Location` may not be the header for the job, let's discuss in https://github.com/AgregoreWeb/agregore-browser/issues/192#issuecomment-1168862909

http-gateways/SUBDOMAIN_GATEWAY.md
lidel: Odd number of `-` here  does not look like a valid encoding:

- `/ipns/foo-bar.example.com` → `foo--bar-example-com.ipns.dweb.link`
- `/ipns/foo--bar.example.com` → `foo----bar-example-com.ipns.dweb.link` 
- `/ipns/foo---bar.example.com` → `foo------bar-example-com.ipns.dweb.link`


Is someone requests `foo---bar-example-com.ipns.dweb.link` the first `--` will  be turned into single `-` producing  `foo--bar.example.com`.  Content will load only if such domain name exists and DNSLink exists.

RangerMauve: Are `foo-.bar.example.com` or `foo.-bar.example.com` valid DNS domains? Those would both translate to `foo---bar.example.com`, right?

lidel: no, `-` is not allowed at the beginning, nor end of a DNS label:

> The labels must follow the rules for ARPANET host names.  **They must start with a letter, end with a letter or digit**, and have as interior  characters only letters, digits, and hyphen.  There are also some restrictions on the length.  Labels must be 63 characters or less.
> – https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.1

lidel:
Thank you all for feedback, it made it way better than the initial version. :heart: 
I am merging it. 

If you were late to review this, please submit  fixes or improvement proposals in new PRs. 

PRs with new features such as new response types, new headers should include an Improvement Proposal document (template in #289)  that explains motivation behind proposed change. It will make it easier to review. :pray:

Issue #282, opened 2022-06-02
reframe: HTTP Caching support


lidel:
Reframe spec was added in #272, and it includes HTTP as an interoperable way of sending request-response dag-json messages.

## Problem

Reframe messages used in contexts like indexers are queries, being able to cache popular responses could significantly reduce the cost and improve performance when running an indexer or  a similar service.

Initial spec in #272 requested all requests to be over HTTP POST method.

Caching in  POST-only API will be... really tricky:
- POST is not idempotent and not "cacheable" by default, or at least not the way one would think.
- Specs are vague around this, but in practice clients and CDNs work more or less like this:
  - HTTP Caching model is around "resources", not "responses": 
  - if you set  `Cache-Control `or `Expires` headers AND set  `Content-Location`  on POST response, then response payload will be cached and returned  for any future GET requests to the `Content-Location`   – bit sketchy, and may not work this way everywhere
  - POST requests will not benefit from any preexisting cache, because they potentially mutate the resource, overriding anything that was cached. 

## Search for solution

Ideas welcome, feel free to comment below

lidel:
Below are my initial ideas how we could  get "reframe caching on HTTP":

- (A) Define custom caching based on the hash of the message in request
  - con
    - will only work in places where we control both client and server
    - if we want something that is cacheable out of the box by HTTP CDNs and web browsers -- bad luck
- (B)  There is IETF proposal to add `QUERY` HTTP method, which is idempotent: https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html  
  - pro
    - JS running in browser can send arbitrary request method, so this should work  fine (`fetch("https://cid.contact/reframe", {method: 'QUERY'})`)
    - easier to get CDNs to set up caching of QUERY than POST, [the spec states](https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html#name-caching) that cache key "MUST incorporate the request content"  
  - con
    - this is RFC draft
- (C) add support sending idempotent (or cacheable for some time) queries as a GET with reframe message in `?msg=<dag-cbor-in-multibase>` query parameter.
  - pro: standard [HTTP caching](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching) works out of the box everywhere
  - cons: there will be artificial message size limit depending on runtime

BigLep:
Per 2022-06-07 conversation: for option C:
1. we believe path arg is better than query arg
[...]

PR #287, opened 2022-06-07
reframe: cacheable GET endpoint for HTTP transport

lidel:
> Closes https://github.com/ipfs/specs/issues/282

Surgical spec tweak that aims to resolve caching problems described in https://github.com/ipfs/specs/issues/282, by providing both GET and POST endpoints (similar to json-rpc) 

- adding `GET /reframe/{mbase64url-dag-cbor}` endpoint for use cases where being able to leverage HTTP caching provides significant value
- adding _HTTP Caching Considerations_ section with
  - POST vs GET explainer
  - `Etag` for avoiding sending same response twice
  - `Last-Modified` as a mechanism for clients to force response freshness 
  
  lmk if this would do
  
@aschmahmann @guseggert  @willscott @thattommyhall @masih – this is by no means "final" – opened this PR to get the conversation started, looking for  an acceptable solution.

## TODO

- [x] <del>TBD POST should  return `Content-Location` with `GET` URL to ensure POST response is still cacheable, under correct GET URL</del>
  - probably better to skip - in most cases this introduces more cost (dual serialization to dag-json and dag-cbort in GET URL) than we get in return from opportunistic cache hits
- [x] split transports to separate REFRAME_TRANSPORTS.md  (and methods like "IPFS PUT" to  REFRAME_METHODS.md ?)
  - split proposed in https://github.com/ipfs/specs/pull/292
- [x] make it clear that we require support for both GET and POST to be implemented   – leaving it to implementation to decide when to use which  based on Reframe method inside of message sent

### DIFF
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -54,7 +54,7 @@ Requests MUST be sent as either:
   - Ephemeral HTTP `POST` request with message passed as DAG-JSON in HTTP request body
   - Suitable for bigger messages, and when HTTP caching should be skipped for the most fresh results
 
-Severs MUST support `GET` for methods marked as cacheable and MUST support `POST` for methods marked as non-cacheable.
+Severs MUST support `GET` for methods marked as cacheable and MUST support `POST` for all methods (both cacheable and not-cacheable). This allows clients to use `POST` as a fallback in case there is a technical problem with bigger Reframe messages not fitting in a `GET` URL.
 
 If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
 
@@ -156,24 +156,19 @@ Every message except the Error type should end in Request/Response.
 
 #### Cachable/Non-Cachable Methods
 
-The following methods are cachable:
+The following method responses are _cachable_:
 
 ```ipldsch
-type Request union {
+type CachableRequest union {
     | "IdentifyRequest" IdentifyRequest
     | "FindProvidersRequest" FindProvidersRequest
     | "GetIPNSRequest" GetIPNSRequest
 }
 ```
 
-The following methods are non-cachable:
+Methods that are not listed above are considered _non-cachable_.
 
-```ipldsch
-type Request union {
-    | "PutIPNSRequest" PutIPNSRequest
-}
-```
-
+Implementations are encouraged to improve performance of  `CachableRequest` methods by applying transport and method-specific caching strategies.
 #### Error
 
 The Error message type should be used in Responses to indicate that an error has occurred.
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -40,10 +40,17 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.reframe+dag-json; version=1`
+All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.reframe+dag-json; version=1`
 
+Requests MUST be sent as either:
+- `GET /reframe/{mbase64url-dag-cbor}`
+  - Cacheable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
+  - Suitable for sharing links, sending smaller messages, and when a query result MUST benefit from HTTP caching (see _HTTP Caching Considerations_ below).
+- `POST /reframe`
+  - Ephemeral HTTP `POST` request with message passed as DAG-JSON in HTTP request body
+  - Suitable for bigger messages, and when HTTP caching should be skipped for the most fresh results
 
-Requests MUST be sent as HTTP POST requests. If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
+If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
 
 Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time.
 
@@ -51,8 +58,38 @@ If a server chooses to respond to a single request message with a group of messa
 
 Requests and responses MUST come with `version=1` as a _Required Parameter_  in the `Accept` and `Content-Type` HTTP headers.
 
-
 Note: This version header is what allows the transport to more easily evolve over time (e.g. if it was desired to change the transport to support other encodings than DAG-JSON, utilize headers differently, move the request data from the body, etc.). Not including the version number is may lead to incompatibility with future versions of the transport.
+
+#### HTTP Caching Considerations
+
+**POST vs GET**
+
+HTTP `POST` requests do not benefit from any preexisting HTTP caching because
+every `POST` response will overwrite the cached resource.
+
+While it is possible to write custom middleware to cache `POST` responses based on
+request body, this is not a standard behavior and is discouraged.
+
+Use of `GET` endpoint is not mandatory, but suggested if a Reframe deployment
+expects to handle the same message query multiple times, and want to leverage
+existing HTTP tooling to maximize HTTP cache hits.
+
+**Avoiding sending the same response messages twice**
+
+Implementations MUST always return strong
+[`Etag`](https://httpwg.org/specs/rfc7232.html#header.etag) HTTP header based
+on digest of DAG-JSON response messages. This allows clients to send
+inexpensive conditional requests with
+[`If-None-Match`](https://httpwg.org/specs/rfc7232.html#header.if-none-match)
+header, which will skip when the response message did not change.
+
+**Client controls for time-based caching**
+
+Implementations MUST always return
+[`Last-Modified`](https://httpwg.org/specs/rfc7232.html#header.last-modified)
+HTTP header, allowing clients to send conditional requests with
+[`If-Modified-Since`](https://httpwg.org/specs/rfc7232.html#header.if-modified-since)
+header to specify their acceptance for stale (cached) responses.
 
 ## Protocol Message Overview
 
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -97,6 +97,23 @@ HTTP header, allowing clients to send conditional requests with
 [`If-Modified-Since`](https://httpwg.org/specs/rfc7232.html#header.if-modified-since)
 header to specify their acceptance for stale (cached) responses.
 
+**Rate-limiting non-cachable POST requests**
+
+HTTP endpoint can return status code
+[429 Too Many Requests](https://www.rfc-editor.org/rfc/rfc6585#section-4)
+with `Retry-After` header to throttle the number of `POST` requests a client can send.
+
+The body returned with `429` response should suggest use of HTTP `GET` endpoint
+for cacheable Reframe methods:
+
+```
+HTTP/1.1 429 Too Many Requests
+Content-Type: text/plain
+Retry-After: 3600
+
+too many POST requests: consider switching to cachable GET or try again later (see Retry-After header)
+```
+
 ## Protocol Message Overview
 
 We can represent each message as an IPLD Schema to denote its abstract representation independent of the serialization scheme. 
--- a/REFRAME.md
+++ b/REFRAME.md
@@ -36,6 +36,10 @@ Given that a client C wants to request information from some server S:
 1. C opens sends a request to S that is a single message
 2. S will respond to C with either a single message or a group of messages
 
+## Cachability
+
+Some methods are prone to being cachable while others are not. Methods are tagged within the spec as cachable/not cachable. Transports may leverage that information to decide if cachable/not cachable methods should be treated differently.
+
 ## Transports
 
 ### HTTP + DAG-JSON
@@ -49,6 +53,8 @@ Requests MUST be sent as either:
 - `POST /reframe`
   - Ephemeral HTTP `POST` request with message passed as DAG-JSON in HTTP request body
   - Suitable for bigger messages, and when HTTP caching should be skipped for the most fresh results
+
+Severs MUST support `GET` for methods marked as cacheable and MUST support `POST` for methods marked as non-cacheable.
 
 If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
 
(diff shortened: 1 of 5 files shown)


REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
+All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
 
+Requests MUST be sent as either:
+- `GET /reframe/{mbase64url-dag-cbor}`
+  - Cacheable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
```
willscott, line 47: i wonder if base64 of dag-cbor of the query ends up potentially too long (~2k limit iirc) for query paths

REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
...
 Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time.
 
 If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of `\n` delimited DAG-JSON messages (i.e. `{Response1}\n{Response2}...`).
 
 Requests and responses MUST come with `version=1` as a _Required Parameter_  in the `Accept` and `Content-Type` HTTP headers.
 
-
 Note: This version header is what allows the transport to more easily evolve over time (e.g. if it was desired to change the transport to support other encodings than DAG-JSON, utilize headers differently, move the request data from the body, etc.). Not including the version number is may lead to incompatibility with future versions of the transport.
 
+#### HTTP Caching Considerations
```
willscott, line 63: should we have a discussion of expectations around `Cache-Control` header, like if `max-age` should be set?

willscott (approved):
This seems reasonable to me as described

REFRAME.md
aschmahmann: Correct, this is part of the problem mentioned in option C https://github.com/ipfs/specs/issues/282#issuecomment-1145000892. It probably depends on the request method and what goes into it. This is one of the reasons it might be nice to just have servers handle proper caching of POST requests on their own.

REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
 
 ### HTTP + DAG-JSON
 
-All messages MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
+All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
 
+Requests MUST be sent as either:
```
aschmahmann, line 45: How does a client know which kind of request it should send? Does it do trial and error per-method to see what it supports (e.g. try GET and if that fails then try POST)?

REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
...
+
+**POST vs GET**
+
+HTTP `POST` requests do not benefit from any preexisting HTTP caching because
+every `POST` response will overwrite the cached resource.
+
+While it is possible to write custom middleware to cache `POST` responses based on
+request body, this is not a standard behavior and is discouraged.
+
+Use of `GET` endpoint is not mandatory, but suggested if a Reframe deployment
```
aschmahmann, line 73: Does this mean `POST` support is mandatory?

REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
...
+expects to handle the same message query multiple times, and want to leverage
+existing HTTP tooling to maximize HTTP cache hits.
+
+**Avoiding sending the same response messages twice**
```
aschmahmann, line 77: I assume this is defined per method and not collectively right? Since some methods (e.g. the proposed #285 or the IPNS Put method) need to actually make their way to the endpoint for processing and response and will likely just respond with something like `OK`.

If so do we need to start marking the various reframe methods as get-like vs modify-like to help implementers know what to do here?

REFRAME.md
```diff
@@ -40,20 +40,57 @@ Given that a client C wants to request information from some server S:
...
 If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of `\n` delimited DAG-JSON messages (i.e. `{Response1}\n{Response2}...`).
 
 Requests and responses MUST come with `version=1` as a _Required Parameter_  in the `Accept` and `Content-Type` HTTP headers.
 
-
 Note: This version header is what allows the transport to more easily evolve over time (e.g. if it was desired to change the transport to support other encodings than DAG-JSON, utilize headers differently, move the request data from the body, etc.). Not including the version number is may lead to incompatibility with future versions of the transport.
 
+#### HTTP Caching Considerations
+
+**POST vs GET**
```
aschmahmann, line 65: @lidel I think you had some thoughts here we didn't get to cover synchronously. Could you help me understand what the user model should reasonably be if someone has a browser app that's able to utilize some Reframe process running on localhost rather than remotely?

If the methods come in via `POST` all seems fine, but if they can come in via `GET` then there's potential for abuse (e.g. a bad actor hurting your reputation by using your ID to provide many bogus provider records) due to browser CORS limitations.

The alternative here seems to be having the user put their auth tokens in the browser page and having that authenticate requests. Is the idea that this isn't realistically more burdensome than dealing with CORS and so this isn't a concern?

aschmahmann (changes requested):

REFRAME.md
lidel: This spec could propose some "safe default"  for `Cache-Control` but don't know what it would be.

My understanding is that this Reframe spec defines a generic RPC, so it is difficult to reason how long a stale (cached) response  is valid/useful – no idea what `max-age` makes sense when RPC messages are opaque.

I am open to suggestions.

Proposed `Last-Modified` and `If-Modified-Since` partially solve the problem: since the client knows what was in the request message,  it can make a caching decision based on message type (do I need a response  that is not older than 5 minutes, or is it ok to have a response generated a week ago?).

REFRAME.md
lidel: @aschmahmann 
I think the high level insight is that we should not use CORS and POST as access control mechanism for general-purpose RPC.  This applies to both localhost and public endpoints.

If you run reframe endpoint that does sensitive things, then we need either:
- HTTPS endpoint  to have auth (require `Authorization` header in requests, reject if missing)
- DAG-JSON messages to have auth (require signature/token in RPC message, reject if missing)

REFRAME.md
lidel: My intention was something along the lines:
- server: MUST support both `POST` and `GET` 
- client: can support either

REFRAME.md
petar: Are we trying to keep HTTP-specific from non-HTTP-specific features of the protocol separate? If we plan to use reframe over other transports (like CBOR-over-libp2p) we should. If this is the case, then non-HTTP-specific authentication would be preferred.

lidel: This is an open question, with huge caching implications: 
- if we include auth inside a reframe message, it will change the `GET /reframe/{inlined-message}` path (because message is different for each user, or even per request if we include a once), making it impossible to cache responses for "reads"

REFRAME.md
```diff
@@ -94,6 +137,26 @@ type Response union {
...
+```
+
+The following methods are non-cachable:
+
+```ipldsch
+type Request union {
+    | "PutIPNSRequest" PutIPNSRequest
+}
+```
+
```
aschmahmann, line 159: @lidel I added this to describe the indicators for transports on cacheable/non-cacheable to support HTTP. I could tag the individual methods, but this seemed perhaps easier to look at. Happy to tag the methods individually though.

REFRAME.md
```diff
@@ -36,24 +36,67 @@ Given that a client C wants to request information from some server S:
...
+Requests MUST be sent as either:
+- `GET /reframe/{mbase64url-dag-cbor}`
+  - Cacheable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
+  - Suitable for sharing links, sending smaller messages, and when a query result MUST benefit from HTTP caching (see _HTTP Caching Considerations_ below).
+- `POST /reframe`
+  - Ephemeral HTTP `POST` request with message passed as DAG-JSON in HTTP request body
+  - Suitable for bigger messages, and when HTTP caching should be skipped for the most fresh results
 
-Requests MUST be sent as HTTP POST requests. If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
+Severs MUST support `GET` for methods marked as cacheable and MUST support `POST` for methods marked as non-cacheable.
```
aschmahmann, line 57: @lidel I added this to deal with https://github.com/ipfs/specs/pull/287/files#r891805528 and because I think this is what we talked about https://github.com/ipfs/specs/pull/287/files#r899107207.

lmk if you disagree

lidel (commented):
We should split Reframe specs to make clear separation between protocol, currently fleshed out methods, and currently  fleshed out transports.

I'm thinking:

- `reframe/REFRAME_PROTOCOL.md` - description of the generic RPC protocol
- `reframe/REFRAME_METHODS.md` - currently 'standardized' methods
- `reframe/REFRAME_TRANSPORTS.md` - list of available transports 

This way we have clear separation of concerns and specs don't leak transport abstractions upwards.

@aschmahmann lmk if it is ok for me to apply these (and below) changes

REFRAME.md
aschmahmann: I think this is fine since the MUST requirement for cacheable GETs means clients won't have to guess which to use, but they'll have a fallback.

@willscott any objections to also supporting POST here?

REFRAME.md
aschmahmann: > Implementations are encouraged to improve performance of  `CachableRequest` methods by applying transport-specific caching strategies.

Transport and method-specific caching strategies, right?

REFRAME.md
willscott: should the server be able to choose to respond with 'method not supported' to posts when it is under heavy load and needs to offload to a cache. 

are we allowed to mitigate clients that don't attempt a GET and go directly to a POST?

lidel: @willscott  Yes, HTTP service can decide to refuse.. service ;-) but there are better ways.

HTTP servers can return  [429 Too Many Requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/429) and [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) to limit use of expensive POST while encouraging GET for cachable Reframe methods.

lidel: I've added a section about rate-limiting [here](https://github.com/ipfs/specs/pull/287/commits/62eb11e7e93c5972665c4bf3c38cbedac41e9ea8?short_path=08fb031#diff-08fb031656230573a5a243374d2f35a25fc604ca11aa36b96c27b1f96aa3f9be)

REFRAME.md
lidel: (we added list of cachable Reframe methods (that should default to GET) + rate-limiting of POSTs on HTTP layer (`429 Too Many Requests`, which will suggest to use GET for these highly cachable responses)

lidel:
Proposed split in https://github.com/ipfs/specs/pull/292 (PR against this one for easier review, same content)

aschmahmann (approved):
LGTM.

@petar could you take a look at the changes here to get an understanding of the implementation complexity in Edelweiss + go-delegated-routing to support the changes here (i.e. supporting POST for all requests and GET for some, as well as being able to mark some of the requests as cacheable so we know they can use GET)?

lidel:
@petar the implementation gist is for reframe "server" to support responding to both GET and POST,  and for "client" to send GET for cachable Reframe methods to leverage HTTP Caching where possible.

petar:
@aschmahmann @lidel I will follow up here. I'm traveling the next two days and have food poisoning on top. So I'll chime in when i'm back.

petar (approved):

reframe/README.md
```diff
@@ -0,0 +1,5 @@
+# Reframe Specifications
+
+- [`REFRAME_PROTOCOL.md`](./REFRAME_PROTOCOL.md) ← start here
+  - [`REFRAME_KNOWN_METHODS.md`](./REFRAME_KNOWN_METHODS.md)
```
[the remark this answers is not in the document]
lidel: Good question. Document had "Known methods" section already, so I assumed it is similar terminology as things like  https://www.w3.org/TR/did-spec-registries/#did-methods (anyone can come up with own reframe method, just like anyone can come up with DID method). 

API is super vague.  Would   `REFRAME_MESSAGE_TYPES.md` be better?

reframe/REFRAME_HTTP_TRANSPORT.md
```diff
@@ -0,0 +1,103 @@
...
+
+Requests MUST be sent as either:
+- `GET /reframe/{mbase64url-dag-cbor}`
+  - Cachable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
+  - Suitable for sharing links, sending smaller messages, and when a query result MUST benefit from HTTP caching (see _HTTP Caching Considerations_ below).
+- `POST /reframe`
+  - Ephemeral HTTP `POST` request with message passed as DAG-JSON in HTTP request body
+  - Suitable for bigger messages, and when HTTP caching should be skipped for the most fresh results
+
+Severs MUST support `GET` for methods marked as cachable and MUST support `POST` for all methods (both cachable and not-cachable). This allows servers to rate-limit `POST` when cachable `GET` could be used instead, and enables clients to use `POST` as a fallback in case there is a technical problem with bigger Reframe messages not fitting in a `GET` URL. See "Caching Considerations" section.
```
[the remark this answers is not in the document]
lidel:
```suggestion
Servers MUST support `GET` for methods marked as cachable and MUST support `POST` for all methods (both cachable and not-cachable). This allows servers to rate-limit `POST` when cachable `GET` could be used instead, and enables clients to use `POST` as a fallback in case there is a technical problem with bigger Reframe messages not fitting in a `GET` URL. See "Caching Considerations" section.

```

willscott:
is this spec PR okay to merge? i'd like to rebase https://github.com/ipfs/specs/pull/285 into the reframe directory this makes, but it seems like it might be easier to wait for this to complete first if it isn't expected to take too much longer

lidel:
@aschmahmann @petar 
I believe it is ok to merge:  if anything changes, that will be editorial or adding more clarification and features, and not changing the underlying GET/POST approach.

We have sync tomorrow when US folks are back, let's wait until then.

REFRAME.md
lidel: Honestly, DAG-JSON, as PeerIDs, Multiaddrs are not in their text representation, but `bytes` anyway..:

```json
{"FindProvidersRequest":{"Key":{"/":"bafybeigvgzoolc3drupxhlevdp2ugqcrbcsqfmcek2zxiw5wctk3xjpjwy"}}}

{"FindProvidersResponse":{"Providers":[{"Node":{"peer":{"ID":{"/":{"bytes":"EiAngCqwSSL46hQ5+DWaJsZ1SPV2RwrqwID/OEuj5Rdgqw"}},"Multiaddresses":[{"/":{"bytes":"NiJwZWVyLmlwZnMtZWxhc3RpYy1wcm92aWRlci1hd3MuY29tBgu43QM"}}]}},"Proto":[{"2304":{}}]}]}}
```

REFRAME.md
willscott: this seems reasonable yeah. Figuring out some etag, like the hash/cid of the contents, seems like something that should basically always be possible.

Issue #257, opened 2021-11-18
Manifest for unixfs loaded via HTTP Gateway


lidel:
This is a placeholder issue for creating a spec for "content root's manifest file" which would be part of the data  published on IPFS (eg.  `/ipfs/CID/_something`) that provides hints to HTTP Gateway around things like redirects, content types and custom HTTP headers.

Related reading:
- [Proposal: manifest file for the IPFS gateway](https://github.com/ipfs/go-ipfs/issues/6214)
- [Proposal: native redirects config](https://github.com/ipfs/go-ipfs/issues/7392)

Feel free to discuss/post ideas below, or open PRs with spec drafts that could be reviewed independently.

BigLep:
Related are notes from the 2021-11-16 IPFS operators meeting: https://hackmd.io/MiYYUCR-RWqhAKfJ_B15wA

justincjohnson:
Hi @lidel.  I started working at Fission this year and have recently been tasked with implementing redirects support in go-ipfs.  @cbrake started working on this [in this PR](https://github.com/fission-suite/go-ipfs/pull/4) 🙏 and I plan on continuing the work in [this PR](https://github.com/ipfs/go-ipfs/pull/8816).

If you don't mind, I'd like to make sure I understand the process here as far as specs and issue management go so I don't cause any annoying friction.

From the issues I've read through it appears that it was decided to split out redirects support from manifests support when Cloudflare announced Cloudflare Pages with `_redirects` support, similar to Netlify's `_redirects` support.  This issue I'm commenting on appears to be specific to spec'ing out manifest support.  Should redirects support get its own spec and therefore spec issue or were you thinking it would be included in the same spec?

Regardless of whether or not manifests and redirects are separate specs, my assumption is that the redirects portion of the spec will both inform and be informed by the go-ipfs redirects support tentative implementation.  Does that sound right?  Is there any preferred order to tackling this (spec vs implementation first)?  I'm currently assuming I should start with essentially implementing what Cloudflare supports and see what feedback others provide, adjust for feedback, and then eventually work backward from there into text for the spec.

Assuming I'm on the right track, I was planning on submitting 1) an issue in go-ipfs to track this implementation work vs just my draft PR, and 2) possibly an issue in this specs repo to track the redirects portion of the spec, unless it should all be included in this spec issue.

[...]

PR #290, opened 2022-06-15
IPIP: Gateway _redirects File

justincjohnson:
> This is part of https://github.com/ipfs/specs/issues/257, working code in https://github.com/ipfs/kubo/pull/8890

👀 Preview
- [IPIP - 0002-gateway-redirects.md](https://github.com/justincjohnson/specs/blob/feat/gateway-redirects/IPIP/0002-gateway-redirects.md)
- [Spec - REDIRECTS_FILE.md](https://github.com/justincjohnson/specs/blob/feat/gateway-redirects/http-gateways/REDIRECTS_FILE.md)
- [HTTP Gateways - README.md](https://github.com/justincjohnson/specs/blob/feat/gateway-redirects/http-gateways/README.md)

For reference, working code in Kubo PR: https://github.com/ipfs/kubo/pull/8890


RFC/0000-gateway-redirects.md
```diff
@@ -0,0 +1,99 @@
...
+
+Rules in the file are evaluated top to bottom.
+
+For performance reasons this proposal does not include forced redirect support (i.e. redirect rules that are evaluated even if the `from` path exists).  This means that no redirect logic will be executed unless the requested path doesn't exist.
+
+If a `_redirects` file has invalid logic, perhaps not even parsing correctly, we may want to show the errors to the user viewing the site via the Subdomain Gateway.
+
+## Test fixtures
+
+A sample site be found in QmaiAcL7pFedPJXxNJNDVDTUR78We7yBhdLzg151ZMzLCv.
```
justincjohnson, line 38: From CAR file in my draft PR implementing redirect support.

http-gateways/REDIRECTS_FILE.md
```diff
@@ -0,0 +1,109 @@
...
+The URL or path to redirect to.
+
+## Status
+
+An optional integer specifying the HTTP status code to return from the request.  Supported values are:
+
+- `301` - Permanent Redirect (default)
+- `302` - Temporary Redirect
+- `404` - Not Found
+  - Useful for redirecting invalid URLs to a pretty 404 page.
```
lidel, line 63: People using DNSLink/IPNS should be able to respond to takedown notices etc:
```suggestion
  - Useful for redirecting invalid URLs to a pretty 404 page.
- `410` - Gone
- `451` - Unavailable For Legal Reasons  
```

RFC/0000-gateway-redirects.md
```diff
@@ -0,0 +1,102 @@
...
+- `to` - specifies the path or URL to redirect to (can include placeholders or splat matched in `from`)
+- `status` - optional HTTP status code (301 if not specified)
+
+Rules in the file are evaluated top to bottom.
+
+For performance reasons this proposal does not include forced redirect support (i.e. redirect rules that are evaluated even if the `from` path exists). In other word, redirect logic will be evaluated if and only if the requested path does not exist.  If the requested path exists, we won't even check for the existence of the `_redirects` file.
+
+If a `_redirects` file exists but is unable to be processed, perhaps not even parsing correctly, errors will be returned to the user viewing the site via the Gateway.
+
+## Test fixtures
```
lidel, line 36: Move this section to `REDIRECTS_FILE.md` and only list CID that should be pinned in IPIP document.

RFC/0000-gateway-redirects.md
```diff
@@ -0,0 +1,102 @@
...
+- `from` - specifies the path to intercept (can include placeholders and a trailing splat)
+- `to` - specifies the path or URL to redirect to (can include placeholders or splat matched in `from`)
+- `status` - optional HTTP status code (301 if not specified)
+
+Rules in the file are evaluated top to bottom.
+
+For performance reasons this proposal does not include forced redirect support (i.e. redirect rules that are evaluated even if the `from` path exists). In other word, redirect logic will be evaluated if and only if the requested path does not exist.  If the requested path exists, we won't even check for the existence of the `_redirects` file.
+
+If a `_redirects` file exists but is unable to be processed, perhaps not even parsing correctly, errors will be returned to the user viewing the site via the Gateway.
+
```
lidel, line 35:
```suggestion

The detailed specification is added in `http-gateways/REDIRECTS_FILE.md`.

```

RFC/0000-gateway-redirects.md
```diff
@@ -0,0 +1,102 @@
...
+Provides general URL redirect and rewrite support.
+The primary benefit is that developers will be able to host single-page applications in IPFS.
+Pretty 404 pages will likely be the next most common use of this feature.
+
+### Compatibility
+
+If by some chance developers are already hosting sites that contain a `_redirects` file that does something else, they may need to update the contents of the file to match the new functionality. Errors returned to the user due to parsing errors will guide them regarding the required updates.
+
+### Security
```
lidel, line 89: Move to spec, and reference here.

lidel (changes requested):
Thank you for filling this!

Will circle back to this when we land gateway specs and IPIP process, but made a quick pass from IPIP perspective (figuring out what works, what not, and what we can simplify),  so thank you for your patience.

Left some comments inline (but feel free to ignore/park until final IPIP template lands)

justincjohnson:
@lidel out of curiosity, did you mean to close this?

lidel:
@justincjohnson  no, I never closed this, thanks for the ping. 

It  looks like a bug in Github – your PR should be switched to `main` after I merged #290 , but instead it closed this PR. I'll reopen and fix it manually, sorry for the noise. 

ps. I'll be going  over IPIP reviews  next week.

RFC/0000-gateway-redirects.md
justincjohnson: I moved this, but will await further feedback. 👍

RFC/0000-gateway-redirects.md
justincjohnson: Done, but will await further feedback. 👍

lidel (changes requested):
Thank you for your angelic patience @justincjohnson  :heart: 

I did the first pass review of this along with your working code in https://github.com/ipfs/kubo/pull/8890 – some questions / asks apply to both – details inline.

http-gateways/REDIRECTS_FILE.md
```diff
@@ -0,0 +1,155 @@
...
+The `_redirects` file is as follows.
+
+```
+$ ipfs cat /ipfs/QmaiAcL7pFedPJXxNJNDVDTUR78We7yBhdLzg151ZMzLCv/_redirects
+/redirect-one /one.html
+/301-redirect-one /one.html 301
+/302-redirect-two /two.html 302
+/200-index /index.html 200
+/posts/:year/:month/:day/:title /articles/:year/:month/:day/:title 301
+/splat/:splat /redirected-splat/:splat 301
```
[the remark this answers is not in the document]
justincjohnson: Good find.  Yes, I meant `/splat/*` and will confirm that the CAR with fixtures is correct.

IPIP/0000-gateway-redirects.md
```diff
@@ -0,0 +1,75 @@
...
+
+The format for the file is `from to [status]`.
+
+- `from` - specifies the path to intercept (can include placeholders and a trailing splat)
+- `to` - specifies the path or URL to redirect to (can include placeholders or splat matched in `from`)
+- `status` - optional HTTP status code (301 if not specified)
+
+Rules in the file are evaluated top to bottom.
+
+For performance reasons this proposal does not include forced redirect support (i.e. redirect rules that are evaluated even if the `from` path exists). In other word, redirect logic will be evaluated if and only if the requested path does not exist.  If the requested path exists, we won't even check for the existence of the `_redirects` file.
```
thibmeu, line 34: Side comment: `kubo` supports [custom 404 pages](https://github.com/ipfs/kubo/commit/dfceafdbd3debca60959bdb9ecb28dd4c431c163) already. It does not appear to be in the HTTP gateway spec.
Is this in the scope of this IPIP to provide backward compatibility to this unspecified behavior?

cc @lidel

justincjohnson: Per a conversation with @lidel, that pretty 4045 functionality isn't officially documented, so we're not including it in the spec.  I'm still falling back to that logic in the code but we plan on removing that code sometime down the road.

justincjohnson: Ah, and now there's https://github.com/ipfs/ipfs-docs/issues/1155.

IPIP/0000-gateway-redirects.md
```diff
@@ -0,0 +1,75 @@
...
+
+If a `_redirects` file exists but is unable to be processed, perhaps not even parsing correctly, errors will be returned to the user viewing the site via the Gateway.
+
+The detailed specification is added in `http-gateways/REDIRECTS_FILE.md`.
+
+### Test fixtures
+QmaiAcL7pFedPJXxNJNDVDTUR78We7yBhdLzg151ZMzLCv
```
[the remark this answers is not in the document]
justincjohnson: Here's the car file, updated with the new CID with the splat fix.

[redirects.car.zip](https://github.com/ipfs/specs/files/9309443/redirects.car.zip)

http-gateways/REDIRECTS_FILE.md
```diff
@@ -0,0 +1,155 @@
...
+## To
+
+The URL or path to redirect to.
+
+## Status
+
+An optional integer specifying the HTTP status code to return from the request.  Supported values are:
+
+- `301` - Permanent Redirect (default)
+- `302` - Temporary Redirect
```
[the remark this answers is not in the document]
justincjohnson: @lidel I'm not sure this makes sense.  If I understand correctly, these additional methods assume we will be receiving requests with PUT and POST and then redirect rules can redirect those requests to a separate dynamic backend for processing, using the same PUT or POST method.  The problem I see with this is the gateway currently isn't handling PUT and POST requests.  Any requests with PUT or POST result in an error that the gateway is read-only.  So any redirect rules that are processed will only be for GET or HEAD requests.

Are you wanting these changes in anticipation of future rearchitecting of the gateway to support writes?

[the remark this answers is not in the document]
lidel: @justincjohnson no, this is purely to maximize interop with HTTP semantics.

Supporting these codes makes it easier for people who have a regular web2 website to move hosting of that website to IPFS. They can redirect legacy POST endpoints to a different hostname (307, 308), or turn them to GET (303) and show user-friendly error page.  

By making the migration less painful, we allow for migrating bigger websites in stages.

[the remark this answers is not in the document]
lidel: You are correct, but the behavior you described is a bug/limitation in Kubo implementation, not a real technical block for having them in this spec.

Existing `_redirects` implementations have different behaviors already:
- [Cloudflare Pages docs](https://developers.cloudflare.com/pages/platform/redirects) state they support 301, 302, 303, 307, 308
- [Netlify](https://docs.netlify.com/routing/redirects/redirect-options/#http-status-codes) only supports 301 and 302, explicitly does not support 307, does not mention others


My suggestion is to include all redirect codes in the spec for completeness, and fix Kubo at some point:
- the specification for "redirects file" should support all HTTP redirect types.
- implementations are free to be more pragmatic and implement only a subset of the spec that serves their user base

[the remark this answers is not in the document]
justincjohnson: Got it.  That makes sense to me.  👌

lidel (commented):
Big fan of comprehensive text fixtures, that implementers can use in their tests, thank you!
I think it is ready for final reviews. 

We were missing some folks at IPFS Implementers Sync call today, so pinging here:

- @lidel will drop :+1:  after the review https://github.com/ipfs/kubo/pull/8890
- @dignifiedquire will you have time to eyeball this from [Iroh](https://github.com/n0-computer/iroh)'s point of view?
- @thattommyhall  / @gmasgras  / @alanshaw / @vasco-santos  / @thibmeu any concerns from gateway operator side?

fabricedesre:
I have a couple of questions:

1. Why not use a `/.well-known/` url instead of `_redirect` ? There are many use cases that seem similar (see https://en.wikipedia.org/wiki/Well-known_URI)
2. Is this applicable to gateways only? I guess we will expect native `ipfs://` and `ipns://` protocol handlers to behave the same in various contexts.

bmann:
@fabricedesre the way I think about this is it mirrors a Netlify or Cloudflare Pages deploy. So, in part, we are following cow paths that are already paved, so using _redirects and the literal same syntax that the two mentioned systems already use.

We'll see where other use cases end up, but the one to look at is PWAs / SPAs that need redirects support like this to be able to be deployed correctly and then viewed / function correctly across many IPFS contexts.

So yes, this is "gateways only" to support publish your app to IPFS but load in an HTTP browser context.

fabricedesre:
I would argue that netlify / cloudflare made the wrong choice of location here, but that ship has sailed :)

Where I'm a bit more worried is focusing on a "HTTP browser context". Accessing IPFS through http is a temporary workaround until ipfs:// protocol handlers are available. Once this is the case (we have one in https://capyloon.org), this proposal adds a constraint to the code that resolves ipfs:// urls, turning it in some kind of application server to keep feature parity with "ipfs over http" urls. This is somewhat worrisome, depending on the implementation details.

rvalle:
@fabricedesre I want to add my POV as an DAPP developer that has been waiting for this functionality for long time, and following this PR closely.

Bringing these comments at this stage, when the PR was open for months is slightly unfair to the effort of contributors. I am not saying that your points are not valid, but they could have been introduced earlier in the process.

For the part of IPFS fundamentals and HTTP workaround. I fully understand the IPFS design principles and see your point. However, HTTP is not going anywhere anytime soon, and coexistence and migration paths will be required for decades. To make a simile from the telco world SMS is alive and kicking in an industry that is already in its 6th generation.

From an DAPP developer point of view, the rewrite feature is a must. My use case is having multiple  URIs been served from the same "text" (template), a core feature of internet's content architecture for which a migration path to IPFS will be required. 

An ipfs:// protocol handlers alone would not fix the problem. And populating millions of CIDs for slightly different content (templating) would be as wasteful as creating an HTTP resource for each slightly different content, it is not going to happen.

Will there be an efficient, IPFS native, way to implement a substitute for content templating how would an ipfs:// entrypoint to templated content look like? That is a very fair question to make, but would probably require research both at IPFS level then followed by DAPP application stack adoption. Not happening anytime soon.

All points to the fact that IPFS stack needs to offer a migration path, otherwise, why an HTTP gateway at all?

I think the contributor has made an exceptional effort in providing this feature, carefully integrating with the current state of art and minimizing the impact on the HTTP gateway functionality, respecting its "transitional" nature. 

> We'll see where other use cases end up,

This is a wise statement, for many of those use cases will need an IPFS native alternative.

http-gateways/REDIRECTS_FILE.md
```diff
@@ -0,0 +1,173 @@
...
+For example:
+
+```
+/posts/* /articles/:splat
+```
+
+This rule will redirect a URL like `/posts/2022/06/15/hello-world` to `/articles/2022/06/15/hello-world`.
+
+Splat logic MUST only apply to a single trailing asterisk, as this is a greedy match, consuming the remainder of the path.
+
```
thibmeu, line 102: ## Comment

A line starting with `#` is going to be considered a comment, and ignored at evaluation time.

For example:

```
# Redirect home to index.html
/home /index.html 301
```

is functionally equivalent to

```
/home /index.html 301
```

thibmeu: the feature is supported by Netlify and Cloudflare Pages implementation of `_redirects` file

justincjohnson: @thibmeu I do have the following in the File Format section.

> Any line beginning with `#` will be treated as a comment and ignored.

Are you suggesting I remove that and call it out more prominently like this?

thibmeu: yes, that's what I am suggesting. thanks for making sure there is more context.

justincjohnson: 👍 I have now given both comments and line termination their own heading.  Thanks!

lidel:
Late feedback is ok, I've [pinged](https://discuss.ipfs.tech/t/web-gateway-support-for-redirects-file/15011) folks on discussion forums for this very reason :)
There is a solid, real life problem raised by @fabricedesre: native `ipfs://` in browsers like Brave.

Let's explore ramifications.

## How `_redirects` leaks into browsers with native URIs

> [..] redirects support like this to be able to be deployed correctly and then viewed / function correctly across many IPFS contexts

This is important.

Today, `ipfs://` and `ipns://` are  backed by HTTP Gateway provided by Kubo, so `_redirects` will "just work". 
But what happens when Brave decides to replace Kubo with built-in IPFS support? Or some other browser?

This is a real question with real ramifixations:  we want to have native `ipfs://` in Brave Mobile next, and we can't bundle Kubo there. Based on initial discussions, most likely we will have block/car fetches from gateway in trustless mode, with UnixFS re-assembly and CID verification on the client. 

Should Brave Mobile ignore `_redirects` and break websites, which work fine on Desktop version? I'd say that is a big no. User's expectation is to load the same website they see on gateway, with the same behaviors, but with additional integrity guarantees. 

:point_right:  This means `_redirects`, for pragmatic reasons, will be something that user agents focused on websites have to support on `ipfs://` and `ipns://`, even if it got born as  "HTTP interop thing". 

## Should we make any changes to this spec to account for native URIs?

I know additional bikeshed is frustrating, but we have some open questions to resolve now, and perhaps solve some long-standing problems around `_redirects` standard:

- :point_right: How to deal with native URIs?
  - I believe we should add a section to this spec. Clearly define behavior on native `ipfs://` and `ipns://` (one way or another)
    - Sounds like we have to support redirects  in these contexts, does not seem like we can ignore _redirects there, but would like to hear arguments against.
- :point_right:  Should we move `_redirects` to [`.well-known/..` URI?](https://en.wikipedia.org/wiki/Well-known_URI)
  - pros 
    - If ever is a time to do it, it is now.
    - UX is better: 
      - it has a nice property of being clearly separate from the website content, makes it less awkward in `ipfs://` and `ipns://` URI contexts
      - removes expectation that lookup for `_redirects` resolution happens on every level of deeper directory trees (our spec is limited to root only, which may be confusing for people who don't believe in reading docs)
  - cons
    - redirect paths are defined in a file which is one or two directories away from the content root
  - If we do this, when specification is finalized I'd like to register it at [IANA](https://www.iana.org/assignments/well-known-uris/well-known-uris.xhtml)
    - Should it be IPFS-specific `.well-known/ipfs/redirects`, or a generic `.well-known/redirects` that other projects could adopt as an alternative location of `_redirects` files? 
      - My preference would be to go with generic `.well-known/redirects` location, maximizing the utility of introduced convention.
      
* * *

In short, I'm leaning towards `ipns://example.com/.well-known/redirects`, as it is an explicit hint for user agents accessing the data, making it more clear for implementers how to keep the same behavior in both HTTP Gateway and native URI contexts.
Web browsers and Web gateways will leverage it, other agents are free to ignore it.

Would appreciate a sanity check on this approach.

justincjohnson:
Thinking aloud here, but I'm open to any needed changes, so this isn't push back per se.

**Are many developers familiar with well-known URIs and the `.well-known/` folder?**
I would argue that most aren't, and exposing it to them feels a bit jarring.

**Are there other projects outside the IPFS ecosystem that desire to use a well known, shared format for redirect specifications?**
I'm not familiar with any, but perhaps this is a known need that truly has demand?
If there are, a downside to being part of this standard is that IPFS implementations and software would need to be updated to keep up with changes to this standard instigated outside the IPFS ecosystem.

**Is there value in simply consolidating config into a single directory?**
As mentioned, it makes clear that the config content isn't to be served.
If so, would `.ipfs/` be better and less jarring for developers?  To me this feels more developer friendly and consistent with how config works in other projects (e.g. `.github/`).

justincjohnson:
Also to be clear, I appreciate @fabricedesre commenting on the implications for serving web content over `ipfs://` and `ipns://`.  I don't have much input there, and those challenges feel broader than this spec, but I'm listening.

rvalle:
I don't know if this will very useful to you, but I am going to relate how I would expect this to work when developing a DAPP front-end meant to be served via IPFS:

It is normally regarded as good practice, i.e. the community is trying to convert development stacks so that the deliverable always works as relative content (there is no URL prefix kown)

So, a DAPP front-end should work when "mounted" on a domain with a _dnslink, but should also work when accessed via CID directly, like:

https://my-dapp.com works
https://ipfs.io/ipfs/ROOT_CID works also
ipfs://ROOT_CID works 

So that the application is not dependent no how in accessed in the IPFS-verse, which makes it feel more "decentralized". 

I am not too familiar with .well-known but I associate it with metadata relative to the serving domain. A quick look at wikipedia states that the prefix "/well-known" is used. that would imply that there can be only one /well-known per domain.

In my opinion _redirect medatadata should be maintained by the same person that maintains the content itself, as they are the ones that know which redirects or rewrites are required. There should not be any assumption of where it is going to be placed: i.e. should work in a "relative" fashion, working regardless of the level at which the content is finally placed.

For example, I can think, imagine I want my application to serve maps/region/region_name content, but all of them are going to be served by a script called map_reagion.js in that folder. Then I could rewrite anything under maps/region/* into map_region.js living under the maps/region folder. map_region.js will use the information on the URL ifself to workout what exactly to render (i.e. zoom in a SVG, for example). Thanks to this setup, I can distribute links pointing to particular regions across the application of third parties. then:

https://my_dapp.com/maps/reagion/US would work thanks to _dnslink
https://ipfs.io/ipfs/ROOT_CID/maps/region/US would work by generic content gateway access
ipfs://ROOT_CID/maps/region/US would also work

Of course It would be possible to call it ./well-known instead of _redirects, but I believe /well-known is meant to be at the root, and that would be strange for the generic content gateway access as the  resource would be https://ipfs.io/ipfs/CID/.well-known/ i guess.

The case of _redirects is specified as living in the "publishing folder", whatever that means, seems more like ROOT_CID. or?

justincjohnson:
Thanks for sharing that background info @rvalle.

As mentioned in the spec, for [security reasons](https://github.com/justincjohnson/specs/blob/feat/gateway-redirects/http-gateways/REDIRECTS_FILE.md#security)...

> This functionality will only be evaluated for [Subdomain](https://docs.ipfs.tech/concepts/ipfs-gateway/#subdomain) or [DNSLink](https://docs.ipfs.tech/concepts/ipfs-gateway/#dnslink) Gateways, to ensure that redirect paths are relative to the root CID hosted at the specified domain name.

So the `_redirects` files is always at the root, and so would `.well-known` or `.ipfs` or whatever if we decided to switch.

It seems to me that making web apps securely accessible from any (possibly nested) IPFS path or from `ipfs://` and `ipns://` URLs is outside the scope of this work, though it certainly makes sense to discuss potential impact.

fabricedesre:
I agree with @lidel comment https://github.com/ipfs/specs/pull/290#issuecomment-1220692311

The IPFS community needs to decide how much "smarts" become standard that need to be supported by all gateways and by native ipfs:// unixfs handlers. Maybe an aside, but the http community didn't standardize that kind of behavior as being part of the HTTP protocol. What is proposed here with redirects looks very much like a  simplified Apache mod_rewrite : while very useful, you don't need that to implement a compliant http server.

Kubo started that trend with serving `index.html` when loading a path with a trailing `/` if `/index.html` exists, which is a legacy we have to live with. Let's be careful to not go too far in that direction and add more undue complexity to IPFS implementations.

I'd be interested to hear from the Iroh people as they implement a new stack (cc @b5 and @dignifiedquire).

bmann:
> Maybe an aside, but the http community didn't standardize that kind of behavior as being part of the HTTP protocol. What is proposed here with redirects looks very much like a simplified Apache mod_rewrite : while very useful, you don't need that to implement a compliant http server.

Also, to be super clear, this is meant to be gateway redirects, not ipfs protocol redirects. Why should we support this? So that web apps work at all, in a portable way, when served up on hosted IPFS HTTP Gateway infrastructure. I think "you built your app wrong, if you publish it to IPFS it will only work on these three commercial gateways" is not a good developer experience.

Saying, run your own gateway and put nginx in front of it with custom commands, is also not a good developer experience.

We totally CAN just have a consortium of major IPFS operators who choose to run interoperable gateways. That is what has happened with the adoption of _redirects in the static file publishing world across Cloudflare, Netlify, Vercel, the buildpacks specs, and other providers and standards. I think this particular spec, applying only to HTTP Gateways, means a strong case for portable interop for static front end publishing.

Switching topics, the IPFS protocol needs to decide what the common patterns are for supporting web apps when NOT served over HTTP. The current mode of Single Page Apps (SPAs) written in React or whatever may not be the common pattern. Right now there are no native IPFS apps.

Should `ipfs://` optionally follow this spec? Maybe, as in llidel's example. With a spec, a particular client could opt into that.

Also +1 that `.well-known` is another HTTP primitive designed for discovery at a DNS root, and shouldn't be conflated with "what should the IPFS protocol do natively".

fabricedesre:
I don't understand how it can be useful if we don't ensure the same behavior when loading from a gateway for browsers with no native ipfs:// support and from a ipfs:// url in browsers that support it. That would break user experience badly, no?

From there, that means that this _redirect feature needs to be more than a gateway spec, and also in a `fetching from ipfs://` spec.

rvalle:
@justincjohnson I understand the goal: 

> to ensure that redirect paths are relative to the root CID hosted ...

But still don't see how that translates into leaving ipfs:// out. I am probably missing something.

ipfs://CID/_redirects where paths are all relative and .. is not allowed, wouldn't that always be inside the root CID ? or absolute paths that will get prefixed by root CID. I think making redirects stay within root CID is easier than the potential ramifications of leaving ipfs out.

justincjohnson:
> But still don't see how that translates into leaving ipfs:// out.

I don't disagree with you @rvalle. 🙂  My response to you was specific to http, clarifying why redirect support only applies to subdomain and dnslink gateways and not to any web content nested arbitrarily deep within the path.

A more general comment, in response to the entire discussion...

The work I did so far focused only on http and gateways because my assumption was there was general agreement on that as a path forward and it just needed a spec and implementation.  The discussion on this issue has helped me see some of the tension in design decisions around what to include in IPFS or something as widely used as gateways, path of config file used, etc, so that has been good for me.  For what it's worth, if a separate spec is needed to clarify how `ipfs://` and `ipns://` URLs should be handled, I'm not opposed.  My gut feeling is that this functionality is still needed though.

lidel (approved):
## This is a net positive improvement.

I went over this discussion again, and believe this IPIP is net positive. 

It is better to have a specification of a built-in solution for common web problems like 404s, redirects, and PWAs, than not. And it is better to reuse preexisting conventions around `_redirects`  than ignore the problem and have different implementations and gateways invent competing things, creating soft vendor lock-in, sabotaging the promise of content-addressed websites.

I am voting YES for two reasons:
- this closes the gap around website hosting, and this IPIP solves multiple problems in this space in a way that is easy to reason about, implement, and translate from Gateway-backed `ipfs://` to native one.
- removes the need for specifying and documenting Kubo's `ipfs-404.html`, which I've seen incorrectly used in catch-all situations (because there was no other way to do PWAs)


## Wrapping up some discussion threads

> What is proposed here with redirects looks very much like a simplified Apache mod_rewrite : while very useful, you don't need that to implement a compliant http server.

In content-addressed contexts, abstractions and separation of concerns is a bit different.
Shipping `_redirects` file with the data itself allows the publisher to set certain expectations (hints) around how content should be deserialized for use in web browser contexts (deserialization happening either on a gateway, or native client).  In my mind the function here is closer to CSS than mod_rewrite (part of published data, provides hints on how to do final presentation, but clients are free to ignore it).

> [..]  a separate spec is needed to clarify how ipfs:// and ipns:// URLs should be handled [..]

We will create separate specs for URIs  – that work is tracked in https://github.com/ipfs/specs/issues/138

> Kubo started that trend with serving index.html when loading a path with a trailing / if /index.html exists, which is a legacy we have to live with. Let's be careful to not go too far in that direction and add more undue complexity to IPFS implementations.

I fully agree, and believe this is a net positive: the idea behind `_redirects` is to provide an upstream solution to multiple real world needs around website hosting (custom 404, catch-all PWA, redirecting legacy links to new paths) that is acceptable to implementers. 

We already have (undocumented)  `ipfs-404.html` implemented in Kubo, which is pretty bad because it is a special thing that solves only one problem, and ignores others (PWAs, redirects). `_redirects` solves multiple problems without inventing IPFS-specific, special snowflake convention for each of them.

Digression: these are not the only problems around legacy behaviors on deserialized paths. 
Some additional context around UnixFS vs other IPLD formats [here](https://discuss.ipfs.tech/t/supporting-ipld-tooling-in-uris/15094?u=lidel).  This is to illustrate that we have a relatively small window to solve problems like `_redirects` before we have so many implementations that things ossify too much.

## Keep _redirects in the root dir

This simple convention ensures paths defined inside `_redirects` start at the same root as the location of the file.
Introducing `.well-known` here would introduce an unnecessary level of indirection.

## Finalizing this IPIP by 2022-09-15

During the IPFS Implementers Sync today, we've picked this IPIP to be used as an example to test the process during the next Implementers Sync on 15th, 6PM UTC (see [IPFS Community Calendar](https://github.com/ipfs/community#community-calendar) for details). 

If you feel strongly about this, please state your concerns below, and feel free to join the call.
We will make the final decision during that call. Until then, this PR remains open.

lidel:
## Finalizing this IPIP by 2022-09-15

During the IPFS Implementers Sync today, we've picked this IPIP to be used as an example to test the process during the next Implementers Sync on 15th, 6PM UTC (see [IPFS Community Calendar](https://github.com/ipfs/community#community-calendar) for details). 

If you feel strongly about this, please state your concerns below, and feel free to join the call.
We will make the final decision during that call. Until then, this PR remains open.

IPIP/0000-gateway-redirects.md
```diff
@@ -0,0 +1,76 @@
...
+URL rewriting in particular is a critical feature for hosting SPAs, allowing routing logic to be handled by front end code. SPA support is the primary impetus for this RFC.
+
+Currently the only way to handle URL redirects or rewrites is with additional software such as NGINX sitting in front of the Gateway. This software introduces operational complexity and decreases the uniformity of experience when navigating to content hosted on a Gateway, thus decreasing the value proposition of hosting web sites in IPFS.
+
+This IPIP proposes the introduction of redirect support for content hosted on Subdomain or DNSLink Gateways, configured via a `_redirects` file residing underneath the root CID of the web site.
+
+## Detailed design
+
+Allow developers to configure redirect support by adding redirect rules to a file named `_redirects` stored underneath the root CID of their web site.
+The format for this file is similar to those of [Netlify](https://docs.netlify.com/routing/redirects/#syntax-for-the-redirects-file) and [Cloudflare Pages](https://developers.cloudflare.com/pages/platform/redirects) but only supporting a subset of their functionality.
```
b5, line 24: Is it a strict subset? Deviation would not be nice here

http-gateways/REDIRECTS_FILE.md
```diff
@@ -0,0 +1,190 @@
...
+For example:
+
+```
+/posts/* /articles/:splat
+```
+
+This rule will redirect a URL like `/posts/2022/06/15/hello-world` to `/articles/2022/06/15/hello-world`.
+
+Splat logic MUST only apply to a single trailing asterisk, as this is a greedy match, consuming the remainder of the path.
```
SgtPooki, line 97: With a rule like `/posts/:month/:day/:year/:slug /articles/:year/:month/:day/:slug`

If a gateway receives a request `gateway.ipfs/posts/september`, should the gateway attempt to redirect to `/articles/NULL/september` ? What is the process for handling this?

SgtPooki: we should call out how we handle partial paths without an explicit `*` in the rule.

IPIP/0000-gateway-redirects.md
justincjohnson: Currently it's a strict subset, but my thinking is that IPFS will have different tradeoffs to consider that will warrant not attempting to stay fully in sync.

b5:
We ratified this proposal in the IPFS Implementers call Thursday, Sept. 15! 🎉 

Meeting notes: https://pl-strflt.notion.site/IPFS-Implementers-Sync-2022-09-15-57ee45406157428db7b2df6a3b5538bd

http-gateways/REDIRECTS_FILE.md
justincjohnson: > With a rule like `/posts/:month/:day/:year/:slug /articles/:year/:month/:day/:slug`
> 
> If a gateway receives a request `gateway.ipfs/posts/september`, should the gateway attempt to redirect to `/articles/NULL/september` ? What is the process for handling this?

This will not match at all since the path is missing parts of the specified `from` path.  You would have to explicitly add additional rules to handle different paths.

justincjohnson: > we should call out how we handle partial paths without an explicit `*` in the rule.

If I understand you correctly, there's no need to document this since there's no such thing as a partial path / partial match.  The entire `from` path must match for the rule to fire, just like the prior art we're mimicking.

lidel:
My plan is to merge this  next week. Checklist before that happens:
- [ ] make editoral changes (rename to IPIP 0002 etc)
- [ ] add  max filesize of accepted _redirects file (https://github.com/ipfs/go-ipfs-redirects-file/issues/11)
- [ ] add any clarifications that come up before now and then, and during reviews of https://github.com/ipfs/kubo/pull/8890

IPIP/0000-gateway-redirects.md
```diff
@@ -38,7 +38,7 @@ If a `_redirects` file exists but is unable to be processed, perhaps not even pa
 The detailed specification is added in [`http-gateways/REDIRECTS_FILE.md`](../http-gateways/REDIRECTS_FILE.md).
 
 ### Test fixtures
-QmcZzEbsNsQM6PmnvPbtDJdRAen5skkCxDRS8K7HafpAsX
+QmXqMQ5Sja34NaBEFD7x1hH4VTZWEGmmFF8fGQvAqRsJ1e
```
lidel: @justincjohnson this CID is not provided on public DHT, and the CID in https://github.com/ipfs/kubo/pull/8890 is now different (`QmQprDnhzptPQZwC5dnSvGgmeF9yKCvnNqWA94yVAZCv3a`).

Should we update spec to `QmQprDnhzptPQZwC5dnSvGgmeF9yKCvnNqWA94yVAZCv3a`?

justincjohnson: Yes indeed, we should update to `QmQprDnhzptPQZwC5dnSvGgmeF9yKCvnNqWA94yVAZCv3a`.  Sorry about that.  Are you going to update that or should I?

justincjohnson: Updates have been pushed.

lidel:
Thank you @justincjohnson for driving this IPIP, and everyone for feedback :heart: 


Quick summary:

- This is the first real IPIP that passed improvement process, it has both spec and working implementation.
- We've rattified this spec last week.
- The first Implementation just landed in Kubo  (https://github.com/ipfs/go-ipfs/pull/8816), and  is scheduled to ship next week in Kubo 0.16 RC, allowing us to start using this in web contexts.
- We've updated test fixtures to include learning from tests and fuzzing done in Kubo.
- If anyone is interested, we are also preparing end-user docs in  https://github.com/ipfs/ipfs-docs/pull/1275


Thanks again, and see you on future IPIPs!  :ship:

PR #292, opened 2022-06-22
refactor: split reframe specs

lidel:
This PR decouples Transports and Known Methods from the generic RPC protocol and moves everything under `./reframe` directory, as proposed in https://github.com/ipfs/specs/pull/287#pullrequestreview-1013620192

Filled as PR against that PR for easier review (same content, different presentation).

(cc @aschmahmann @petar @willscott)

## Why should we do this split?

We don't have to, but it will make things easier to navigate going forward:
It allows us to add/modify methods and transports without  touching the protocol itself.

## PREVIEW

Rendered Markdown: https://github.com/ipfs/specs/tree/refactor/split-reframe-specs/reframe

### DIFF
--- a/reframe/README.md
+++ b/reframe/README.md
@@ -1,0 +1,5 @@
+# Reframe Specifications
+
+- [`REFRAME_PROTOCOL.md`](./REFRAME_PROTOCOL.md) ← start here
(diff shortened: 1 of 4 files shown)


petar (approved):

lidel:
@aschmahmann ok to this into #287, and merge that one as well?

aschmahmann (approved):

PR #294, opened 2022-06-27
unixfs: suggest, but do not strictly mandate unixfs<>dagpb

rvagg:
The strict coupling poses some problems for IPLD implementations that:

 1. do not retain codec information between the serialization and UnixFS (ADL) reification; and/or
 2. do not have a mechanism to strictly mandate that UnixFS data be _only_ encoded in a particular codec.

A mandate in the specification that strictly defines the layering on top of the codec makes it difficult to implement it as an ADL, which also presents difficulty for using tooling that builds on IPLD lens-style layering.

Ref: https://github.com/ipfs/go-unixfsnode/pull/27
Ref: https://github.com/ipfs/specs/pull/271
Ref: https://github.com/ipld/go-car/pull/304

---

So, this is basically a take-two on https://github.com/ipfs/specs/pull/271, but with an extra helping of nuance in an attempt to open the door slightly, but not too far that this complicates life for existing implementations. The main aim here is to _allow for_ the kind of layering approach that we have with go-ipld-prime + ADLs. As per https://github.com/ipld/go-car/pull/304 we already have situations where we have enough distance between block decode and ADL interpret that we can't properly apply a type check. This gives us wriggle-room to pass the data through the generic go-ipld-prime "lens" paradigm and let it come out as UnixFS; _even if_ in practice we tightly couple it when encoding the data and may impose an expectation when decoding, where possible.

But also, it'd be good to have at least the potential for separation so we can treat the UnixFS ADL as a full ADL and not a special-cased beast.

At the same time this PR updates the language of this section to current IPLD parlance.

Any stomach for nuance?

### DIFF
--- a/UNIXFS.md
+++ b/UNIXFS.md
@@ -78,7 +78,7 @@ This `Data` object is used for all non-leaf nodes in Unixfs.
 
 For files that are comprised of more than a single block, the 'Type' field will be set to 'File', the 'filesize' field will be set to the total number of bytes in the file (not the graph structure) represented by this node, and 'blocksizes' will contain a list of the filesizes of each child node.
 
-This data is serialized and placed inside the 'Data' field of the outer merkledag protobuf, which also contains the actual links to the child nodes of this object.
+This data is serialized and placed inside the 'Data' _Bytes_ node of the containing IPLD block, which also contains the actual links to the child nodes of this object in a 'Links' _List_ node. Typically this is encoded using the [DAG-PB](https://ipld.io/specs/codecs/dag-pb/spec/) codec. An implementations of this UnixFS specification may opt to strictly link DAG-PB to UnixFS for encoding and/or decoding as this is the originally intended layering of this format. For this reason, producers of UnixFS data that do not use DAG-PB as its codec should not expect other implementations of UnixFS to be able to interpret the data.
 
 For files comprised of a single block, the 'Type' field will be set to 'File', 'filesize' will be set to the total number of bytes in the file and the file data will be stored in the 'Data' field.
 


willscott:
As in #271 I would prefer that the ADL works as an ADL, so i would be in favor of this spec change.

UNIXFS.md
```diff
@@ -78,7 +78,7 @@ This `Data` object is used for all non-leaf nodes in Unixfs.
 
 For files that are comprised of more than a single block, the 'Type' field will be set to 'File', the 'filesize' field will be set to the total number of bytes in the file (not the graph structure) represented by this node, and 'blocksizes' will contain a list of the filesizes of each child node.
 
-This data is serialized and placed inside the 'Data' field of the outer merkledag protobuf, which also contains the actual links to the child nodes of this object.
+This data is serialized and placed inside the 'Data' _Bytes_ node of the containing IPLD block, which also contains the actual links to the child nodes of this object in a 'Links' _List_ node. Typically this is encoded using the [DAG-PB](https://ipld.io/specs/codecs/dag-pb/spec/) codec. An implementations of this UnixFS specification may opt to strictly link DAG-PB to UnixFS for encoding and/or decoding as this is the originally intended layering of this format. For this reason, producers of UnixFS data that do not use DAG-PB as its codec should not expect other implementations of UnixFS to be able to interpret the data.
```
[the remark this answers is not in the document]
ribasushi: @rvagg adding a sentence ( it's not great, rewording welcome ) to hammer home that the "outer" struct is pretty much ossified as well: you might encode it using something else, but the flat list of links must remain a flat list of named links.

ribasushi (changes requested):

UNIXFS.md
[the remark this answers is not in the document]
rvagg: checkout latest commit @ribasushi, I've referenced the logical format schema in the dag-pb spec to lock in the shape of encoded unixfs data, regardless of codec used

UNIXFS.md
```diff
@@ -78,7 +78,7 @@ This `Data` object is used for all non-leaf nodes in Unixfs.
...
+This data is serialized and placed inside the 'Data' _Bytes_ node of the containing IPLD block, which also contains the actual links to the child nodes of this object in a 'Links' _List_ node. As such, a serialized UnixFS block must conform to the [DAG-PB Logical Format](https://ipld.io/specs/codecs/dag-pb/spec/#logical-format) schema. Typically this is then encoded using the [DAG-PB](https://ipld.io/specs/codecs/dag-pb/spec/) codec. An implementation of this UnixFS specification may opt to strictly link DAG-PB to UnixFS for encoding and/or decoding as this is the originally intended layering of this format. For this reason, producers of UnixFS data that do not use DAG-PB as its codec should not expect other implementations of UnixFS to be able to interpret the data.
```
ribasushi, line 81:
```suggestion
This data is serialized and placed inside the 'Data' _Bytes_ node of the containing IPLD block, which also contains the actual links to the child nodes of this object in a 'Links' _List_ node. As such, a serialized UnixFS block **MUST** conform to the [DAG-PB **Logical Format**](https://ipld.io/specs/codecs/dag-pb/spec/#logical-format) schema. Typically this is then encoded using the [DAG-PB](https://ipld.io/specs/codecs/dag-pb/spec/) codec. An implementation of this UnixFS specification may opt to strictly link DAG-PB to UnixFS for encoding and/or decoding as this is the originally intended layering of this format. For this reason, producers of UnixFS data that do not use DAG-PB as its codec should not expect other implementations of UnixFS to be able to interpret the data.
```

ribasushi: @rvagg this works! Adding a little more screamers to make sure people do not miss it, but good enough. Thank you!

rvagg: I would normally highlight SHOULD, MUST, MAY, etc. in a spec but this one doesn't already have that where it uses those words and I didn't want to make this section too different, nor do I want to increase the scope of the PR.
Happy to change it if others feel strongly enough though.

warpfork (approved):
My 2c:
- this is a desirable change, directionally (same reasoning as Will)
- I think this reads pretty great -- the parlance updates are good and clear, and the cautions in all directions about compatibility seem accurate and appropriate.

Well written, @rvagg <3

willscott:
what else is needed to merge this?

lidel (changes requested):
> open the door slightly, but not too far that this complicates life for existing implementations

I am late to the party, but how confident are we the below won't happen?

1. something like  https://github.com/ipfs/go-unixfsnode/pull/27 allows a dag-pb node to have been serialized to cbor or json, and still be interpreted as unixfs data – it gets merged, every go app supports it now
2. we relax spec here
3. people see both spec and go library allow cbor, start producing UnixFS dags with cbor  instead of dag-pb because "it works in Kubo and Gateways already and we have 1 less codec to implement"
4. complicated life begins:
   - someone tries to implement a minimal IPFS node in new language from scratch, and they now need to debug things because their implementation can't correctly read "valid UnixFSv1"
   - legacy node is unable to read new "valid UnixFSv1"  created by some alternative IPFS implementation that decided to use cbor instead of pb by default because both specs and most common go-library allow it, and they dont care about "legacy" dag-pb

In my mind, this PR should bump version to UnixFS v2 because it allows for new type of UnixFS data to be created, that is not  interoperable with v1.5.
Am I missing something?

PR #297, opened 2022-07-06
gateway(only-if-cached): note on handling big DAGs

lidel:
This PR adds a note for Gateway implementers handling `only-if-cached` requests for DAGs that they should at the very least check if the root block is cached.


http-gateways/PATH_GATEWAY.md
```diff
@@ -158,13 +158,17 @@ Used for HTTP caching.
 
 Client can send `Cache-Control: only-if-cached` to request data only if the
 gateway already has the data (e.g. in local datastore) and can return it
-immediately.
+immediately.
```
thibmeu, line 161: nits: I think this is a typo

lidel: yep, fixed!

vasco-santos (approved):

thattommyhall (approved):

iand (approved):

Issue #298, opened 2022-07-15
Create spec for (allow|deny)lists for IPFS Nodes and Gateways


foreseaz:
Open issue for IPIP: Denylists for IPFS Nodes and Gateways.

PR: https://github.com/ipfs/specs/pull/299

lidel:
Thank you for filling this. I will ping interested stakeholders & historical, related threads:
- [Allow and deny lists for CIDs js-ipfs#2152](https://github.com/ipfs/js-ipfs/issues/2152)
- [Implement Allow Lists and Block Lists notes#284](https://github.com/ipfs/notes/issues/284)
  - [gateway limitations kubo#1070](https://github.com/ipfs/kubo/issues/1070)
- [Murmuration-Labs/Songbird: Decentralized Content Moderation & Assessment Proposal](https://github.com/Murmuration-Labs/songbird-decentralized-moderation/)
- IPFS Camp: [Do not replicate lists - DNR](https://github.com/ipfs/camp/blob/master/DEEP_DIVES/25-do-not-replicate-lists-dnr.md)

PR #299, opened 2022-07-15
IPIP: format for denylists for IPFS Nodes and Gateways

foreseaz:
This IPIP adds spec for (deny|allow)lists for IPFS Nodes and Gateways.

Supported rules:
- plain
  - `cid` 
  - `content_path`
- hashed (for trully bad bits)
  - `hashed_cid`
  - `hashed_content_path`

Closes https://github.com/ipfs/specs/issues/298

### DIFF
--- a/IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
+++ b/IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
@@ -97,6 +97,8 @@ The other operation benefit comes after the wide adoption of the proposed denyli
 
 No existing implementations yet.
 
+JSON format is used to maximize interoperability.  The intent is for IPFS implementations and services to standardize content filtering around this format for exchanging and storing allow and deny lists.
+
 ### Security
 
 The following concern may not lie in this scope, but it is worth to be mentioned in this proposal. The blocking of CIDs which are not malicious and are widely used can potentially jeopardize the availability of multiple sites on that IPFS gateway. Possible examples include common images, javascript bundles, or stylesheets.
--- a/IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
+++ b/IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
@@ -1,0 +1,112 @@
+# IPIP 0002: Denylists for IPFS Nodes and Gateways
+
+- Start Date: 2022-07-14
+- Related Issues:
+  - [ipfs/specs/issues/298](https://github.com/ipfs/specs/issues/298)
+
+## Summary
+
+This issue proposes a descriptive and maintainable denylist, leaving open the possibility of allowlists in the future. These can be used to help standardize content moderation for IPFS nodes and gateways.
+
+## Motivation
+
+The current [Bad Bit Denylist](https://badbits.dwebops.pub/denylist.json), which is a list of hashed CIDs, has multiple major drawbacks for implementation. One is lacking support for content path, it's hard to block an IPNS path only based on CIDs. The other is lacking description and status code response for each anchor. The description is especially for hashed blocking items, since the hashing is unidirectional, the description can outline the reason why it's been blocked thus increasing maintainability.
+
+A well-thought-out denylist schema can ease the implementation of the denylist, and bring consensus on how content blocking works for gateway and node operators in the community.
+
+## Detailed design
+
+
+### Denylist Schema
+
+Here is the proposed denylist and each field will be explained in detail later.
+
+```js=
+{
+  action: "block",
+  entries: [
+    {
+      type: "cid",
+      content: "bafybeihfqymzmqhbutdd7i4mkq2ltzznzgoshi4r2pnv4hsc2acsojawoe",
+      description: "ipfs quick start",
+      status_code: 410
+    },
+    {
+      type: "content_path",
+      content: "/ipns/example.com",
+      description: "example.com",
+      status_code: 410
+    },
+    {
+      type: "content_path",
+      content: "/ipfs/bafybeihfqymzmqhbutdd7i4mkq2ltzznzgoshi4r2pnv4hsc2acsojawoe",
+      description: "ipfs readme",
+      status_code: 451
+    },
+    {
+      type: "hashed_cid",
+      content: "9056e0f9948c942c16af3564af56d4bb96b6203ad9ccd3425ec628bcd843cc39",
+      description: "sensitive cid that needs to be blocked",
+      status_code: 451
+    },
+    {
+      type: "hashed_content_path",
+      content: "65e60fcaa506ca5b0b49d7aa73df5ba32446bddb4e72a1f8bb5df12eaaaa8745",
+      description: "sensitive content path that needs to be blocked",
+      status_code: 410
+    }
+  ]
+}
+```
+
+#### `action`
+
+Though it's called `denylist`, the only allowed action field here is `block`. Other actions such as `allow` can be added in the future to enable allowlists or other types of content lists.
+
+
+#### Each denylist entry
+
+- `type`: specifies the type of the content should be blocked. E.g. `cid`, `hashed_content_path`.
+- `content`: stores the content that should be blocked according to the type. It's suggested that all CIDv0 needs to be converted into CIDv1 to keep the consistency.
+- `content_path`: the content path needs to be blocked).
+- `description`: description of the CIDs or content paths.
+- `status_code`: status code to be responded for the blocked content. E.g. [410 Gone](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#410-gone); [451 Unavailable For Legal Reasons](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#451-unavailable-for-legal-reasons) or `200 OK` for allowed entry.
+
+**Side notes on `hashed_cids` & `hashed_content_paths` types**
+
+The main difference between non-hashed entries and hashed ones is that the CIDs or content paths in the entry will be hashed and no plaintext is shown in the list. Following the [bad bits](https://badbits.dwebops.pub/), each CID or content path is `sha256()` hashed, so it's easy to determine one way but not the other. The hashed entries are designed to store sensitive blocking items and prevent creating an easily accessible list of sensitive content.
+
+Before the hashing, all CIDv0 in both `cid` field and `content_path` fields are converted to CIDv1 for consistency.
+
+## Design rationale
+
+The gist of the rationale is tackling the inconveniences of blocking implementation when using [current denylist](https://badbits.dwebops.pub/denylist.json). Adding support for content path, description, status code and action can make the denylist more maintainable, extensible, and easier to implement. This is especially true when a list is maintained by multiple parties or needs to keep records for auditing.
+
+Other minor design decisions, including CIDv1 normalization, allowing both plain text and hashed entries in one denylist, are also made to ease the implementation. It will force the consistency between multiple denylists that pave the way for wider adoption.
+
+Denylists are important to empower gateway to make their own policy decisions. While gateway can access any IPFS content, they might decide to not serve it all. This could be for reputation concerns, for safety, or for internal reasons.
+
+
+### Operation benefit
+
+The proposed schema could ease the implementation of denylist for gateway and node operators. It supports both CID and content path and each entry has a customizable description and response status code.
+
+The other operation benefit comes after the wide adoption of the proposed denylist, a new onboard gateway operator can use a shared to start a gateway right away.
+
+### Compatibility
+
+No existing implementations yet.
+
+### Security
+
+The following concern may not lie in this scope, but it is worth to be mentioned in this proposal. The blocking of CIDs which are not malicious and are widely used can potentially jeopardize the availability of multiple sites on that IPFS gateway. Possible examples include common images, javascript bundles, or stylesheets.
+
+### Alternatives
+
+[Bad Bits Denylist](https://badbits.dwebops.pub/) is focusing on blocking public flagged CIDs for IPFS node operators, the blocking mainly happens between the nodes.
+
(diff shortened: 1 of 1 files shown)


IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,113 @@
...
+
+```js=
+{
+  action: 'block',
+  cids: [
+    {
+      cid: 'bafybeihfqymzmqhbutdd7i4mkq2ltzznzgoshi4r2pnv4hsc2acsojawoe',
+      description: "ipfs quick start",
+      status_code: 410,
+      type: "hashed_cids"
```
foreseaz, line 32: Typo here, should be removed.

lidel (changes requested):
Thank you for submitting this @foreseaz.

As noted in https://github.com/ipfs/specs/issues/298#issuecomment-1192538478 IPFS project has a clear need for standardizing allow/deny lists, and discussion around this IPIP is welcome.

I will be pinging some stakeholders for additional review, but my initial feedback/asks:
- [ ] To be aligned with the rest of IPFS stack, we need an escape hatch for other hash functions than `sha2-256` – details inline
- [ ] Make it clear which fields are optional (e.g. `description` and `status_code`) and what is the implicit default  (empty `description` and `status_code` `410`)
- [ ] Add top level optional `description` field where denylist maintainers can include additional context (purpose, policy, link to more details)

Jorropo:
Why adding content path ? Everything could be done with cids, you just include intermediaries cids.

It seems we will have to write code checking intermediaries cids anyway in case the path is reachable multiple ways.

mathew-cf:
> Why adding content path ? Everything could be done with cids, you just include intermediaries cids.
> 
> It seems we will have to write code checking intermediaries cids anyway in case the path is reachable multiple ways.

Content paths allow us to block a specific domain. So if /ipns/bad-website.com keeps updating their content, we'll continue to block requests to that domain.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,112 @@
...
+
+
+### Denylist Schema
+
+Here is the proposed denylist and each field will be explained in detail later.
+
+```js=
+{
+  action: "block",
+  entries: [
```
lidel, line 27: Question raised during triage today: what happens when the block list grows to megabytes?
This is a real concern, as https://badbits.dwebops.pub/ alone is getting close to 900KiB, and history shows that even efficient pattern-matching things like adblock lists are multiple megabytes in size ([example](https://easylist.to/easylist/easylist.txt)).

This spec should provide a way for representing handle big, big blocklists. 

:bulb:  One idea is to introduce a special entry with `type: "import"` and  either `cid` or `content_path` pointing at some other list.   This provides a solution for sharding and maintaining big lists AND allows composing denylists using existing ones.

mathew-cf: For sharded lists, we may need to specify a prefix or something to indicate which list is for which shard

Jorropo: @mathew-cf you want to spec JSON based HAMT ?

mathew-cf: now that you mention it @Jorropo i think composability is probably enough for now lol

hsanjuan:
I'll throw my two cents:

* We don't need ONE list-format spec. If anything we need several.
* This one is particularly bad for processing (parsing an object, with a huge array inside, with objects inside), when all I need are CIDs or paths. Go cannot read each CID in this file without parsing it all into memory for example.
* Whatever the list format, it is probably going to be quite easy to parse anyways. The question is what is the best format for this? But for a start let's just stick to the badbits format which is no good, but at least we don't have multiple formats that suck in the same ways (like being JSON objects).
* Where is the badbits format documented? I always forget that.
* We need a discussion on what formats are good and for what:
  - Easy to read and edit and append: a text file with 1 cid per line, or one path per line, or one double-hash per line (plus modifiers)
  - Fast to move around (gzipped stuff)
  - Easy to share over IPFS (an IPLD linked lists of things)
  - Store-the-hash database files
  - CAR files.
  - What about multiple files?
  - How does ublock &co does it?
  - What about blocking by regexp'ed path?
* Another discussion is what to exactly to doublehash:
  - Cid + path (like badbits)
  - Resolved paths?
  - Multihash? 
* Block lists should be have as little information as possible.
* Any double hashing should happen using the fastest hash function available (and no need to use very long hashes here)
* I think everything should be double hashed. Why not.
* Details, response codes, descriptions, original hashes etc etc do not belong in the denylists. If you need that keep it out of the denylist, which is derived from it. If anything, put is as comments and not as part of the schema, so that the parser can ignore it.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,112 @@
...
+Though it's called `denylist`, the only allowed action field here is `block`. Other actions such as `allow` can be added in the future to enable allowlists or other types of content lists.
+
+
+#### Each denylist entry
+
+- `type`: specifies the type of the content should be blocked. E.g. `cid`, `hashed_content_path`.
+- `content`: stores the content that should be blocked according to the type. It's suggested that all CIDv0 needs to be converted into CIDv1 to keep the consistency.
+- `content_path`: the content path needs to be blocked).
+- `description`: description of the CIDs or content paths.
+- `status_code`: status code to be responded for the blocked content. E.g. [410 Gone](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#410-gone); [451 Unavailable For Legal Reasons](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#451-unavailable-for-legal-reasons) or `200 OK` for allowed entry.
```
Jorropo, line 73: This should be authoritative.
I think anything not `451`, `410` or `200` has no place on a badlist list. I don't want people start to use 3xx or whatever.

Secondly, I don't fully understand why giving out HTTP codes, assuming the goal is to join forces and let gateway operators share ban reasons, **this is really unspecific**,
I would like a machine readable version enum, containing things like `Dos`, `Legal` (maybe with a reason like `Legal Copyright`, `Legal Hatespeech`, `Legal other`, ...), ...
Some gateway operators might prefer faking `500`s or just flat out ignore certain reasons.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,112 @@
...
+
+**Side notes on `hashed_cids` & `hashed_content_paths` types**
+
+The main difference between non-hashed entries and hashed ones is that the CIDs or content paths in the entry will be hashed and no plaintext is shown in the list. Following the [bad bits](https://badbits.dwebops.pub/), each CID or content path is `sha256()` hashed, so it's easy to determine one way but not the other. The hashed entries are designed to store sensitive blocking items and prevent creating an easily accessible list of sensitive content.
+
+Before the hashing, all CIDv0 in both `cid` field and `content_path` fields are converted to CIDv1 for consistency.
```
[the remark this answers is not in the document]
Jorropo: I don't support base32, whoever is hashing this is already IPFS specific details so I don't think there is value in using a text based format.
Hashing raw binary CIDs is faster and use less memory.

This is a non forward compatible change, pls be carefull.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
Jorropo, line 73: A status code of `200` is weird, you already have the action field saying if it's blocked are allowed, what about just saying the usage of `status_code` is undefined if the content is allowed ?

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,112 @@
...
+- `content`: stores the content that should be blocked according to the type. It's suggested that all CIDv0 needs to be converted into CIDv1 to keep the consistency.
+- `content_path`: the content path needs to be blocked).
+- `description`: description of the CIDs or content paths.
+- `status_code`: status code to be responded for the blocked content. E.g. [410 Gone](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#410-gone); [451 Unavailable For Legal Reasons](https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#451-unavailable-for-legal-reasons) or `200 OK` for allowed entry.
+
+**Side notes on `hashed_cids` & `hashed_content_paths` types**
+
+The main difference between non-hashed entries and hashed ones is that the CIDs or content paths in the entry will be hashed and no plaintext is shown in the list. Following the [bad bits](https://badbits.dwebops.pub/), each CID or content path is `sha256()` hashed, so it's easy to determine one way but not the other. The hashed entries are designed to store sensitive blocking items and prevent creating an easily accessible list of sensitive content.
```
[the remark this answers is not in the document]
Jorropo: I don't see why not using the multibase prefix, it's cheap to compare, and allows people to use a more compact base64 or base2048 in the future.

We don't need the full cid, so I do like @lidel's idea of just using the multihash.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
vasco-santos: Hey folks,
We have been operating an internal denylist that is synced with badbits in {nft/web3}.storage. We need to align well on this direction with content path. There is at least a limitation around this that we have found: 
- if we block a CID with content path, if user just tries to fetch the resource in the path via its own CID (present in response etag from gateway), it won't be flagged or we will need a follow up check in the denylist
- same as before, but on the other way around

mathew-cf: A workaround for the first case (content path blocked, bypassed via CID) could be to resolve the content path and then block that CID as well.
For the second case (CID blocked, bypassed with content path), we're planning on using `x-ipfs-roots` to ensure that none of the resolved CIDs  are blocked

b5:
This IPIP is scheduled for discussion & ratification at the next IPFS Implementers call September 29th, 2022! https://calendar.google.com/calendar/u/0/embed?src=ipfs.io_eal36ugu5e75s207gfjcu0ae84@group.calendar.google.com

fabricedesre:
I have a few questions about this proposal:

- How are IPFS services expected to exchange lists?
- How are conflicts resolved when the same resource has different status_code (or whatever reason mechanism ends up being used) in different lists?
- If other actions than `deny` are added in the future, how is conflict resolution envisioned  if the same resource ends up being denied and allowed?
- When IPFS will be successful, scalability of these lists will be a problem. Json is not very good even as an interchange format for large data sets.

Also, how do we trust these lists? If we keep a decentralized design, I feel that a signature should be added to blocking records to verify the source of a blocking decision.

Web browsers use a blocking mechanism called "Safe Browsing" (see https://wiki.mozilla.org/Security/Safe_Browsing for details on the Firefox implementation). The design is very different, since it relies on a centralized service but it has interesting privacy characteristics to prevent leaking the client's navigation while minimizing the amount of extra resources needed (both bandwidth and storage). I think it would be interesting to build something similar on top of the denylist so that light IPFS nodes don't have to download and update large datasets.

IPIP/0002-denylists-for-ipfs-nodes-and-gateways.md
vasco-santos: We are currently hitting a limitation with current badbits implementation, and thoughts here would not solve that as well.

Currently in nftstorage.link + w3s.link, we rely on anchors fetched from badbits. We put them in a KV Store in the edge and before resolving content (including via Cache), we check KV Store and if is there, we don't serve content. If cid+path resolution, we will also check `ETag` presence in our denylist before serving.

However, we want to move forward with more agressive caching, and current approach leads to problems that we cannot solve. Let's consider for instance, https://blog.cloudflare.com/introducing-cache-reserve/ that allows content to be permanently cached on the Edge for up to 30 days (without HITs). This would be a desirable feature for a gateway, given the costs to use this kind of feature are considerably lower than the needed bandwidth to go to the origin all the time. So, if a gateway wants to rely on this kind of feature, it needs to be proactive on purging from Cache bad content. And so, what do we need to purge the content? We need its HTTP URL, aka cid+path.

With the above in mind, I think we should attempt to move out of the `hashed_cid` + `hashed_content_path` types. Having root CIDs and content path are likely what we need.

lidel (changes requested):
:warning:  I'm positive, but  this IPIP needs some additional work before merge:

- Some open comments about the way hashes are generated/normalized are not resolved
- We are missing a working implementation – is there any library in the works @foreseaz?
- Missing test fixtures – if we had sample implementation, we could reuse test vectors.
- Bit worried about low engagement from IPFS implementations in general 
  - I suggest we continue the discussion, but perhaps park ratification talk   until https://github.com/ipfs/kubo/issues/8492 is picked up? (or a sibling JS/Rust implementation starts similar work)

lidel:
We've discussed this IPIP during IPFS Implementers call today and want to move forward, byt @foreseaz  is no longer working on it which makes applying changes challenging.


Next steps:
- I am closing this PR.
- @mathew-cf will go over feedback and open a new PR and link below so we can continue there.

foreseaz:
> ⚠️ I'm positive, but this IPIP needs some additional work before merge:
> 
> * Some open comments about the way hashes are generated/normalized are not resolved
> * We are missing a working implementation – is there any library in the works @foreseaz?
> * Missing test fixtures – if we had sample implementation, we could reuse test vectors.
> * Bit worried about low engagement from IPFS implementations in general
>   
>   * I suggest we continue the discussion, but perhaps park ratification talk until [ipfs/kubo#8492](https://github.com/ipfs/kubo/issues/8492) is picked up? (or a sibling JS/Rust implementation starts similar work)

I apologize for the late reply and really sorry that I haven't been following the discussion since I was back to school last month.
 
> * We are missing a working implementation – is there any library in the works @foreseaz?
There is working implementation related to work, but it may not be shared.

Thanks for the discussion on the IPIP today, I just saw the issue has been closed, I may miss the on-going discussion in work channels, so either @mathew-cf creates a new IPIP PR or re-open the issue and I'll update the current PR works for me. Thanks again!

LaurenSpiegel:
Hi, is another PR in the works?

thibmeu:
I've iterated over this commit and opened https://github.com/ipfs/specs/pull/340

PR #306, opened 2022-08-08
fix: links in  naming/pubsub.md

mrodriguez3313:
Link in last line (line 94) is broken. I couldn't find which one it was supposed to be redirected to.

### DIFF
--- a/naming/pubsub.md
+++ b/naming/pubsub.md
@@ -50,7 +50,7 @@ where base64url-unpadded is an unpadded base64url as specified in [IETF RFC 4648
 
 # Layering persistence onto libp2p PubSub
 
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thanks! Namesys got moved to https://github.com/ipfs/go-namesys, I'll update the link.

naming/pubsub.md
```diff
@@ -91,4 +91,5 @@ every 10 seconds)
 
 # Implementations
 
-  - <https://github.com/ipfs/go-ipfs/tree/master/namesys>
+  - <https://github.com/ipfs/go-namesys>
```
lidel, line 94:
```suggestion
  - Kubo:  <https://github.com/ipfs/go-namesys> and <https://github.com/libp2p/go-libp2p-pubsub-router>
```

PR #311, opened 2022-08-23
Reframe: specify HTTP response status codes

guseggert:


willscott (approved):
should we specify semantics of 4xx vs 5xx responses?
e.g. 4xx on client errors (malformed request / etc) and 5xx on server errors

guseggert:
How about I say something like "Non-200 status codes SHOULD follow best practices, e.g. 4xx on client errors, 5xx on server errors, etc.".

willscott:
yeah, or it's fine as is. i think the implicit conventions of HTTP basically work for us here.

guseggert:
(Side note: @lidel is out this week so I am merging this without approval.)

PR #312, opened 2022-08-23
update the reframe provide proposed spec

willscott:
* functionally, change `Key` from a single CID to an array of CIDs, to support extension to a `ProvideMany` case.
* update the example message to reflect current implementation in go-delegated-routing

### DIFF
--- a/reframe/REFRAME_KNOWN_METHODS.md
+++ b/reframe/REFRAME_KNOWN_METHODS.md
@@ -26,6 +26,8 @@ This document is defining known methods (request-response message types) and sem
     - [GetIPNS DAG-JSON Examples](#getipns-dag-json-examples)
   - [PutIPNS](#putipns)
     - [PutIPNS DAG-JSON Examples](#putipns-dag-json-examples)
+  - [Provide](#provide)
+    - [Provide DAG-JSON Examples](#provide-dag-json-examples)
 - [Method Upgrade Paths](#method-upgrade-paths)
 - [Implementations](#implementations)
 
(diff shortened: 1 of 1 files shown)


reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -269,7 +269,7 @@ A message for indicating that the client is able to act as a provider for a give
 
 ```ipldsch
     type ProvideRequest struct
```
aschmahmann, line 271: It doesn't seem needed at the moment, but flagging that down the line I could see you wanting to tag a provide request a particular way, or allow some data passed back in the response to allow bulk manipulation if we want to update or delete a provide message (e.g. as storetheindex allows).

reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -269,7 +269,7 @@ A message for indicating that the client is able to act as a provider for a give
 
 ```ipldsch
     type ProvideRequest struct 
-        Key &Any
+        Key [&Any]
```
aschmahmann, line 272: LGTM. This could be a kinded union for 1 or more, but the couple extra bytes for a single provide don't seem like a big deal.

aschmahmann (approved):
No issues from me. @ajnavarro should take a look too since he'll be reviewing PRs for the implementation.

Aside: I noticed the table of contents is missing `Provide`, could you add it in?

ajnavarro (commented):
After updating the TOC, LGTM.

willscott:
er, except that i can't 
image

PR #317, opened 2022-09-01
chore: lint http-gateways markdown

hacdias:
Procedure:

```
markdownlint *.md --disable MD013 --disable MD032 --fix
```

And some manual checks that are not automatically fixed by the linter.


SgtPooki (approved):
Can we modify the linter workflow so that these things are handled automatically in the future?

hacdias:
@SgtPooki we should. Right now, it seems the CI is only linting the files within the changed directory, giving the false impression that everything is fine when it isn't. That explains why the majority of the PRs are actually failing.

PR #319, opened 2022-09-05
feat: update IPNS specification

### LINKED ISSUE #314: Update IPNS specification
- [x] Update [IPNS record section](https://github.com/ipfs/specs/blob/main/IPNS.md#ipns-record) to document support for extensible records (and v2 signature), matching implementation in [GO](https://github.com/ipfs/go-ipns/commit/3deb032d28934baa818e4624507dd0b1fefbbdc6) and [JS](https://github.com/ipfs/js-ipns/pull/121)
- [x] Explain v2 signature generation and validation 
- [x] Make it clear that "Routing record" key is `/ipns/{binary_blob}`  so implementers and devs  don't think it is a bug  :upside_down_face: 
- [x] Make it sure specs are aligned with implementations such as Kubo and JS-IPFS
- [x] Move `IPNS.md` and `naming/pubsub.md` to `ipns` directory 
- [ ] Triage https://github.com/ipfs/specs/issues?q=is%3Aissue+is%3Aopen+ipns
- [ ] Incorporate any insights from https://github.com/aschmahmann/ipns-utils  and https://github.com/ipfs/notes/issues/249

lidel:
> Closes #314

This PR aims to document implementable IPNS Record creation and validation.

I've documented IPNS record creation and verification based on what latest Kubo and JS-IPFS do, but without v1 signatures  (as we want to remove support for them anyway).

Would appreciate additional :eye: from implementers and people working with IPNS.

## TODO

- [x] Describe IPNS Name
  - [x] explain what `libp2p-key` is and how tis protobuf looks like
  - [x] explain when inlining makes sense
  - [x]  document String representation as CID
- [x] Describe IPNS Record protobuf and DAG-CBOR part
  - [x] extensible record data in strict DAG-CBOR
  - [x] signatures v1 creation
  - [x] signatures v2: creation and validation
- [x] Describe IPNS Keys section
- [x] reference DHT and PubSub specs
- [x] backward-compatibility policy
- [ ] go over additional asks listed in #314

### Open questions

- [x] move everything to `ipns/` ?
- [x] is it ok to only require Ed25519? 
  - it is MUST, we also  have RSA which is 'SHOULD'
- [x]  is it ok to state that `IpnsEntry.data` DAG-CBOR  (to avoid duplicating strictness spec from https://ipld.io/specs/codecs/dag-cbor/spec/)
  - yes, go-ipns uses dag-cbor from https://github.com/ipld/go-ipld-prime  internally
- [x] is it ok to simply not document v1 signatures?
  - yes, this is legacy which we want to go away (only relevant to Kubo and JS-IPFS)
- [x] if we dont support v1, do we need to duplicate `value`, `validity`, `validityType`, `sequence`, and `ttl`   (as `IpnsEntry.*` and in CBOR ar `IpnsEntry.data`)?
  - yes, for now we need to keep requirement of duplicated values – see rational in https://github.com/ipfs/go-ipns/pull/42
  

  
cc @2color @mrodriguez3313 @TMoMoreau @rvagg @vasco-santos

### DIFF
--- a/IPNS.md
+++ b/IPNS.md
@@ -1,0 +1,3 @@
+# IPNS Specs Moved
+
+Moved to [./ipns](./ipns/)
--- a/IPNS.md
+++ b/IPNS.md
@@ -42,11 +42,17 @@ IPNS is based on [SFS](http://en.wikipedia.org/wiki/Self-certifying_File_System)
 ### Key Types
 
 Implementations MUST support Ed25519 with signatures defined in [RFC8032](https://www.rfc-editor.org/rfc/rfc8032#section-5.1).
+Ed25519 is the current default key type.
 
-Implementations MAY support RSA, Secp256k1 and ECDSA for private use, but peers
+Implementations SHOULD support RSA if they wish to interoperate with legacy
+IPNS names (RSA was used before Ed25519).
+
+Implementations MAY support Secp256k1 and ECDSA for private use, but peers
 from the public IPFS swarm and DHT may not be able to resolve IPNS records
-signed by these optional key types. When implementing support for these optional key
-types, follow signature implementation notes from [PeerID specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types).
+signed by these optional key types.
+
+When implementing support for key types, follow signature implementation notes
+from [PeerID specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types).
 
 In all cases, the IPNS implementation MAY allow the user to enable/disable specific key types via configuration. Note that disabling support for compulsory key type will hinder IPNS interop.
 
--- a/IPNS.md
+++ b/IPNS.md
@@ -145,9 +145,12 @@ A logical IPNS record is a data structure containing the following fields:
   - Extensible record data in [DAG-CBOR](https://ipld.io/specs/codecs/dag-cbor/spec/) format.
   - The default set of fields can be augmented with additional information.
     - Implementations are free to leverage this, or simply ignore unexpected fields.
-    - A good practice is to prefix custom field names with `_` to avoid
-      collisions with any new mandatory fields that may be added in a future
-      version of this specification.
+    - A good practice is to:
+      - prefix custom field names with `_` to avoid collisions with any new
+        mandatory fields that may be added in a future version of this
+        specification.
+      - and/or create own namespace by setting value to DAG-CBOR:
+        `IpnsEntry.data[_namespace][customfield]`.
 
 IPNS records are stored locally, as well as spread across the network, in order to be accessible to everyone.
 
--- a/IPNS.md
+++ b/IPNS.md
@@ -283,6 +283,8 @@ Record's data and signature verification MUST be implemented as outlined below, 
 6. Create bytes for signature verification by concatenating `ipns-signature:` prefix (bytes in hex: `69706e732d7369676e61747572653a`) with raw CBOR bytes from `IpnsEntry.data`
 7. Verify signature in `IpnsEntry.signatureV2` against concatenation result from the previous step.
 
+Value in `IpnsEntry.signatureV1` MUST be ignored.
+
 ## Integration with IPFS
 
 Below are additional notes for implementers, documenting how IPNS is integrated within IPFS ecosystem.
--- a/IPNS.md
+++ b/IPNS.md
@@ -190,7 +190,10 @@ message IpnsEntry {
 
 Notes:
 
-- The maximum size of `IpnsEntry` is 2 MiB. Bigger records MUST be ignored by IPNS implementations.
+- IPNS implementations must support sending and receiving serialized
+  `IpnsEntry` of size less or equal 10 kiB. Handling records larger than 10 kiB
+  is not recommended so as to keep compatibility with implementations
+  and transports which only support up to 10 kiB.
 
 - For legacy reasons, some values must be stored in both `IpnsEntry` protobuf and `IpnsEntry.data` CBOR.
   This should not be ignore, as it impact interoperability with old software.
@@ -237,13 +240,14 @@ Creating a new IPNS record MUST follow the below steps:
    - This step SHOULD be skipped for Ed25519, and any other key types that are inlined inside of [IPNS Name](#ipns-name) itself.
 6. Create bytes for signing by concatenating `ipns-signature:` prefix (bytes in hex: `69706e732d7369676e61747572653a`) with raw CBOR bytes from `IpnsEntry.data`
 7. Sign concatenated bytes from the previous step using the private key, and store the signature in `IpnsEntry.signatureV2`
+8. Confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
 
 ### Record Verification
 
 Implementations MUST resolve IPNS Names using only verified records.
 Record's data and signature verification MUST be implemented as outlined below, and fail on the first error.
 
-1. Before parsing the protobuf, confirm that `IpnsEntry` is less than 2MiB in size
+1. Before parsing the protobuf, confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
 2. Confirm `IpnsEntry.signatureV2` and `IpnsEntry.data` are present and are not empty
 3. Extract public key
    - Use `IpnsEntry.pubKey` or a cached entry in the local key store, if present.
--- a/IPNS.md
+++ b/IPNS.md
@@ -24,6 +24,7 @@ IPNS records provide cryptographically verifiable, mutable pointers to objects.
   - [String Representation](#string-representation)
 - [IPNS Record](#ipns-record)
   - [Record Serialization Format](#record-serialization-format)
+  - [Record Size Limit](#record-size-limit)
 - [Protocol](#protocol)
   - [Overview](#overview)
   - [Record Creation](#record-creation)
@@ -194,13 +195,17 @@ message IpnsEntry {
 
 Notes:
 
-- IPNS implementations must support sending and receiving serialized
-  `IpnsEntry` of size less or equal 10 kiB. Handling records larger than 10 kiB
-  is not recommended so as to keep compatibility with implementations
-  and transports which only support up to 10 kiB.
-
 - For legacy reasons, some values must be stored in both `IpnsEntry` protobuf and `IpnsEntry.data` CBOR.
-  This should not be ignore, as it impact interoperability with old software.
+  This should not be ignored, as it impacts interoperability with old software.
+
+### Record Size Limit
+
+IPNS implementations MUST support sending and receiving a serialized
+`IpnsEntry` less than or equal to **10 KiB** in size.
+
+Records over the limit MAY be ignored. Handling records larger than the
+limit is not recommended so as to keep compatibility with implementations and
+transports that follow this specification.
 
 ## Protocol
 
@@ -244,14 +249,14 @@ Creating a new IPNS record MUST follow the below steps:
    - This step SHOULD be skipped for Ed25519, and any other key types that are inlined inside of [IPNS Name](#ipns-name) itself.
 6. Create bytes for signing by concatenating `ipns-signature:` prefix (bytes in hex: `69706e732d7369676e61747572653a`) with raw CBOR bytes from `IpnsEntry.data`
 7. Sign concatenated bytes from the previous step using the private key, and store the signature in `IpnsEntry.signatureV2`
-8. Confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
+8. Confirm that the serialized `IpnsEntry` bytes sum to less than or equal to [the size limit](#record-size-limit).
 
 ### Record Verification
 
 Implementations MUST resolve IPNS Names using only verified records.
 Record's data and signature verification MUST be implemented as outlined below, and fail on the first error.
 
-1. Before parsing the protobuf, confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
+1. Before parsing the protobuf, confirm that the serialized `IpnsEntry` bytes sum to less than or equal to [the size limit](#record-size-limit).
 2. Confirm `IpnsEntry.signatureV2` and `IpnsEntry.data` are present and are not empty
 3. Extract public key
    - Use `IpnsEntry.pubKey` or a cached entry in the local key store, if present.
--- a/IPNS.md
+++ b/IPNS.md
@@ -142,7 +142,11 @@ A logical IPNS record is a data structure containing the following fields:
   - Implementations MUST include this value in `IpnsEntry.signatureV2` and follow signature creation and verification as described in [Record Creation](#record-creation) and [Record Verification](#record-verification).
 - **Extensible Data** (DAG-CBOR)
   - Extensible record data in [DAG-CBOR](https://ipld.io/specs/codecs/dag-cbor/spec/) format.
-  - The default set of fields can be augmented with additional information. Implementations are free to leverage this, or simply ignore unexpected fields.
+  - The default set of fields can be augmented with additional information.
+    - Implementations are free to leverage this, or simply ignore unexpected fields.
+    - A good practice is to prefix custom field names with `_` to avoid
+      collisions with any new mandatory fields that may be added in a future
+      version of this specification.
 
 IPNS records are stored locally, as well as spread across the network, in order to be accessible to everyone.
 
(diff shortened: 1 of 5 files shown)


Winterhuman:
Just to confirm, the spec states "Before parsing protobuf, confirm IpnsEntry size is less than 2 MiB", however, will IPFS implementations (specifically Kubo) likely only support smaller IPNS records? Is there a minimum max size of IPNS record that all implementations must support?

IPNS.md
```diff
@@ -91,24 +178,59 @@ As soon as the node has the most recent record, the signature and the validity m
...
+   - This step SHOULD be skipped for Ed25519 keys.
+6. Create bytes for signing by concatenating `ipns-signature:` prefix (bytes in hex: `69706e732d7369676e61747572653a`) with raw CBOR bytes from `IpnsEntry.data`
+7. Sign bytes from the previous step using the private key, and store the signature in `IpnsEntry.signatureV2`
+
+### Record Validation
+
+Implementations MUST resolve IPNS Names using only verified records.
+Record's data and signature verification MUST be implemented as outlined below, and fail on the first error.
+
+1. Before parsing protobuf, confirm `IpnsEntry` size is less than 2 MiB
```
Winterhuman, line 203: Maybe "Before parsing the protobuf, confirm that `IpnsEntry` is less than 2MiB in size" would be better wording?

lidel:
```suggestion
1. Before parsing the protobuf, confirm that `IpnsEntry` is less than 2MiB in size
```

lidel:
@Winterhuman there is no  explicit max size of IPNS record defined anywhere, that is why I added explicit number to this spec (so we can implement the same thing across implementations).
 
The only reason I picked 2 MiB is to have the same number as the bitswap block limit – see https://github.com/ipfs/kubo/pull/8968

Winterhuman:
Ah I see! I do think a minimum max size is needed, it'd be nice to have a safe amount of data that can be inlined into IPNS records (by `Value` and/or by `data`) without any risk of being rejected by specific implementations, even like 4KiB or something would be good just so it's a definitive spec-defined size

lidel:
Pushed some changes, expanding separate sections about "IPNS Name" and "IPNS Keys", formally decoupling IPNS spec from PeerIDs.

IPNS.md
```diff
@@ -29,95 +36,258 @@ Each time a file is modified, its content address changes. As a consequence, the
...
-  - Public key used to sign this record
-  - Note: The public key **must** be included if it cannot be extracted from the peer ID (reference [libp2p/specs#100](https://github.com/libp2p/specs/pull/100/files)).
-- 7. **ttl** (uint64)
-  - A hint for how long the record should be cached before going back to, for instance the DHT, in order to check if it has been updated.
+### Key Types
+
+Implementations MUST support Ed25519 with signatures defined in [RFC8032](https://www.rfc-editor.org/rfc/rfc8032#section-5.1).
+
+Implementations MAY support RSA, Secp256k1 and ECDSA for private use, but peers
+from the public IPFS swarm and DHT may not be able to resolve IPNS records
```
aschmahmann, line 46: Seems mostly reasonable as this copies from the libp2p peerID spec, however RSA keys might still be very common since they used to be the default in implementations like kubo (formerly go-ipfs) and js-ipfs. This is the same as in the libp2p world, but given we're not just referencing that spec directly figured I'd call it out here in case anyone comes looking.

It'd be nice if we could post some numbers here (on GitHub) coming from the Hydra nodes on the distribution of key types justifying that Ed25519 covers most usage.

IPNS.md
```diff
@@ -29,95 +36,258 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+- **Public Key** (bytes)
+  - Public key used to sign this record.
+    - If public key is small enough to fit in IPNS name (e.g., Ed25519 keys inlined using `identity` multihash), `IpnsEntry.pubKey` field is redundant and MAY be skipped to save space.
+    - The public key MUST be included if it cannot be extracted from the IPNS name (e.g., legacy RSA keys). Implementers MUST follow key serialization defined in [PeerID specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types).
+- **Signature** (bytes)
+  - Provides the cryptographic proof that the IPNS record was created by the owner of the private key.
+  - Implementations MUST include this value in `IpnsEntry.signatureV2` and follow signature creation and verification as described in [Record Creation](#record-creation) and [Record Verification](#record-verification).
+- **Extensible Data** (DAG-CBOR)
+  - Extensible record data in [DAG-CBOR](https://ipld.io/specs/codecs/dag-cbor/spec/) format.
+  - The default set of fields can be augmented with additional information. Implementations are free to leverage this, or simply ignore unexpected fields.
```
aschmahmann, line 145: Probably obvious, but might be worth flagging that any new fields that aren't added to the spec could run into collisions and/or misinterpretations (e.g. like custom HTTP headers).

Also, technically you can add more data to the protobuf as well unless we're speccing that implementations should be stricter in how they interpret the protobufs (e.g. no adding field number 100 into the protobuf) it's just not covered by the signature so not something most users would want.

Allowing for new protobuf fields does give us some extensibility if we want to modify the format more into the future though, so probably worth keeping that around so long as there are record size limits in place.

aschmahmann (commented):
😍 at this writeup. I didn't read through things too precisely for correctness, but left a few comments that might be useful.

IPNS.md
```diff
@@ -29,95 +36,258 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+ // extensible record data in DAG-CBOR format
+ optional bytes data = 9;
 }
 ```
 
-## Protocol
+Notes:
 
-Taking into consideration a p2p network, each peer should be able to publish IPNS records to the network, as well as to resolve the IPNS records published by other peers.
+- The maximum size of `IpnsEntry` is 2 MiB. Bigger records MUST be ignored by IPNS implementations.
```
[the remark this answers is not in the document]
lidel: Things are more complicated: the introduction of the extensible `IpnsEntry.data` CBOR field means some people and services WILL start putting additional metadata there, and that means all bets and estimates will be off.

Still, we need to create some arbitrary headroom. It is easier to increase the limit than to decrease it.
I've switched to max to 10 kiB as it is closer to realistic needs, and adjusted the language to match bitswap one (see 1baf069)

@alanshaw  / @vasco-santos do you have any metrics around record sizes on w3name?  or thoughts on this matter?

IPNS.md
lidel: Will try to get %  usage of Ed25519, RSA, Secp256k1 and ECDSA and get back to this with numbers.

IPNS.md
lidel: Added lower size limit https://github.com/ipfs/specs/pull/319/commits/cc5941b55ee0af7f05f83cb84e9451dd786f3842, think this is easier from implementation perspective (does not require custom protobuf implementation, and the size check of entire thing is very inexpensive).

IPNS.md
```diff
@@ -29,95 +36,266 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+ optional bytes data = 9;
 }
 ```
 
-## Protocol
+Notes:
 
-Taking into consideration a p2p network, each peer should be able to publish IPNS records to the network, as well as to resolve the IPNS records published by other peers.
+- IPNS implementations must support sending and receiving serialized
+  `IpnsEntry` of size less or equal 10 kiB. Handling records larger than 10 kiB
```
Winterhuman, line 198: Maybe "IPNS implementations must support sending and receiving a serialized `IpnsEntry` less than or equal to 10KiB."?

IPNS.md
```diff
@@ -29,95 +36,266 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+   - This is paramount: this CBOR will be used for signing.
+3. Store DAG-CBOR in `IpnsEntry.data`.
+   - If you want to store additional metadata in the record, add it under unique keys at `IpnsEntry.data`.
+   - The order of fields impacts signature verification. If you are using an alternative CBOR implementation, make sure the CBOR field order follows [RFC7049](https://www.rfc-editor.org/rfc/rfc7049) sorting rules: length and then bytewise. The order of fields impacts signature verification.
+4. If your public key can't be inlined inside the IPNS Name, include a serialized copy in `IpnsEntry.pubKey`
+   - This step SHOULD be skipped for Ed25519, and any other key types that are inlined inside of [IPNS Name](#ipns-name) itself.
+6. Create bytes for signing by concatenating `ipns-signature:` prefix (bytes in hex: `69706e732d7369676e61747572653a`) with raw CBOR bytes from `IpnsEntry.data`
+7. Sign concatenated bytes from the previous step using the private key, and store the signature in `IpnsEntry.signatureV2`
+8. Confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
```
Winterhuman, line 247: Maybe "Confirm that the serialized `IpnsEntry` bytes sum to less than or equal to 10KiB in size"?

IPNS.md
```diff
@@ -29,95 +36,266 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+
+### Record Verification
+
+Implementations MUST resolve IPNS Names using only verified records.
+Record's data and signature verification MUST be implemented as outlined below, and fail on the first error.
+
+1. Before parsing the protobuf, confirm that bytes with serialized `IpnsEntry` are less than or equal 10 kiB in size
```
Winterhuman, line 254: Maybe "Before parsing the protobuf, confirm that the serialized `IpnsEntry` bytes sum to less than or equal to 10KiB in size"?

IPNS.md
Winterhuman: @lidel Maybe the spec should recommend applications make their own field with sub-fields under it, e.g. `IpnsEntry.data.my-app.field`?

IPNS.md
lidel: Thanks. I applied your suggestions + created `Record Size Limit` section so we have single place with the value: https://github.com/ipfs/specs/pull/319/commits/1a8f0c908f7ada0a6164bdbfe76b1354355b8953

IPNS.md
lidel: Thanks, added best practices in https://github.com/ipfs/specs/pull/319/commits/78c9f045cb79ed7587ec8ed889dc2d90e4b486f4

IPNS.md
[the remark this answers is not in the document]
lidel: Some quick numbers from hydras (when the validity is at max 7d in the past)

```
Total DynamoDB Records: 394585
  ipns: 150175
  pk:   244410
  
PK Records Key Types:
  rsa:       244403
  ed25519:   3
  secp256k1: 0
  ecdsa:     4

DynamoDB Record Value Sizes:
  avg:    318.08328497028526
  std:    127.86082973202544
  median: 299
  max:    3350
  min:    36
```

And just the extensible data field: 
```
IPNS 'IpnsEntry.Data' Size:
  avg:    139.27228998453873
  std:    9.81409166725822
  median: 141
  max:    182
  min:    91
```

10 KiB sounds ok, leaving some room for growth and experimentation without being blocked by implementations that follow this spec, but lmk if we should adjust.

IPNS.md
lidel: Some quick numbers from hydras (records that are currently valid)

```
Total DynamoDB Records: 256697
  ipns: 12287
  pk:   244410

DynamoDB Record Value Sizes:
  avg:    299.62269913555673
  std:    45.88040769545226
  median: 299
  max:    2033
  min:    36
IPNS Records Signatures:
  v1:     12287
  v2:     6892
  with pk: 913
PK Records Key Types:
  rsa:       244403
  ed25519:   3
  secp256k1: 0
  ecdsa:     4
```
ed25519 is used by default and inlined, "3" is probably someone writing own  IPNS tool and embedding it the same way RSA - can be ignored.

“with pk” means IPNS records with IpnsEntry.pubKey present.
RSA  is dominating here, and used in ~7% of cases. 

I think Ed25519 being ~93%  and RSA ~7% is a good argument to keep Ed25519 as MUST and change RSA to SHOULD (same as in PeerID spec). 

I adjusted language in https://github.com/ipfs/specs/pull/319/commits/41612da4bebec847612e3efcce83094205903f63

IPNS.md
```diff
@@ -29,95 +38,302 @@ Each time a file is modified, its content address changes. As a consequence, the
...
+Implementations that want to interop with the public IPFS swarm MUST maintain
+backward compatibility for legacy consumers of IPNS records:
 
-Finally, the network nodes may also republish their records, so that the records in the network continue to be valid to the other nodes.
 
-## Overview
+- A legacy publisher MUST always be able to update to the latest implementation
+  of this specification without breaking record resolution for legacy consumers.
+- A legacy consumer MUST always be able to resolve IPNS name, even when publisher
+  updated to the latest implementation of this specification.
```
lidel, line 229: ## Some DHT  stats

We've looked at some stats from Hydra nodes (a small sample of recently valid records):
```
IPNS Records Signatures:
  v1:       9513
  v2:       5821
  with pk:  733 (RSA)
  nil data: 3692 (V1-only)
```

Every V2 record is also V1 (V1-only records have no data field, so their count is in  "nil data").
This means for the sample size of 9513 we have ~61% of V2+V1 and ~38% V1-only. 
We also looked at other time-ranges and V2 is growing, V1-only is decreasing – seems to follow the usual slow update curve we also see for Kubo itself.

## On this upgrade strategy

It is reasonable to expect both clients and  publishers will update over time, 
but the act of updating a publisher should not introduce any record resolution regressions for legacy nodes.

To make this clear, I've documented backward compatibility policy in https://github.com/ipfs/specs/pull/319/commits/b926e8c696a31686c26dfa931632cc307b411df5

> - A legacy publisher MUST always be able to update to the latest implementation
>   of this specification without breaking record resolution for legacy consumers.
> - A legacy consumer MUST always be able to resolve IPNS name, even when publisher
>   updated to the latest implementation of this specification.

This policy ensures the spec is aligned with existing code that was created to ensure IPNS implementations in GO and JS are backward-compatible from the perspective of legacy consumers:
- both Kubo and js-ipfs produce V1+V2 records
  - This allows legacy go-ipfs <0.9.0 node to correctly resolve IPNS records published by modern Kubo 0.15+
  - Maintaining this contract is important, I've added it to this spec
- Old go-ipfs <0.9.0 node publishing legacy RSA+V1-only records   should be able to upgrade to Kubo 0.16  and start publishing V1+V2 records, allowing for V2 resolution on modern clients (but not break V1-only for legacy ones).

IPNS.md
```diff
@@ -29,95 +38,302 @@ Each time a file is modified, its content address changes. As a consequence, the
...
-As soon as the node has the most recent record, the signature and the validity must be verified, in order to conclude that the record is still valid and not compromised.
+Implementations that want to interop with the public IPFS swarm MUST maintain
+backward compatibility for legacy consumers of IPNS records:
 
-Finally, the network nodes may also republish their records, so that the records in the network continue to be valid to the other nodes.
 
-## Overview
+- A legacy publisher MUST always be able to update to the latest implementation
+  of this specification without breaking record resolution for legacy consumers.
+- A legacy consumer MUST always be able to resolve IPNS name, even when publisher
```
paulgmiller, line 228: does "resolve IPNS" name mean they should be able to validate the signature? Or just that they can read the record?
Isn't that already not the case since supporting all Key types isn't required there's already consumers that can't validate all signtures? And new key types (or signing prefixes) would invalidate old consumers?

lidel: > does "resolve IPNS" name mean they should be able to validate the signature?

@paulgmiller  Yes, IPNS resolution includes signature verification step: records that have invalid signatureV2 are ignored by implementations like Kubo or js-ipfs.

> And new key types (or signing prefixes) would invalidate old consumers?

The use case we are protecting here is preexisting IPNS client resolving preexisting IPNS names for mission-critical updates. In that case, the key type will not change, records are published with the same key, and the old signature must be present next to the new one to ensure the name still resolves for the legacy client.

An example scenario is as follows:
-  IoT device implemented minimal support for RSA and V1, and uses it for polling for new firmware updates.
- It retrieves record with V1+V2 signatures, validates V1 and fetches new firmware
- The new firmware has support for IPNS V2, and uses a new IPNS name for polling for updates (requires V2 signatures, and can also use a different key type like Ed25519).

lidel:
Merging since the hardened IPNS record verification shipped in:
- [go-ipns v0.3.0](https://github.com/ipfs/go-ipns/releases/tag/v0.3.0)
  - included in Kubo 0.16 ([Release Notes](https://github.com/ipfs/kubo/blob/master/docs/changelogs/v0.16.md#%EF%B8%8F-hardened-ipns-record-verification))
- [js-ipns v3.0.0](https://github.com/ipfs/js-ipns/releases/tag/v3.0.0)  
  -  will be  included in next JS-IPFS  (tracked in https://github.com/ipfs/js-ipfs/pull/4221)

PR #322, opened 2022-09-19
IPIP-322: Content Routing Hints

guseggert:
(co-authored by @guseggert  and @lidel )

### DIFF
--- a/CONTENT_ROUTING_HINTS.md
+++ b/CONTENT_ROUTING_HINTS.md
@@ -1,0 +1,96 @@
+# Introduction
+
+Content routing maps a CID to one or more *providers*, which specify locations where the CID can be fetched. 
+
+In order to fetch content-addressed data, there must be *some* location addressing involved. With IPFS, the implicit default starting point is a set of bootstrap nodes. (And perhaps some LAN nodes discovered by mDNS, which has a starting point of the local subnet.)
+
+So far, Kubo has planned to keep this “location addressing” implicit by adding new content routers to the default Kubo config (e.g. Filecoin indexers). But this only solves the problem for whatever specific records are provided by that indexer, and those set of implicit content routers have to be supported by the various implementations to maintain the facade of “pure content addressing”. There are also trust issues in terms of automatically sending user data to indexers that users have not explicitly trusted.
+
+Instead of gateways and IPFS nodes implicitly sending all requests to a set of content routers that changes over time, and the community needing to reach consensus on what default routers to use, this proposes specifying that the default implicit content router is *only* the IFPS public DHT and LAN DHT, and all additional content routers must be opted-in by users when making API requests.
+
+# Specification
+
+The default implicit content router for IPFS nodes is the IPFS public DHT and LAN DHT. Any additional content routers must be opted-in by users when making API requests.
+
+Users may opt-in to additional content routers using “content routing hints”, which give *suggestions* to the IPFS node about where provider records for the given CID may be found. This can include, but is not limited to, Reframe URLs, pubsub topics, multiaddrs, etc. As hints, the IPFS node is free to decide the order and strategy for using hints. If an IPFS node implements support for a hint that is specified below, it must follow the specification for that hint type.
+
+When a node receives a request with content routing hints, it should search for provider records in the IPFS public DHT and at locations specified in the hints.
+
+## Hint Types
+
+Implementations are free to support hint types that make sense for their use cases.
+
+### URI
+
+- **Reframe**
+    - HTTPS URL that ends with `/reframe` MUST be interpreted as a Reframe hint, for example:
+        - [`https://cid.contact/reframe`](https://cid.contact/reframe)
+        - [`https://routing.delegate.ipfs.io/reframe`](https://routing.delegate.ipfs.io/reframe)
+- **Magnet links (TBD, for consideration)**
+    - “De facto standard” outside IPFS: [https://en.wikipedia.org/wiki/Magnet_URI_scheme](https://en.wikipedia.org/wiki/Magnet_URI_scheme)
+- **HTTP mirror  AKA Web seed (TBD, for consideration)**
+    - We could speed up data transfer of leaf nodes by making HTTP range-requests to the provided HTTP URL.
+    - Bit out there, but could provide additional flexibility, especially when a URL of a public gateway is used.
+
+### Multiaddr (instant win)
+
+Multiaddrs alone provide a very flexible solution for routing hints. They enable control over the number of additional  lookups that a client needs to make to reach the data:
+
+- `/ip4/A.B.C.D/tcp/NNN/p2p/{peerID}`
+    - Removes need for any lookups, can try to connect directly and start data transfer
+- `/p2p/{PeerID}`
+    - Saves 1 DHT lookup - we already know potential provider’s PeerID, only need to find their addresssed via `findpeer` (or similar)
+        - On gateways this is highly cacheable
+- `/dnsaddr/{domain}`
+    - Requires resolving [DNSAddr TXT records](https://github.com/multiformats/multiaddr/blob/master/protocols/DNSADDR.md) on DNS, but allows big content storage services to scale / load-balance with ease, leverage DNS-based delegation to nodes that have data and are the closest
+    - Could include fully resolved addresses, PeerIDs, or another DNSAddrs (with some sane recursion limit, could be the same as for resolving /ipns/ paths – 32)
+    - Allows us to collapse a lot of complexity into a single DNS-based hint
+        - 💡IDEA: we could implicitly check for DNSaddr on domains that have DNSLink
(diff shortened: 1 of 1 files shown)


CONTENT_ROUTING_HINTS.md
```diff
@@ -0,0 +1,96 @@
...
+When a node receives a request with content routing hints, it should search for provider records in the IPFS public DHT and at locations specified in the hints.
+
+## Hint Types
+
+Implementations are free to support hint types that make sense for their use cases.
+
+### URI
+
+- **Reframe**
+    - HTTPS URL that ends with `/reframe` MUST be interpreted as a Reframe hint, for example:
```
guseggert, line 26: How do we specify different codecs like `dag-cbor`? Do we need to add this to the Reframe spec for HTTP URLs? And what about non-HTTP transports? Would that need to be something we add to multiaddr?

lidel: > How do we specify different codecs like dag-cbor?

That is up to HTTP client (who controls `Accept` header sent with the request).

I think details of `/reframe` over HTTP are specified in [reframe/REFRAME_HTTP_TRANSPORT.md](https://github.com/ipfs/specs/blob/main/reframe/REFRAME_HTTP_TRANSPORT.md).
It already requires endpoint to be named `/reframe`, but we could add a paragraph to that spec which states is more explicitly. 

> what about non-HTTP transports? Would that need to be something we add to multiaddr?


We probably want to register  `/reframe` protocol at https://github.com/multiformats/multiaddr/blob/master/protocols.csv, allowing for multiaddrs like:

```
/dns4/example.net/tcp/443/tls/http/reframe
/dns4/example.net/tcp/443/wss/reframe
```

(and looping in libp2p team for sanity check, if this is correct way of representing this)

CONTENT_ROUTING_HINTS.md
```diff
@@ -0,0 +1,96 @@
...
+
+So far, Kubo has planned to keep this “location addressing” implicit by adding new content routers to the default Kubo config (e.g. Filecoin indexers). But this only solves the problem for whatever specific records are provided by that indexer, and those set of implicit content routers have to be supported by the various implementations to maintain the facade of “pure content addressing”. There are also trust issues in terms of automatically sending user data to indexers that users have not explicitly trusted.
+
+Instead of gateways and IPFS nodes implicitly sending all requests to a set of content routers that changes over time, and the community needing to reach consensus on what default routers to use, this proposes specifying that the default implicit content router is *only* the IFPS public DHT and LAN DHT, and all additional content routers must be opted-in by users when making API requests.
+
+# Specification
+
+The default implicit content router for IPFS nodes is the IPFS public DHT and LAN DHT. Any additional content routers must be opted-in by users when making API requests.
```
aschmahmann, line 13: This section seems independent of the rest of the specification around specifying routing hints. I suspect it's also the most controversial given prior resistance to even defining a single or collection of routing systems as the "default/standard IPFS content routing systems".

This happens to be how the most popular and oldest IPFS implementations (e.g. kubo) have been operating over the last several years, but it could reasonably change over time. Overall this seems to be related to an independent discussion on what should be "required" for an IPFS implementation and/or if/how we should label a collection of protocols and properties that some IPFS implementations have that will make systems easier to reason about (e.g. Bitswap 1.2.0,  IPFS Public DHT, libp2p with some set of transports and upgraders, etc.).

I'd try and separate this long requested and likely quite useful issue from the more nebulous "what is IPFS" kinds of conversations. Although that discussion seems like a separate important one to have and document the outcome of so it can be referenced and/or modified in the future.

CONTENT_ROUTING_HINTS.md
```diff
@@ -0,0 +1,96 @@
...
+
+Users may opt-in to additional content routers using “content routing hints”, which give *suggestions* to the IPFS node about where provider records for the given CID may be found. This can include, but is not limited to, Reframe URLs, pubsub topics, multiaddrs, etc. As hints, the IPFS node is free to decide the order and strategy for using hints. If an IPFS node implements support for a hint that is specified below, it must follow the specification for that hint type.
```
aschmahmann, line 15: Allowing for content routing hints seems fine, but IMO doing this comes with some items that may need addressing:
- Systems that try and resolve IPFS content will likely need to be able to override user inputs to add/subtract content routers.
   - For example, I could see IPFS companion wanting to have a configuration for auto-appending content routing systems, so that a favorite system is always checked, or for limiting them to control which types of routing systems are permissible
- Gateways might need to be able to return errors or status indicating what types of content routing systems can be used. e.g. if the gateway has blocked use of multiaddr hints but the user passed them and got some timeout error it should probably know that the multiaddr hint wasn't allowed to be used.
- More documentation and modification to tools like https://check.ipfs.network/ to help people understand the implications of using new routing systems and how different systems interoperate.

aschmahmann (commented):
Great to see this being worked on and some proposed syntax and formats. As linked in the document it's been a fairly long time request (I think perhaps even longer than those linked issues).

ajnavarro:
Nice ideas.

About hints using Reframe, I cannot see any limitations about what requests can be made to a specific Reframe endpoint if it came with a particular CID. 

Maybe we should specify that we only can make requests to that reframe endpoint if they are related to the "root" CID, to avoid flooding small indexers with unnecessary requests.

CONTENT_ROUTING_HINTS.md
```diff
@@ -0,0 +1,96 @@
...
+- Even if this type of router is disabled by default, we could leverage the fact that `?providers=/dnsaddr/{domain}` is passed and create one.
+    - Nodes could join a topic based on the DNS name from DNSaddr, allowing peers interested in the content from the same provider to exchange data directly over PubSub, skipping DHT or centralized Reframe endpoint.
+- A variant of this that is especially powerful. s when browsing DNSLink website or IPNS name. Mutable pointer would ensure people having old and new version of
+
+## Gateway Requests
+
+We would add support for an optional `?providers=` URL parameter ([percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding), comma-separated) or HTTP header `X-Ipfs-Providers` sent with HTTP request to a gateway.
+
+- `/ipfs/{cid}?providers=url,multiaddr,somethingelse?`
+    - Example:  `https://dweb.link/ipfs/bafy..acbd?providers=/dnsaddr/storage-provider1.com`
```
willscott, line 68: In designing both delegated routing and the content routing setup, I thought the goal was to try to maintain the property we have today of *content addressed data*.

if we have to specify the origin for the data, we've lost some of this property.

I guess there's a benefit of discovery of content routers through this mechanism, but I would hope that's not the only way we learn about content routers in kubo.

CONTENT_ROUTING_HINTS.md
```diff
@@ -0,0 +1,96 @@
...
+- `X-Ipfs-Providers: url,multiaddr,somethingelse`
+    - Example:  `X-Ipfs-Providers: /dnsaddr/storage-provider1.com`
+- Gateways will be free to leverage this hint to speed up content routing, or ignore it.
+- Allows public gateways to load content from services that do not announce CIDs on DHT (e.g.,  Pinata).
+
+## API Requests
+
+We would add optional `--providers` parameter, that allows for passing as-hoc hints that are scoped to specific command.
```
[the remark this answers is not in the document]
lidel: Yes, we will move this to "Notes for implementers" section as an example of CLI API.

CONTENT_ROUTING_HINTS.md
lidel: > if we have to specify the origin for the data, we've lost some of this property.

We share the concern, and that is why this proposal.

We will lose it for sure if a public gateway decides to use a specific set of indexers, and to not use others.
A user may know that their data is in place X, but will be forced to use a specific indexer Y because that is the only thing the gateway speaks.

> would hope that's not the only way we learn about content routers in kubo.

This proposal  is not replacing [`Routing.Routers`](https://github.com/ipfs/kubo/blob/master/docs/config.md#routingrouters-parameters), but compliments it. Gateways should not be gatekeepers. We want to see multiple indexers, and each indexer should not have to lobby for being included on some default lists to provide utility to users. 

By giving users the ability to pass additional routing hints, we remove the surface for undesired storage and content routing lock-in caused by [the tyranny of the default](https://en.wikipedia.org/wiki/Tyranny_of_small_decisions).

Gateway operators will be in control to set any custom peering and routing they wish, but the users should still be able to improve routing even further by providing an optional hint that the gateway/node can leverage for finding providers when the CID can't be found using conventional methods.  This will be course-correcting any routing gaps that may occur.

(Note to self to incorporate this into the spec)

BigLep:
2022-10-04 conversation: turn this spec into a IPIP for gateway spec as a quality of life item for gateways.

BigLep:
Closing this given it has been open for 6 months but there hasn't been traction on it and I'm not seeing anything in short term that will be pushing on it.  It can be reopened if/when there is a push for it.

PR #326, opened 2022-09-29
fix: document GET /reframe?q=

lidel:
This PR is a small fix that aligns Reframe specs with existing Implementation.

[Percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding) DAG-JSON sent over `?q=`  is what Kubo 0.16.0-rc1 uses,  and what is implemented in https://github.com/ipfs/go-delegated-routing v0.6.0.

I decided to add this to the spec, because it saves us time and does not block Kubo/Indexer work (we don't need to change the wire format).

It is a safe path: makes the spec both **backward-compatible** (with Kubo and go-delegated-routing) and **forward-compatible** (with proper CBOR support proposed in separate IPIP from https://github.com/ipfs/specs/pull/327).

cc @guseggert @petar @willscott

### DIFF
--- a/reframe/REFRAME_HTTP_TRANSPORT.md
+++ b/reframe/REFRAME_HTTP_TRANSPORT.md
@@ -30,10 +30,6 @@ serialization mechanisms for sending Reframe messages over HTTP `POST` and
 All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
 
 Requests MUST be sent as either:
-- `GET /reframe/{mbase64url-dag-cbor}`
-  - Cachable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
-    - DAG-CBOR in multibase `base64url` is used instead of DAG-JSON because JSON may include characters that are not safe to be used in URLs, and re-encoding JSON in base would take too much space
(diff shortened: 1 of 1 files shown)


willscott (approved):
can confirm that this appears to reflect what kubo is currently doing.

reframe/REFRAME_HTTP_TRANSPORT.md
```diff
@@ -30,10 +30,9 @@ serialization mechanisms for sending Reframe messages over HTTP `POST` and
 All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit content type `application/vnd.ipfs.rpc+dag-json; version=1`
 
 Requests MUST be sent as either:
-- `GET /reframe/{mbase64url-dag-cbor}`
-  - Cachable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
-    - DAG-CBOR in multibase `base64url` is used instead of DAG-JSON because JSON may include characters that are not safe to be used in URLs, and re-encoding JSON in base would take too much space
-  - Suitable for sharing links, sending smaller messages, and when a query result MUST benefit from HTTP caching (see _HTTP Caching Considerations_ below).
```
lidel, line 36: :information_source:  Removing this for now since there is no implementation. This is re-introduced in IPIP which adds proper DAG-CBOR support (https://github.com/ipfs/specs/pull/327)

BigLep (approved):
Bummer we're in this situation, but your path forward makes sense here @lidel .  Feel free to ship (although lint markdown check is failing)

PR #327, opened 2022-09-29
IPIP-327: Reframe over HTTP version=2 (DAG-CBOR and better cache controls)

lidel:
#### HTTP Caching problems

- single endpoint makes it hard to scale  (have different methods handled by different services)
- Etag generation requires buffering response, effectively breaking streaming 

#### On missing CBOR 

We've b een using Reframe in Kubo for a while and it is clear that Reframe messages are not designed to be created or read by humans.
 
(My subjective opinion) The plaintext  DAG-JSON representation of messages do not really bring  much to the table, because both CIDs and Multiaddrs are in a format that needs manual encoding/decoding anyway, but are still useful during testing ad debugging.

JSON is just bad production format – we are missing CBOR.

####  Why CBOR?

We already support DAG-JSON via a dedicated content type, and our stack uses CBOR in more places than JSON. 

#### Size savings 

Some numbers in https://arxiv.org/abs/2201.03051:

>  

While companies are free to pay an unnecessary ~20% penalty, we should not create protocols which expect end users to do the same.

### (1) Add CBOR format

The improvement proposed here is to add support for requests and responses sent as DAG-CBOR,
with own content type: `application/vnd.ipfs.rpc+dag-cbor`. 

This way Reframe over HTTP would support both DAG-JSON and DAG-CBOR, and it is up to the client to decide what wire format makes sense to them.

### (2) improve HTTP caching

While at it, we've improved the way HTTP caching works, removed the requirement of Etag which caused streaming issues, and added method names to URL paths, to improve scaling / routing on reverse proxies / CDNs.
 
For details, see changes  made to `reframe/REFRAME_HTTP_TRANSPORT.md`

cc @guseggert @petar @willscott

### DIFF
--- a/IPIP/0000-http-reframe-cbor.md
+++ b/IPIP/0000-http-reframe-cbor.md
@@ -29,7 +29,7 @@ We already support DAG-JSON, with its own content type.
 The change here is to add support for requests and responses sent as DAG-CBOR,
 with own content type: `application/vnd.ipfs.rpc+dag-cbor`.
 
-For details, see changes  made to `reframe/REFRAME_HTTP_TRANSPORT.md`
+For details, see changes made to `reframe/REFRAME_HTTP_TRANSPORT.md`.
 
 ## Test fixtures
 
@@ -38,9 +38,9 @@ TODO: add CIDs of sample DAG-CBOR messages after https://github.com/ipfs/go-dele
 ## Design rationale
 
 IPFS stack aims to support both DAG-CBOR and DAG-JSON. Users can store JSON as
-CBOR and vice versa. Having consitent support for both in Reframe not only
+CBOR and vice versa. Having consistent support for both in Reframe not only
 aligns with user expectations, but also allows us to save some bytes
-(bandwidth, response caching requirements) by using a binary CBOR as  the
+(bandwidth, response caching requirements) by using a binary CBOR as the
 production format.
 
 ### User benefit
@@ -63,7 +63,7 @@ N/A, we will use the same DAG-CBOR encoder/decoder as the rest of the stack.
 
 Alternative is to do nothing, and end up with:
 
-- inconsitent user experience
+- inconsistent user experience
 - wasted bandwidth and cache storage
 
 ### Copyright
--- a/IPIP/0000-http-reframe-cbor.md
+++ b/IPIP/0000-http-reframe-cbor.md
@@ -1,0 +1,71 @@
+# IPIP 0000: Add DAG-CBOR support to Reframe over HTTP
+
+<!-- IPIP number will be assigned by an editor. When opening a pull request to
+submit your IPIP, please use number 0000 and an abbreviated title in the filename,
+`0000-draft-title-abbrev.md`. -->
+
+- Start Date: 2022-09-29
+- Related Issues:
+  - https://github.com/ipld/edelweiss/issues/16#issuecomment-1074161577
+  - https://github.com/ipfs/kubo/issues/8823
+
+## Summary
+
+<!--One paragraph explanation of the IPIP.-->
+This IPIP adds DAG-CBOR support to Reframe over HTTP.
+
+## Motivation
+
+We've been using Reframe in Kubo for a while and it is clear that Reframe
+messages are not designed to be created or read by humans.
+
+The plaintext  DAG-JSON representation of messages does not really bring
+anything to the table (because both CIDs and Multiaddrs are in a format that
+needs manual encoding/decoding anyway),
+
+## Detailed design
+
+We already support DAG-JSON, with its own content type.
+The change here is to add support for requests and responses sent as DAG-CBOR,
+with own content type: `application/vnd.ipfs.rpc+dag-cbor`.
+
+For details, see changes  made to `reframe/REFRAME_HTTP_TRANSPORT.md`
+
+## Test fixtures
+
+TODO: add CIDs of sample DAG-CBOR messages after https://github.com/ipfs/go-delegated-routing implements it, and has own tests.
+
+## Design rationale
+
+IPFS stack aims to support both DAG-CBOR and DAG-JSON. Users can store JSON as
+CBOR and vice versa. Having consitent support for both in Reframe not only
+aligns with user expectations, but also allows us to save some bytes
+(bandwidth, response caching requirements) by using a binary CBOR as  the
+production format.
+
+### User benefit
+
+User will be able to choose between binary and human-readable representation,
+just like they do in other parts of IPFS/IPLD stack.
+
+### Compatibility
+
+Explain the upgrade considerations for existing implementations.
+
+This IPIP add DAG-CBOR next to already existing DAG-JSON. Preexisting clients
+that only speak DAG-JSON will continue working, no change is required.
+
+### Security
+
+N/A, we will use the same DAG-CBOR encoder/decoder as the rest of the stack.
+
+### Alternatives
+
+Alternative is to do nothing, and end up with:
+
+- inconsitent user experience
+- wasted bandwidth and cache storage
+
+### Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/IPIP/0000-http-reframe-cbor.md
+++ b/IPIP/0000-http-reframe-cbor.md
@@ -21,13 +21,22 @@ messages are not designed to be created or read by humans.
 
 The plaintext  DAG-JSON representation of messages does not really bring
 anything to the table (because both CIDs and Multiaddrs are in a format that
-needs manual encoding/decoding anyway),
+needs manual encoding/decoding anyway), and the utility is limited to debugging
+and use in examples.
+
+We've also identified some HTTP caching and scaling issues due to all methods
+sharing the same URL path and the way `Etag` header is generated, and how
+it made streaming responses impossible.
 
 ## Detailed design
 
 We already support DAG-JSON, with its own content type.
 The change here is to add support for requests and responses sent as DAG-CBOR,
 with own content type: `application/vnd.ipfs.rpc+dag-cbor`.
+
+We change the URL to include method name on the path. This allows deployments
+to scale better: set different HTTP cache control policies, or route different
+methods to different backend services.
 
 For details, see changes made to `reframe/REFRAME_HTTP_TRANSPORT.md`.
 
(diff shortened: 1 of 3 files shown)


lidel:
> do we want to try to support the /reframe/{dag-cbor-encoded-query} mode for cache-able queries at the same time?

@willscott  Yes, the idea is that server will support both JSON and CBOR endpoints. 
Modern clients will most likely only use binary CBOR.

I went with separate  `/reframe/{dag-cbor-encoded-query}` for CBOR because it removes footguns related to content-type or query parameters not being used for caching.

reframe/REFRAME_HTTP_TRANSPORT.md
```diff
@@ -16,36 +16,74 @@ The Reframe over HTTP protocol is defining the transport and message
...
 
+### Other notes
+
 If a server supports HTTP/1.1, then it MAY send chunked-encoded messages. Clients supporting HTTP/1.1 MUST accept chunked-encoded responses.
 
 Requests and Responses MUST occur over a single HTTP call instead of the server being allowed to dial back the client with a response at a later time. The response status code MUST be 200 if the RPC transaction succeeds, even when there's an error at the application layer, and a non-200 status code if the RPC transaction fails.
 
-If a server chooses to respond to a single request message with a group of messages in the response it should do so as a set of `\n` delimited DAG-JSON messages (i.e. `{Response1}\n{Response2}...`).
+If a server chooses to respond to a single request message with a group of DAG-JSON messages in the response it should do so as a set of `\n` delimited DAG-JSON messages (i.e. `{Response1}\n{Response2}...`).
+DAG-CBOR responses require no special handling, as they are already self-delimiting due to the nature of the CBOR encoding.
```
lidel, line 86: :thought_balloon:  This is theory. Before we merge we need to attempt at implementing this in https://github.com/ipld/edelweiss

lidel: https://www.rfc-editor.org/rfc/rfc8742.html states:

> CBOR Sequences, unlike JSON Text Sequences [RFC7464], do not use a marker between items. This is possible because CBOR-encoded data items are self delimiting and the end can always be calculated. (Note that, while the early object/array-only form of JSON was self delimiting as well, this stopped being the case when simple values such as single numbers were made valid JSON documents.)

rvagg: Oh, be very careful with this assumption .. we expect the delimiting to be done before the codec—i.e. you give the codec the exact bytes which match the CID and no more; otherwise we have a validation gap.

The only special-casing we've done that's related to this is that Edelweiss has a special flag that it uses for **dag-json** that lets it be sloppy with ending characters (spaces, EOL): https://github.com/ipld/go-ipld-prime/blob/7548eb883bda4712355797547a0628a0ad1c00cb/codec/dagjson/unmarshal.go#L36-L41 but not for additional data that might delimit the block. dag-cbor is very strict about not wanting extraneous bytes; in Go and JS.

lidel: I mean.. yeah, what other options do we have here? 
We can't use `\n` or any other byte sequence as delimiter because it could be part of CBOR bytes.

IPIP/0000-http-reframe-cbor.md
```diff
@@ -0,0 +1,71 @@
...
+For details, see changes  made to `reframe/REFRAME_HTTP_TRANSPORT.md`
+
+## Test fixtures
+
+TODO: add CIDs of sample DAG-CBOR messages after https://github.com/ipfs/go-delegated-routing implements it, and has own tests.
+
+## Design rationale
+
+IPFS stack aims to support both DAG-CBOR and DAG-JSON. Users can store JSON as
+CBOR and vice versa. Having consitent support for both in Reframe not only
```
guseggert, line 41:
```suggestion
CBOR and vice versa. Having consistent support for both in Reframe not only
```

IPIP/0000-http-reframe-cbor.md
```diff
@@ -0,0 +1,71 @@
...
+aligns with user expectations, but also allows us to save some bytes
+(bandwidth, response caching requirements) by using a binary CBOR as  the
```
guseggert, line 43:
```suggestion
(bandwidth, response caching requirements) by using a binary CBOR as the
```

IPIP/0000-http-reframe-cbor.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Security
+
+N/A, we will use the same DAG-CBOR encoder/decoder as the rest of the stack.
+
+### Alternatives
+
+Alternative is to do nothing, and end up with:
+
+- inconsitent user experience
```
guseggert, line 66:
```suggestion
- inconsistent user experience
```

IPIP/0000-http-reframe-cbor.md
```diff
@@ -0,0 +1,71 @@
...
+- wasted bandwidth and cache storage
```
guseggert, line 67: Overall I think dag-cbor support would be net positive, but I think the perf argument for this is rather weak, there are lower hanging fruit like enabling response compression and we have quite voluminous headers in practice. I'm easy to convince here, it would just take some measurements with real delegated routing payloads using dag-json and dag-cbor encodings. I get the theoretical arguments and the evidence from that paper, but for perf it's still weak without direct measurements of the real workload.

IPIP/0000-http-reframe-cbor.md
```diff
@@ -0,0 +1,71 @@
...
+anything to the table (because both CIDs and Multiaddrs are in a format that
+needs manual encoding/decoding anyway),
+
+## Detailed design
+
+We already support DAG-JSON, with its own content type.
+The change here is to add support for requests and responses sent as DAG-CBOR,
+with own content type: `application/vnd.ipfs.rpc+dag-cbor`.
+
+For details, see changes  made to `reframe/REFRAME_HTTP_TRANSPORT.md`
```
guseggert, line 32:
```suggestion
For details, see changes made to `reframe/REFRAME_HTTP_TRANSPORT.md`.
```

IPIP/0000-http-reframe-cbor.md
lidel: I made quick test with a single `FindProvidersResponse` and we get better gains than the paper suggests:
 
 ```json
 {"FindProvidersResponse":{"Providers":[{"Node":{"peer":{"ID":{"/":{"bytes":"EiAngCqwSSL46hQ5+DWaJsZ1SPV2RwrqwID/OEuj5Rdgqw"}},"Multiaddresses":[{"/":{"bytes":"NhFlbGFzdGljLmRhZy5ob3VzZQYBu94D"}}]}},"Proto":[{"2304":{}}]}]}}
 ``` 

- `baguqeerafsujii5p52wlc3fxrxhmbx6nu7uudyeznqjnvr7ogkjkndkuyukq` 
  DAG-JSON block is 143 bytes

- `bafyreignkd26ejp6jteqf4cggzbhp3twuvidqsa2lshknk6kjh7j2pc2h4` 
  DAG-CBOR block  is 223 bytes
  

The diff is ~35%.   

I think it makes sense, since DAG-JSON eats additional space by base64-encoding binary fields (all CIDs and Multiaddrs), which creates consistent ~33% overhead.

reframe/REFRAME_HTTP_TRANSPORT.md
rvagg: OK, well if you want to delimit by whole-value boundaries then we could work that into the dag-cbor decoder but it'll take a little bit of work and have to be a special option for it. If it's needed, then let's get an issue filed in go-ipld-prime and let me know what sort of priority it is and I'll start having a look.

reframe/REFRAME_HTTP_TRANSPORT.md
```diff
@@ -16,36 +16,74 @@ The Reframe over HTTP protocol is defining the transport and message
...
+
+### HTTP methods
 
 Requests MUST be sent as either:
 
+- `GET /reframe/{mbase64url-dag-cbor}`
+  - Cachable HTTP `GET` requests with message passed as DAG-CBOR in HTTP path segment, encoded as URL-safe [`base64url` multibase](https://docs.ipfs.io/concepts/glossary/#base64url) string
+    - DAG-CBOR in multibase `base64url` is used (even when request body is DAG-JSON) because JSON may include characters that are not safe to be used in URLs, and percent-encoding or base-encoding a big JSON query may take too much space.
+  - Suitable for sharing links, sending bigger messages, and when a query result MUST benefit from HTTP caching (see _HTTP Caching Considerations_ below).
 - `GET /reframe?q={percent-encoded-dag-json}`
```
lidel, line 70: TBD, we've found another discrepancy between spec and implementation – see https://github.com/ipfs/specs/pull/333 and https://github.com/ipld/edelweiss/issues/61

reframe/REFRAME_HTTP_TRANSPORT.md
```diff
@@ -16,36 +16,74 @@ The Reframe over HTTP protocol is defining the transport and message
...
+
+- Requests without a matching `Content-Type` header MAY be interpreted as DAG-JSON.
+- Requests without a matching `Accept` header MAY produce a DAG-JSON response.
+- Responses without a matching `Content-Type` header MAY be interpreted as DAG-JSON.
+
+### HTTP methods
 
 Requests MUST be sent as either:
 
+- `GET /reframe/{mbase64url-dag-cbor}`
```
lidel, line 66: Talked with @guseggert and it would be beneficial to caching / scaling if we had request type in the path – allows for custom routing/caching per method without having to deserialize JSON/CBOR:

```suggestion
- `GET /reframe/{FindProvidersRequest}/{mbase64url-dag-cbor}`

```


TBD is this is possible to implement in https://github.com/ipld/edelweiss, dropping this comment so we don't forget (if we decide to keep Reframe).

lidel: Added in https://github.com/ipfs/specs/pull/327/commits/e1448c2eea5be30c62f86c0e81b94e59fe6cae11

lidel:
Food for thought: https://atproto.com/specs/xrpc#requests (Social networking technology created by Bluesky) uses similar GET / POST distinction, and includes method name on path.

IPIP/0000-http-reframe-cbor.md
BigLep: @lidel : these numbers are uncompressed right?  Aren't the numbers we need to compare when compression is employed since that is how the bytes will be sent on the wire (and is it how CDNs cache as well - I don't know).

BigLep: @lidel : these numbers are uncompressed right?  Aren't the numbers we need to compare when compression is employed since that is how the bytes will be sent on the wire (and is it how CDNs cache as well - I don't know).

lidel: @BigLep depends on cache – afaik Nginx does not cache gzipped payload, but supports hosting of pre-compressed files if it finds  `.gz` variant. Also, we do not have compression in libp2p.

 Anyway, assuming we care only about HTTP, and go with default gzip for the samples above (iiuc cbor  grows due to gzip envelope):

- JSON: 197 bytes 
  (`ipfs block get baguqeerafsujii5p52wlc3fxrxhmbx6nu7uudyeznqjnvr7ogkjkndkuyukq | gzip -c | wc -c `)
- CBOR: 157 bytes
   (`ipfs block get bafyreignkd26ejp6jteqf4cggzbhp3twuvidqsa2lshknk6kjh7j2pc2h4 | gzip -c | wc -c `)


That seems to still give.. ~20% savings with CBOR.

lidel:
My understanding is this is not happening, since IPFS Stewards replaced Reframe use with simplier API defined in https://github.com/ipfs/specs/pull/337. 

Closing and moving to deferred column in https://github.com/orgs/ipfs/projects/19/views/1

PR #332, opened 2022-10-12
IPIP-332: Streaming Error Handling on Web Gateways

hacdias:
See https://github.com/ipfs/kubo/pull/9333#issuecomment-1273348381 for more context.


IPIP/0000-gateway-error-handling.md
```diff
@@ -0,0 +1,66 @@
...
+the user will not know that the TAR is incomplete. By introducing consistent error
+handling, the server attempts to notify the user.
+
+## Detailed design
+
+If the server encounters an error before streaming the contents to the client,
+the server must fail with the respective `4xx`  or `5xx` HTTP status code (no change).
+
+If the server encounters an error while streaming the contents, the server must
+force-close the HTTP connection to the user. This way, the user will receive a
```
Jorropo, line 30:
```suggestion
force-close the HTTP stream to the user. This way, the user will receive a
```
using HTTP2 or HTTP3 killing the connection would kill all streams, we want to kill on the impacted stream.

Jorropo:
I would like to be an options for clients to optout of this behaviour (with a header probably).

This has a real performance cost and will require clients to reopen a new connection if they are using `HTTP1.1`.
If your client as able to detect errors in car files or support trailing headers, you would probably like thoses instead of reopenning the connection.

IPIP/0000-gateway-error-handling.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Design rationale
+
+Before starting to stream the body of the response, the server is able to set
+an HTTP status code for the error. However, after the HTTP headers are set
+and the body started being streamed, there are no clear ways in the HTTP
+specification to show an error. Since the gateway is browser-first, it is
+important to show an error and avoid users receiving an incomplete file.
+Therefore, the server can force-close the HTTP connection, leading to a network
+error. This tells the user that an error happened.
```
Jorropo, line 45: Imo we need to specify in more précise terms (TCP RST for http1.1, ...)

Jorropo (changes requested):

http-gateways/PATH_GATEWAY.md
```diff
@@ -701,5 +701,9 @@ The usual optimizations involve:
...
 ## Streaming errors
 
-Gateways MUST force-close HTTP connections if they detect an error while
-streaming a file to avoid that users receive incomplete, yet valid, files.
+To avoid users receiving an incomplete, yet valid, files, the gateway MUST
+close the HTTP stream if an error occurs while streaming a file to the client.
+This can be done via the following mechanisms:
+
+- Sending a `RST` (reset) frame for HTTP/1.1
+- Sending a `RST_STREAM` (reset stream) frame for HTTP/2
```
Jorropo, line 709:
```suggestion
- Sending a `RST_STREAM` (reset stream) frame for HTTP/2
- Sending a `CANCEL_PUSH` frame for HTTP/3, see RFC9114 section A.2.5.3.
```

Jorropo (changes requested):

http-gateways/PATH_GATEWAY.md
marten-seemann: I haven't read the entire spec, but according to [BCP 56](https://www.rfc-editor.org/info/bcp56), HTTP APIs shouldn't go into this level of detail. HTTP is designed in a way that allows building applications without referring to the specific HTTP version.

Jorropo: I didn't actually red the RFC, I know RST_STREAM on HTTP/2 is what we want, I just CTRL + F RST_STREAM in the HTTP/3 RFC and found this text:
>    RST_STREAM (0x03):  RST_STREAM frames do not exist in HTTP/3, since
      QUIC provides stream lifecycle management.  The same code point is
      used for the CANCEL_PUSH frame (Section 7.2.3).

> HTTP is designed in a way that allows building applications without referring to the specific HTTP version.

Streaming errors is not a thing HTTP supports, you can setup trailler headers but they aren't supported by browsers.
So we are stealing the rug from under HTTP and use the underlying protocol to generate an unexpected error (which is actually supported and will return errors in all client implementations I've tested), that why we need to use TCP FIN.

I don't know if we should add this to the gateway spec, because I don't think most server implementations give you as much control as the go std does.
How will you implement that on reverse proxies ?

http-gateways/PATH_GATEWAY.md
```diff
@@ -694,3 +698,12 @@ The usual optimizations involve:
...
   - The downside of this approach is that it will always be slower than
     skipping child block resolution.
+
+## Streaming errors
+
+To avoid users receiving an incomplete, yet valid, files, the gateway MUST
+close the HTTP stream if an error occurs while streaming a file to the client.
+This can be done via the following mechanisms:
+
+- Sending a `RST` (reset) frame for HTTP/1.1
```
Jorropo, line 708:
```suggestion
- Sending a `FIN` (close) frame for HTTP/1.1
```
I confused RST and FIN mb

http-gateways/PATH_GATEWAY.md
hacdias: @marten-seemann as @Jorropo mentioned, the big issue here is that HTTP doesn't have any error handling for streaming. One of my motivations behind this IPIP regards the new TAR format for the gateway (https://github.com/ipfs/kubo/pull/9029).

It is possible that the TAR creation fails while streaming the file to the client due to many reasons. However, if you just stop streaming the TAR, you still get a valid TAR file, but it is incomplete. There's no feedback. Printing a trailer header is useless since browsers are not able to parse them. The only way we found so far to tell the user that something **is** wrong was by force-closing the HTTP stream.

I also have mixed feelings about having this on the spec since it is so specific. In addition, as @Jorropo mentioned it may be worth it having an opt-out of the behaviour through some header.

Jorropo: @hacdias 
> How will you implement that on reverse proxies ?

I think this is an important question.
I would like to see something like: `X-On-Error: reset` header sent by the server to indicate this will be done, then a reverse proxy could just not send thoses headers, and clients would be able to tell it is not gonna do this.

hacdias: @Jorropo please see the updates I've made to the IPIP.

lidel (changes requested):
Dropping some quick notes:

- Agree with Marten,  HTTP API should not go into low level connection detail. We want gateway interface to be future-proof.
- The more gateway implementations exist, the less control we have over these low level details. Need to solve this on higher client application layer.

:point_right:  due to this I am closing this IPIP


Instead, suggest opening a new IPIP with simpler proposal:

- The problem is that there is no way to signal that mid-way transfer error occurred.  Instead of trying to bolt-on something on the server side, instead, spec should specify inexpensive way how clients should detect that the entire CAR got transferred:

- Introduce optimization where we always add a zero-length raw block at the end of a CAR as a tombstone indicator – if this well-known CID is present, the client can skip hashing the last block, because we know the last block arrived safely.

- If tombstone is not present, then the client MUST verify hash of the last block. If  hash does match, means the connection was not interrupted mid-transfer (downside is additional CPU cost due to hashing that one block, that is why we prefer to have tombstone)

The tombstone is backward-compatible, and allows us to have an escape hatch to skip hashing and save CPU.

IPIP/0332-gateway-error-handling.md
```diff
@@ -0,0 +1,72 @@
...
+
+If the server encounters an error while streaming the contents, the server must
+force-close the HTTP stream to the user. This way, the user will receive a
+network error, making it clear that the downloaded file is not valid.
+In addition, the server must set an header `X-On-Error: reset` to indicate that
+connection will be reset if an error occurs.
+
+The client can opt-out of this behavior by sending a `X-Error-Behavior: trailer`.
+In that case, the server will not force-close the HTTP stream. Instead, the server
+will send the trailing header `X-Stream-Error` header with the error message.
```
lidel, line 37: As usual, reminding that Trailers can't be read by JS in web browsers: https://github.com/mdn/browser-compat-data/issues/14703

jbenet:
https://github.com/ipfs/specs/pull/332#pullrequestreview-1275495698

- null objects are a very nasty way to go.
- having a tombstone (ie some car files have it and some dont) is super messy and likely to lead to mysterious errors. 
- it also opens a surface area of attack: now everyone has to check that we arent linking to the null object from some object, because if we are, then it will terminate a car stream early, and we have to do that in a _ton_ of places (think the lovely wonders of `\0` in string operations). this is really nasty and you can guarantee that it will not be running  _ correctly everywhere_, so you can guarantee that it will be run _incorrectly somewhere_, and also have to deal _with that_.
- null termination is just a world of pain, and why ipfs went the way of self-describing objects. usually, most problems can indeed be turned into self-describing structures, even streams of unknown length. -- this is usually done by introducing a wrapper object that contains the information you want.
- for example, define a format like a `car-stream` that includes a header in between every object (similar to `tar`) and _that header_ can signal the end of a `car-stream`. you would have a 1-1 from any `car` to any `car-stream`, and make very explicit what is _data_ and what is _control plane_.

jbenet:
- more succinctly: i think you want a "content addressed stream" (CAS), not a "content addressed archive" (CAR).
- separately, this thread reminded me that we're likely due for a CARv3 with (a) support for explicit finite car files (ie the object count in the header), and (b) verifiable header (explicit roots in the header (as in v1) + a hash of the v2 index objects) -- such that a single hash (of a single small object) can be used to verify the entire thing (the original property CARs were invented for)

rvagg:
> more succinctly: i think you want a "content addressed stream" (CAS), not a "content addressed archive" (CAR).
separately, this thread reminded me that we're likely due for a CARv3 with (a) support for explicit finite car files (ie the object count in the header), and (b) verifiable header (explicit roots in the header (as in v1) + a hash of the v2 index objects) -- such that a single hash (of a single small object) can be used to verify the entire thing (the original property CARs were invented for)

Happily I think we can mush all of that into CARv2 without breaking. We'd just need that as a wishlist and some consumers of that functionality to help drive a spec to lock it in.

* https://github.com/ipld/go-car/pull/322
* https://github.com/ipld/js-car/pull/89

PR #333, opened 2022-10-14
fix: document CBOR in /reframe?q=

lidel:
Due to miscommunication, [go-delegated-routing v0.6.0](https://github.com/ipfs/go-delegated-routing)  uses DAG-CBOR  in `?q`  (and not DAG-JSON, like we expected) :see_no_evil: 
Details: https://github.com/ipld/edelweiss/issues/61
 
This PR updates the spec to describe the current pre-alpha implementation.
 
 @guseggert  if we are not planning to surgically fix this (as proposed in https://github.com/ipld/edelweiss/issues/61)  I suggest we update specs to reflect implementation in Kubo 0.16.

### DIFF
--- a/reframe/REFRAME_HTTP_TRANSPORT.md
+++ b/reframe/REFRAME_HTTP_TRANSPORT.md
@@ -32,7 +32,7 @@ All messages sent in HTTP body MUST be encoded as DAG-JSON and use explicit cont
 
 Requests MUST be sent as either:
 
-- `GET /reframe?q={percent-encoded-dag-json}`
+- `GET /reframe?q={percent-encoded-dag-cbor}`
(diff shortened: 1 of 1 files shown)


willscott (approved):
this does appear to reflect reality

maybe worth a comment about spaces being url encoded to '+'s and not being percent encoded? (although it's probably safer to percent encode them first and the decoder will take that as well)

lidel:
This turned out to be a big time sink – existing wire format and implementation has other issues, such as buffering streaming responses  due to Etag generation.

I am closing this – if we want to fix Reframe we should stop bending backward and changing specs to fix broken implementation – only way forward for Reframe over HTTP is to implement https://github.com/ipfs/specs/pull/327

PR #334, opened 2022-10-15
IPIP-334: Double-Hashed Find Providers in Reframe

willscott:
This PR describes a proposed request/response method for reframe to support "double-hashed" content routing lookups.

cc @guillaumemichel

### DIFF
--- a/reframe/REFRAME_KNOWN_METHODS.md
+++ b/reframe/REFRAME_KNOWN_METHODS.md
@@ -219,6 +219,9 @@ A message for finding nodes with interest in a given key using double hashing to
 ```
 
 The query is a derived hash of the multihash being requested.
+It is constructed by taking the raw bytes of the multihash, prepending the ascii bytes "CR_DOUBLEHASH", and taking the SHA256 hash of that data.
+The resulting digest is then packed itself into a multihash, using the multihash code identifier multihash.DBL_SHA2_256.
+
 The full semantics of double hashing in the context of content routing are described at https://www.notion.so/protocollabs/IPFS-Double-Hashing-Repurpose-8fdaae8748414ae592a5d24d59c0d8ed
 
 ```ipldsch
--- a/reframe/REFRAME_KNOWN_METHODS.md
+++ b/reframe/REFRAME_KNOWN_METHODS.md
@@ -22,6 +22,7 @@ This document is defining known methods (request-response message types) and sem
     - [Identify DAG-JSON Examples](#identify-dag-json-examples)
   - [FindProviders](#findproviders)
     - [FindProviders DAG-JSON Examples](#findproviders-dag-json-examples)
+  - [FindProvidersBlinded](#findproviders-blinded)
   - [GetIPNS](#getipns)
     - [GetIPNS DAG-JSON Examples](#getipns-dag-json-examples)
   - [PutIPNS](#putipns)
@@ -39,6 +40,7 @@ The known Request types are the following and are described below:
 type Request union {
     | "IdentifyRequest" IdentifyRequest
     | "FindProvidersRequest" FindProvidersRequest
+    | "FindHashedSHA256Request" FindHashedSHA256Request
     | "GetIPNSRequest" GetIPNSRequest
     | "PutIPNSRequest" PutIPNSRequest
     | "ProvideRequest" ProvideRequest
@@ -51,6 +53,7 @@ The known Response types are the following and are described below:
 type Response union {
     | "IdentifyResponse" IdentifyResponse
     | "FindProvidersResponse" FindProvidersResponse
+    | "FindHashedSHA256Response" FindHashedSHA256Response
     | "GetIPNSResponse" GetIPNSResponse
     | "PutIPNSResponse" PutIPNSResponse
     | "ProvideResponse" ProvideResponse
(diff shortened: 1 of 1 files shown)


reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -203,6 +208,39 @@ Response:
...
+    type FindHashedSHA256Request struct {
+        Query Bytes
+    }
+```
+
+The query is a derived hash of the multihash being requested.
+It is constructed by taking the raw bytes of the multihash, prepending the ascii bytes "CR_DOUBLEHASH", and taking the SHA256 hash of that data.
+The resulting digest is then packed itself into a multihash, using the multihash code identifier multihash.DBL_SHA2_256.
+
+The full semantics of double hashing in the context of content routing are described at https://www.notion.so/protocollabs/IPFS-Double-Hashing-Repurpose-8fdaae8748414ae592a5d24d59c0d8ed
```
guillaumemichel, line 225: The notion page was migrated [here](https://www.notion.so/pl-strflt/Double-Hashing-for-Privacy-ff44e3156ce040579289996fec9af609)

lidel: The page is not publicly available: 

>

reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -203,6 +208,39 @@ Response:
 }}
 ```
 
+### FindProviders Blinded
+
+A message for finding nodes with interest in a given key using double hashing to blind the key being requested.
+
+```ipldsch
+    type FindHashedSHA256Request struct {
```
lidel, line 216: What is the contingency for replacing SHA256 with other function in the future?
Creating a new message type?

reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -203,6 +208,39 @@ Response:
...
+
+```ipldsch
+    type FindHashedSHA256Request struct {
+        Query Bytes
+    }
+```
+
+The query is a derived hash of the multihash being requested.
+It is constructed by taking the raw bytes of the multihash, prepending the ascii bytes "CR_DOUBLEHASH", and taking the SHA256 hash of that data.
+The resulting digest is then packed itself into a multihash, using the multihash code identifier multihash.DBL_SHA2_256.
```
lidel, line 223: What is the contingency for replacing SHA256 with other function in the future?
Creating a new message type?

Would it be acceptable renaming these requests to `FingHashedProvidersRequest` and move hash function multicodec to struct field?

willscott: The current contingency is a new message type.

having it in the request is sub-optimal because it would mean another round of probing for what types of hashes are supported.

changing the hash algorithm will be 'a big deal' and something we don't do often, so using the existing 'identify' layer to identify which one the server supports seems preferable to the client trying other ones and getting told they're not supported.

we expect handlers of FindProviders to support 1 or at most 2 hash algorithms at any given time.

reframe/REFRAME_KNOWN_METHODS.md
guillaumemichel: Sorry, this link should work https://pl-strflt.notion.site/Double-Hashing-for-Privacy-ff44e3156ce040579289996fec9af609

reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -203,6 +208,39 @@ Response:
...
+A message for finding nodes with interest in a given key using double hashing to blind the key being requested.
+
+```ipldsch
+    type FindHashedSHA256Request struct {
+        Query Bytes
+    }
+```
+
+The query is a derived hash of the multihash being requested.
+It is constructed by taking the raw bytes of the multihash, prepending the ascii bytes "CR_DOUBLEHASH", and taking the SHA256 hash of that data.
```
guillaumemichel, line 222: Should support prefix requests, with variable length Queries
```suggestion
It is constructed by taking the raw bytes of the multihash, prepending the ascii bytes "CR_DOUBLEHASH", and taking a variable length prefix of the SHA256 hash of that data.

```

reframe/REFRAME_KNOWN_METHODS.md
```diff
@@ -203,6 +208,39 @@ Response:
...
+The resulting digest is then packed itself into a multihash, using the multihash code identifier multihash.DBL_SHA2_256.
+
+The full semantics of double hashing in the context of content routing are described at https://www.notion.so/protocollabs/IPFS-Double-Hashing-Repurpose-8fdaae8748414ae592a5d24d59c0d8ed
+
+```ipldsch
+    type FindHashedSHA256Response struct {
+        Providers [Provider]
```
guillaumemichel, line 229: `FindHashedSHA256Response` should contain a map of Bytes -> Encrypted Provider. A single `FindHashedSHA256Request` using prefix lookup may return multiple encrypted provider records matching the prefix.

lidel:
Reframe is no more (https://github.com/ipfs/kubo/issues/9479)

`/routing/v1` version is tracked in https://github.com/ipfs/specs/pull/421

PR #336, opened 2022-10-17
Remove empty API docs

guseggert:


lidel (approved):
Thanks for cleaning this up!

PR #340, opened 2022-11-10
IPIP 298: (allow|deny)lists for IPFS Nodes and Gateways

thibmeu:
Spec for the IPIP-298.

It adds a standard reference for content list definition, should these
be allow lists or deny lists. It supports CIDv1, sha256 hashed CIDv1,
and absolute and partial path matching.
It is aimed at offering a common schema for node operators to interact
with content lists.

It should be noted this IPIP was first written by @foreseaz, and this PR
is an iteration over their work in #299.

cc @Jorropo @lidel


dokterbob:
Very very very interesting to see this moving forward! I will leave some comments/questions inline!

dokterbob (commented):
Super super excited to move this forward. Navigating the space between moderation and censorship is a tricky one and having a well thought-out spec to allow for collaboration is essential to keeping the IPFS/IPLD/Filecoin space accessible to small parties and/or ideologically motivated initiatives.

I gave it a good read and left some comments/questions. Main observations are:

1. We might consider using CID's to do the double hashing (which is apparently standardised now), and gain various other benefits (e.g. CID upgrades, not worrying about codecs).
2. Formalising reasons on a per-item basis might increase transparency and granularity in terms of filtering.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,243 @@
...
+## Schemas
+
+### Definition
+
+The schemas below are defined as [IPLD schemas](https://ipld.io/specs/schemas/).
+
+```ipldsch
+type List struct {
+  Action Int # Suggested action to be taken
+  StatusCode optional Int # HTTP status code to be returned by IPFS gateways by default
```
Jorropo, line 63: Same comment as in https://github.com/ipfs/specs/pull/299#discussion_r934986343 this should be authoritatively limited.
I think anything not `451`, `410` or `200` has no place on a badlist list. I don't want people start to use 3xx or whatever.

Secondly, I don't fully understand why giving out HTTP codes, assuming the goal is to join forces and let gateway operators share ban reasons, **this is really unspecific**,
I would like a machine readable version enum, containing things like `Dos`, `Legal` (maybe with a reason like `Legal Copyright`, `Legal Hatespeech`, `Legal other`, ...), ...
Some gateway operators might prefer faking `500`s or just flat out ignore certain reasons.

dokterbob: > I would like a machine readable version enum, containing things like Dos, Legal (maybe with a reason like Legal Copyright, Legal Hatespeech, Legal other, ...), ...

❤️

LaurenSpiegel: Would remove the word "Legal" since not really a legal opinion or necessarily illegal depending on the jurisdiction.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,243 @@
...
+
+  Entries optional Entry[] # Entries included in the list
+  Imports optional Bytes # CID to a chain of entries included in the list
+
+  Description optional String # Context to be shared with the list consumers
+}
+
+type Entry struct {
+  Type Int # Type of the content string. Refer to CID_V1, ABSOLUTE_PATH, PARTIAL_PATH, SHA256_CID_V1
```
hsanjuan, line 72: There should be a type that has has exactly the same meaning as the current badbits anchor.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,243 @@
...
+      Type: 2, // Type.PARTIAL_PATH
+      Content: "bag.again.example.com",
+      StatusCode: 410
+    },
+    {
+      Type: 3, // Type.SHA256_CID_V1
+      Content: "9056e0f9948c942c16af3564af56d4bb96b6203ad9ccd3425ec628bcd843cc39",
+      Description: "sensitive cid that needs to be blocked"
+    }
+  ]
```
hsanjuan, line 114: Note that it is very difficult to append anything to this list. You cannot have one process writing to it and another one reading from it at the same time. You need to open, read, deserialize, serialize, write, close.

That means that if you have an external process generating items to be blocked, every update needs to edit the file and then trigger the IPFS implementation to re-read it (possibly in full). Or the IPFS implementation needs to provide API to make the updates itself (and then write the file (again, possibly in full).

I have the feeling this format is not well suited to very large lists, or those updated frequently.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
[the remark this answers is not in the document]
thibmeu: I've added a comment to remove the HTTP specific handling.

thibmeu (commented):
I added replies to comments. New feedback still appreciated.
I'll update the document on next week, with the change of StatusCode to Categories.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,242 @@
...
+
+An IPFS node is a service that interacts with the IPFS network. This could be because it's providing content to the network, because it's retrieving content, or because it's serving IPFS hosted content through an HTTP interface.
+
+To interact with content lists, it operates a rule engine. The rule engine decides what action should be taken for a particular request.
+For instance, if CID bafy...badcid is denied, the bitswap engine should not request it, and access to bafy...badcid through a node exposed gateway are going to 403 forbidden.
+
+### User
+
+User is an actor that can make request to the IPFS Node. It's considered authenticated already.
+Access to the node can be made via an HTTP gateway, Bitswap, or another transport method exposed by the IPFS Node.
```
lidel, line 50: nit: covering more ground

```suggestion
Access to the node can be made via an HTTP gateway, command-line, RPC API, Bitswap, or another transport method exposed by the IPFS Node.
```

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,242 @@
...
+  Categories optional [Int] # Content category entries should be associated to by default
+
+  Entries EntryChunk # Chain of Entries included in the list
+
+  Description optional String # Context to be shared with the list consumers
+}
+
+type EntryChunk struct {
+  Entries [Entry] # List of Entries objects
+  Next optional Link # CID for the next EntryChunk object. Entries referenced this way are not inlined, and possibly require an additional request for discovery.
```
lidel, line 72: Having a single `Link` makes it:
- impossible to  compose a list from two preexisting ones (e.g., having "anti-phishing" meta-list composed of lists provided by gateways)
- expensive to maintain a big list (instead of having a DAG, and only recomputing changed branch, one needs to change all lists along the single chain).


Would it be ok to change this field into array of links?

```suggestion
  Next optional [Link] # List of CIDs for the next EntryChunk objects. Entries referenced this way are not inlined, and possibly require an additional request for discovery.
```

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
lidel: Would defining a max list size as part of this spec help?
Together with `Link` field, it allows for composing big lists from smaller ones, and updates would only change part of a bigger DAG.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
thibmeu: What about having multiple `Entries [EntryChunk]` instead of multiple `Next [Link]`? I feel it makes it:
* Easier to read when lists are composed: it's simply a new EntryChunk.
* Does not modify a specific EntryChunk upon the addition of a new list.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
hsanjuan: That doesn't improve anything if instead of 1 item of 100GB we have 100 items of GB, somehow linked.

I also don't think the list format should include a Link field. Links to multiple lists etc should be a meta-format above that can be implemented later if we need some glue to handle multi-subscriptions (right now not needed).

I personally think lists should be line-based files, as small as possible, editable by hand for simple things like appending a CID. See lists used by UBlock etc.

lidel: ## Inefficiencies of CBOR lists vs prior art from adblockers

Ack. A compact, efficient, line-based syntax similar to [static filter lists used by adblockers](https://github.com/gorhill/uBlock/wiki/Static-filter-syntax) ([example](https://easylist.to/easylist/easylist.txt)) will be more efficient and easier to consume than having CBOR.


Main problem: duplicated  "Type" "Content" and "Description" strings for every entry – havign these strings over and over again in CBOR map keys gets very wasteful very fast if the list is big.

## Manifest vs Compact list

So.. this PR feels like the meta format for aggregating lists you describe.  In [the other comment](https://github.com/ipfs/specs/pull/340#discussion_r1026509423) @hsanjuan  suggested adding a type for already existing double hashed format from https://badbits.dwebops.pub, we could do the same here and specify a separate type for lists in the "compact syntax". 

I think we should decouple and work on these things in parallel (meta-manifest and the compact list formats).
Will make discussion more productive, and allow us to ship things faster (e.g., in Kubo, we need compact format sooner than the meta-manifest one: users should be able to self-service, will take time).

## Next steps

@hsanjuan i'd like to explore your proposal – would you have time to propose a syntax for compact list? 
(gist / hackmd is fine – I can turn it into a separate IPIP  if you have limited bandwidth – this way we could flesh it out without blocking this one, and then we can decide which approach makes more sense)

(side-question) @thibmeu how strong is the CBOR  requirement?

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
lidel: - Yes, `[EntryChunk]` also works.  
  - Remaining problem around `Link` is that there is no place to provide description for it– If i am aggregating third-party lists into a single meta-lists, would be nice to note why each link was included. Maybe add `Description optional String` to `EntryChunk` ?
- re: max size, i only meant max size per document, encouraging people to "shard"  at the same cut-off and making it easier for implementers to fail fast when consuming a single document. (total across lists is a separate problem).

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
thibmeu: `Entries` structure could be transformed to a CSV. That's also the reason why multihash is serialised to a string via multibase, so it's easier to export.
```
type,content,description,categories
1,/path/to/content,questionable content,1001 1002 1005
...
```
I agree this would be easier to update. Metadata around it could be passed as a frontmatter possibly
```
action: 1
categories: 1000 1001
description: an example content list
---
type,content,description,categories
1,/path/to/content,questionable content,1001 1002 1005
```

The front matter still allows for compostability with CBOR, but makes the list self contained in a text format if need be.

I would not worry about the existing format for bad bits list. It's a single list, and I would avoid requiring the spec and implementation to be constrained by one historical list instance (similar to Cloudflare safemode in a way).

I have no need for CBOR specifically, it's just a format supported by IPFS and to which an IPLD schema can serialise to.

hsanjuan: > i'd like to explore your proposal – would you have time to propose a syntax for compact list?

yes, next week at some point? I have ideas but I also need to explain them (because they link to other aspects that are not related to the list itself), which takes more time. Format would be in the direction @thibmeu proposes.

BigLep:
@thibmeu @mathew-cf and others: I wanted to make sure you saw the expansion of this topic that is occurring in https://github.com/ipfs/specs/pull/383 .  Your feedback on whether that is meets Cloudflare's needs is welcome.

IPIP/0298-denylists-for-ipfs-nodes-and-gateways.md
```diff
@@ -0,0 +1,243 @@
...
+|2|PARTIAL_PATH|Part of a path of a content|Path as a string|Requested path contains|
+|3|SHA256_CID_V1|SHA256 hased CID v1. This matches the [Double Hashing IPIP-334](https://github.com/ipfs/specs/pull/334).|SHA256 of the CIDv1 as a string|Requested CID v1 SHA256|
+|4-255|Unassigned||||
+|256-512|INTERNAL|This range can be used for private content type|||
+
+> **Note on `SHA256_CID_V1` type**
+>
+> The main difference between non-hashed entries and hashed ones is that the CIDs or content paths in the entry will be hashed and no plaintext is shown in the list. Following the [bad bits](https://badbits.dwebops.pub/), each CID or content path is `sha256()` hashed, so it's easy to determine one way but not the other. The hashed entries are designed to store sensitive blocking items and prevent creating an easily accessible list of sensitive content.
+>
+> Before the hashing, all CIDv0 in `CID_V1` field is converted to CIDv1 for consistency.
```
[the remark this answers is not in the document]
expede: I'm super late to the party, and this is maybe fixed elsewhere, but I wanted to +1 @dokterbob's suggestion about salting the double hashes (and making sure that the salt doesn't match the one in the double-hashed DHT). Otherwise you're handing a treasure map to bad actors.

There's perhaps a broader discussion about the privacy properties that we need from such a system. These are all some form of private set intersection (PSI), where double hashing is the "naive" case — there are much more sophisticated techniques. (Fancier _does not_ nessesarily mean better, but may be worth exploring)

PR #337, opened 2022-10-18
IPIP-337: Delegated Content Routing HTTP API

guseggert:

PR #342, opened 2022-11-12
Proposal for automatic discovery of content routers

willscott:
This follows the previously circulated proposal outline at https://hackmd.io/bh4-SCWfTBG2vfClG0NUFg

A basic motivation is included in the PR - but essentially this is the best path I've heard for reducing our dependence on hydras as a centrally operated choke point for moving the bulk of the IPFS network beyond sole reliance on the current KAD DHT.

### DIFF
--- a/IPIP/0000-content-router-discovery.md
+++ b/IPIP/0000-content-router-discovery.md
@@ -1,0 +1,250 @@
+# IPIP 0000: Content Router Ambient Discovery
+
+<!-- IPIP number will be assigned by an editor. When opening a pull request to
+submit your IPIP, please use number 0000 and an abbreviated title in the filename,
+`0000-draft-title-abbrev.md`. -->
+
+- Start Date: 2022-11-11
+- Related Issues:
+  - https://hackmd.io/bh4-SCWfTBG2vfClG0NUFg
+  - https://github.com/ipfs/kubo/issues/9150
+  - https://github.com/filecoin-project/storetheindex/issues/823
+
+## Summary
+
+The Interplanetary stack has slowly opened itself to support extensibility of
+the content routing subsystem. This extensibility is used today by network
+indexers, like https://cid.contact/, to bridge content from large providers
+that cannot practically provide all content to the IPFS DHT. A missing piece
+of this story is that there is not a process by which IPFS nodes can discover
+these alernative content routing systems automatically. This IPIP proposes
+a mechanism by which IPFS nodes can discover and make use of content routing
+systems.
+
+## Motivation
+
+There is currently not a process by which IPFS nodes can discover alernative
+content routing systems automatically. This has led to a reliance on
+centralized systems, like the hydra boosters, to fill the gap and offer
+content only available in network indexer to current IPFS nodes. This strategy
+is also insufficient long term because:
+1. It limits speed to the use of a globally distributed kademlia DHT
+2. It is insufficient for providing content in applications where content grows
+    super-linearly to peers, such that the burden on a traditional DHT would
+    become unsustainable.
+
+
+## Detailed design
+
+### 0. content-router discovery state tracking
+
+Nodes will conceptually track a registry about known content routers.
+This registry will be able to understand for a given content router two
+properties:
+* reliability - how many good vs bad responses has this router responded
+with. This statistic should be windowed, such that the client can calculate
+it in terms of the last week or month.
+* performance - how quickly does this router respond.
+
+This protocol expects nodes to be able to keep reliability (a metric
+capturing both availability and correctness) separate from performance
+for the purpose of propagating content routing information.
+
+In addtion, nodes may wish to track the most recent time they have learned
+content routing information from the other peers they are and have been
+connected with.
+
+### 1. content-routing as a libp2p protocol
+
+IPFS nodes will advertise and coordinate discover of content routers using a
+new libp2p protocol advertised as "".
+
+The protocol will follow a request-response model.
+A node will open a a stream on the protocol when it wants to discover new
+content routers it does not already know.
+It will send a bloom filter as it's query.
+* The size of the bloom filter is chosen by the client, and is sized such
+that it receives a greater than 99% certainly that it receives a useful
+response. The maximum size of a query may be capped by the server, but can be
+effectively considered to be under 10kb.
+* The client will hash it's known content routers into the bloom filter
+to set bits in the filter at the locations to which these known routers
+hash.
+* The server will have a parameter for a number of servers it wants to return
+to content routing queries. By default this will be 10. (This default is picked
+as the result of modeling router propagation). It will iterate through it's
+list of known content routers, hashing htem against the bloom filter and
+selecting the top routers that are not already known to the client. It will
+return this list, along with it's reliability score for each. This response
+is structured as an IPLD list lists, conceptually:
+```json
+[
+  ["https://cid.contact/", 0.95],
+  ["https://dev.cid.contact/", 0.90],
+]
+```
+
+### 2. probing of the discovery protocol
+
+A node will probe it's connected peers for content routing updates in two
+situations:
+
+1. When it needs to perform a content routing query, and has not
+successfully performed a sync in over a day.
+2. When it's auto-nat status indicates it is eligible to be a DHT server, and
+it has not successfully performed a synce in over a day.
+
+These parameters are also set through modeling.
+
+To perform a probe, the node will consider the set of peers it is currently
+connected to. It will order peers. The specific ordering is left to the
+node, but it should strive for diversity - an example ordering would be to
+rank peers by how recently a content routing discovery query has been make
+to that peer, with tie breaking preference for LAN nodes and for boostrap
+nodes.
+
+### 3. selection of routers
+
+Nodes are free to make content routing queries across content routing
+systems they are aware of as they wish. An example strategy balancing
+user experience and discovery is described.
+
+The node maintains two thresholds:
+* good (reliability > 99%, performance < 100ms)
+* uncertain (queries < 5)
+
+Content routers meeting the good reliability threshold are ordered by
+perforamnce. the top one is queried, as is an 'uncertain' router if
+one exists.
+
+These threshold values are maintained for a year for the purposes
+of local selection.
+They are maintained for a month for the purpose of admitting
+knowledge of routers to others - so a client will no longer set bits for
+routers it is aware of but which do not meet it's threshold for 'good'
+after a month. If peers then subseuqently respond with these nodes
+on discovery probes, the local node may use that to consider the
+node as again 'uncertain' and attempt additional probes against it less than
+a year later.
+
+Nodes which participate as DHT servers should also consider if they
+are being used only in an infrastructural capacity. If they are
+receiving content routing requests from other peers, but there have been
+no direct requests from the node itself that can be used to move
+known content routers past the 'uncertain' threshold, the node may
+choose to issue content routing queries for a fraction of the DHT
+lookup queries it receives as a way to maintain a more accurate
+table of content  routers.
+
+## Test fixtures
+
+TK is a CID currently only available through the content routing system,
+and not through the IPFS DHT. This is a piece of content that can be queried
+to validate the presence of alternative content routing systems.
+
+## Design rationale
+
+As expressed in the motivation section, we need to design a system through
+which nodes can discover content routers without a centralized point of
+failure, and can use these routers to improve user performance for content
+routing to levels faster than the current DHT.
+
+This design is self-contained - it does not require standing up additional
+infrastructure or making additional connections for discovery but rather
+gossips routers over existing peer connections.
+
+The design limits the ability of an adversary to impact user experience:
+1. it does not propose at this stage to replace DHT queries, but only to
+supplement them with content routing queries, which minimized user
+noticable impact.
+2. nodes will only propogate content routers they believe to work,
+limiting the spread of spam / unavailable content routers to the directly
+connected peers of an adversary.
+
+With the exception of LAN tables, the other connections made by IPFS
+nodes do not have geographic locality. As a result, performance is
+separated in the tracking of content routers because it will not be
+effective as a ranking factor in the non-geographically-aware
+gossip system described here. As an optimization, nodes may choose to
+prioritize 'fast' content routers when responding to queries from peers
+where sharded latency observations may be relevant. For example:
+* Peers on the local LAN
+* Peers in the local /16 IPv4 subnet
+* Peers with observed latency less than 25ms
+
+### User benefit
+
+Users will benefit from faster discovery of content providers.
+Users will also benefit from access to more CIDs than they currently do through
+queries limited to the IPFS DHT
+
+### Compatibility
+
+Nodes which do not upgrade to support this IPIP will be limited to the sub-set of
+content available in the DHT. this will potentially degrade over time as more
+large providers limit their publishing per the IPNI ingestion protocol.
+
+Nodes may limit their complexity through a hard-coded list of known content
+routers, essentially limiting their implementation to design section 3 of this
+IPIP. In doing so, they may limit their risk of exposure to malicious parties.
+They risk being out of date and to offer sub-optimal performance through their
+failure to discover additional near-by content routing instances.
+
+### Security
+
+TODO: this section provides a rough sketch of arguments, but has not been fully
+developed into prose at this time. At present, it is most useful for
+comments and suggestions of other security considerations that should be
+included as this draft develops.
+
+#### 1. Malicious Content Routers
+##### a. Providing Bad Content Routing Records
+
+* records under double hashing are signed, so can't provide a record for a real peer
+* if you provide non-working records, you are down-ranked
+
+##### b. Availability Attacks / failing to provide records
+
+* if list of records insufficient, client will get more from other providers in subsequent queries, leading to downranking
+
+#### 2. Exposure of IPFS Clients (enumeration of network participants)
+
+* a new provider is only visible to directly connected peers. they only forward it to peers asking them if it meets their bar
+for reliability.  This means propogation through the network is only posisble for routers that behave correctly.
+* because clients only propogate their 'top' routers, latency is also relevant, and with sufficient number of routers, the would only
+propogate in their local geographic area before becoming uncompetitive on latencyk
+
+### Alternatives
+
+#### Ambient discovery in the style of circuit relays
+
+Circuit relays are discovered ambiently by nodes during protocol enumeration.
+When connecting with another libp2p node, IPFS nodes will probe
+supported protocols. If they notice circut relay support at this time, they
+make use of such aggregated knowledge when making connections needing the
+support of relays.
+
+This is not considered sufficient for content routing, because most content
+routers will not act as general peers within the IPFS mesh, so they would
+not be directly discovered. Instead, the gossip discovery protocol is
+ambiently discovered in much the same way as circuit relays.
+
+#### Advertisement in the DHT
+
+This suffers from one of two problems depending on tuning: Either it results in
+a global list that all clients see new providers, or it takes an inordinant
+amount of querying before a client happens to run into a provider, leading to
+degraded experiences for most clients. The single global list that a provider
+can automatically add itself to leads to issues for how to mitigate an
+enumeration of all network participants by a malicious content router.
+
+#### Static list of known routers distributed with IPFS clients
+
+This has worked for the current IPFS bootstrap node, but leads to the need for
+policies around how to decide which content routers will be included in such a
+list, and fails to evolve efficiently as new content routers are added to the
+system.
+
+### Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


IPIP/0000-content-router-discovery.md
```diff
@@ -0,0 +1,250 @@
...
+that it receives a greater than 99% certainly that it receives a useful
+response. The maximum size of a query may be capped by the server, but can be
+effectively considered to be under 10kb.
+* The client will hash it's known content routers into the bloom filter
+to set bits in the filter at the locations to which these known routers
+hash.
+* The server will have a parameter for a number of servers it wants to return
+to content routing queries. By default this will be 10. (This default is picked
+as the result of modeling router propagation). It will iterate through it's
+list of known content routers, hashing htem against the bloom filter and
```
mxinden, line 76:
```suggestion
list of known content routers, hashing them against the bloom filter and
```

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+receiving content routing requests from other peers, but there have been
+no direct requests from the node itself that can be used to move
+known content routers past the 'uncertain' threshold, the node may
+choose to issue content routing queries for a fraction of the DHT
+lookup queries it receives as a way to maintain a more accurate
+table of content  routers.
+
+## Test fixtures
+
+TK is a CID currently only available through the content routing system,
```
ajnavarro, line 137: TK? do you mind to uncompress the alias? :)

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+## Design rationale
+
+As expressed in the motivation section, we need to design a system through
+which nodes can discover content routers without a centralized point of
+failure, and can use these routers to improve user performance for content
+routing to levels faster than the current DHT.
+
+This design is self-contained - it does not require standing up additional
+infrastructure or making additional connections for discovery but rather
+gossips routers over existing peer connections.
```
ajnavarro, line 150: this is the first time gossip is mentioned. Should we be more specific in the `Detailed design` section about the protocol and how nodes will be interconnected?

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+
+Users will benefit from faster discovery of content providers.
+Users will also benefit from access to more CIDs than they currently do through
+queries limited to the IPFS DHT
+
+### Compatibility
+
+Nodes which do not upgrade to support this IPIP will be limited to the sub-set of
+content available in the DHT. this will potentially degrade over time as more
+large providers limit their publishing per the IPNI ingestion protocol.
```
ajnavarro, line 181: IPNI?

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+supported protocols. If they notice circut relay support at this time, they
+make use of such aggregated knowledge when making connections needing the
+support of relays.
+
+This is not considered sufficient for content routing, because most content
+routers will not act as general peers within the IPFS mesh, so they would
+not be directly discovered. Instead, the gossip discovery protocol is
+ambiently discovered in much the same way as circuit relays.
+
+#### Advertisement in the DHT
```
ajnavarro, line 228: The good things about advertising using the DHT are:
- Network that is already there, no need to create a new protocol to "provide" new providers instead of CIDs.
- You can provide associating your provider with a specific root CID content. I seriously doubt that all providers will be eager to provide all CIDS in the universe.

ajnavarro (commented):
I'm missing a way to link content with the provider because I seriously doubt that all parties will be eager to provide all the CIDs in the universe, they will focus on providing the content they care about.

We could link root CID/CIDs with the provider to let know to nodes where they can find the DAG for specific content.

willscott:
> I'm missing a way to link content with the provider because I seriously doubt that all parties will be eager to provide all the CIDs in the universe, they will focus on providing the content they care about.
> 
> We could link root CID/CIDs with the provider to let know to nodes where they can find the DAG for specific content.

This is a philosophical disagreement about what a 'content router' is.
I'm going to refer you to the [broad guidance](https://www.youtube.com/watch?v=t0u-xMwWra4) that was presented over the summer for the evolution of content routing.

We currently have a couple examples of content routers that do have all the CIDs in the universe, and we do not have convincing examples of, or a definition for, sub-content-routers as you're proposing here.

Why are we trying to compromise to a much-harder-to-make-work complexity without trying for the thing that makes sense and the direction we're heading first?

IPIP/0342-content-router-discovery.md
willscott: https://github.com/ipni

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+for the purpose of propagating content routing information.
+
+In addtion, nodes may wish to track the most recent time they have learned
+content routing information from the other peers they are and have been
+connected with.
+
+### 1. content-routing as a libp2p protocol
+
+IPFS nodes will advertise and coordinate discover of content routers using a
+new libp2p protocol advertised as "".
```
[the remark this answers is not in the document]
willscott: do you have thoughts on what a good name would be?
maybe '/ipfs/content-router-discovery/1.0.0' ?

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+
+### 0. content-router discovery state tracking
+
+Nodes will conceptually track a registry about known content routers.
+This registry will be able to understand for a given content router two
+properties:
+* reliability - how many good vs bad responses has this router responded
+with. This statistic should be windowed, such that the client can calculate
+it in terms of the last week or month.
+* performance - how quickly does this router respond.
```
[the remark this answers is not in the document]
willscott: Yes, this would be windowed. I was imagining a window of ~ "last week" by default, but this seems like a good candidate to evaluate through simulation.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,246 @@
...
+* The client will hash it's known content routers into the bloom filter
+to set bits in the filter at the locations to which these known routers
+hash.
+* The server will have a parameter for a number of servers it wants to return
+to content routing queries. By default this will be 10. (This default is picked
+as the result of modeling router propagation). It will iterate through it's
+list of known content routers, hashing them against the bloom filter and
+selecting the top routers that are not already known to the client. It will
+return this list, along with it's reliability score for each. This response
+is structured as an IPLD list lists, conceptually:
```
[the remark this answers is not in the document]
willscott: a json or cbor array are both examples that would fulfill this, I'll leave it more generic but i have a hard time imagining we'd encode this in a way that wouldn't conform to being considered an IPLD list

ajnavarro:
@willscott I think more than a philosophical disagreement is a physical one.

Right now we are able to keep all the CIDs on the network in one provider for two reasons:
- The network is relatively small
- Right now we have only two places where nodes are providing CIDs: DHT and Bitswap, so it is easier to get that information because is only in two specific places.

When we start to have different ways of providing CIDs, will be near impossible to have everything replicated by everyone. 

Also when the network scales, having centralized all the information in several places will be quite challenging and costly. But on the other hand, allowing both approaches (providing everything vs providing a subset of the CIDs) will have for sure their use cases for people with not a huge amount of money to maintain big providers.

willscott:
We currently have providers stepping up to provide full replicas of a content routing database. That is what network indexers have been doing over the last year.
I don't see the physical issue here: indexers are handling trillions of records, vs the 100 billion in the DHT, and the scale we expect as we grow will still fit at the level of a single rack in a data center.

In designing delegated routing so far, the eye has been towards a design where delegated routers need to fall back and do the additional work of querying other routers in order to collect a full replica if they don't possess it themselves, rather than making that the end kubo node's responsibility, as that leads to an untenable performance and decision process for end user nodes that are not equipped to handle that.

I'm not entirely sure of your counter proposal here: I think there are very strong counter arguments against both https://github.com/ipfs/specs/pull/322 - which compromises trying to be a content addressed network, or limited DHT providing (e.g. to roots) - which still couldn't handle the current indexer database scale.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+
+The protocol will follow a request-response model.
+A node will open a stream on the protocol when it wants to discover new
+content routers it does not already know.
+The node wants to request the best set of known content routers from it's peer
+that it does not already know. The query will make use of a bloom filter to
+support this prioritization without leaking the exact list of known content
+routers that the client already knows.
+
+* The size of the bloom filter is chosen by the client. It is sized such
```
aschmahmann, line 76: Unfortunately GitHub doesn't allow threads not tied to a line, but wanted to add some thoughts to this discussion https://github.com/ipfs/specs/pull/342#issuecomment-1316854045 in a way that responses would be easy to trace.

1) Per @ajnavarro's comment https://github.com/ipfs/specs/pull/342#issuecomment-1318830270 I too find it a little hard to believe that there will be so many routers each providing full replicas of all data tracked by IPNI that a bloom filter would be required given that running these servers is expensive and incentivization is IIUC mostly TBD (I think @guseggert had some napkin math here showing the large costs around storing 10^15 CIDs even if we exclude bandwidth costs). That being said this is `#not-this-ipips-problem`. If the IPNI team thinks thousands of nodes all over the world will spring up hosting PBs of data and that lack of consistency between replicas isn't going to cause problems with the evaluation criteria that clients use that problem seems to live elsewhere.
2) Whether or not #322 is a good/bad idea is also `#not-this-ipips-problem` since this describes how to find routers for a given content routing system (i.e. IPNI) not whether it should be passable as a hint.
    - As an aside my 2c is that you've got to be careful here to not break IPLD properties if you go this route as I've flagged in #322, however it's potentially useful to add hints as long as they're not mandatory.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+their knowledge of router existance to each other. Initially, we expect that
+the current topology will look a bit more like a feedback loop over a
+bipartite graph - where one side of the graph is the set of general purpose
+IPFS nodes, and the other side are the bootstrap and core-infrastructural
+nodes with high connectivity in the network.
+
+### 1. content-routing as a libp2p protocol
+
+IPFS nodes will advertise and coordinate discover of content routers using a
+new libp2p protocol advertised as "/ipfs/content-router-discovery/1.0.0".
```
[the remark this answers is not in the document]
willscott: > This currently seems to be specifically for IPNI routers so at best this is /ipfs/ipni-discovery/1.0.0

this is for discovery of content routers per the delegated content router API - what about this is IPNI specific?

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+
+Nodes which do not upgrade to support this IPIP will be limited to the sub-set of
+content available in the DHT. this will potentially degrade over time as more
+large providers limit their publishing per the IPNI ingestion protocol.
+
+Nodes may limit their complexity through a hard-coded list of known content
+routers, essentially limiting their implementation to design section 3 of this
+IPIP. In doing so, they may limit their risk of exposure to malicious parties.
+They risk being out of date and to offer sub-optimal performance through their
+failure to discover additional near-by content routing instances.
```
lidel, line 201: >  hard-coded list [..] In doing so, they may limit their risk of exposure to malicious parties.

Hardcoding HTTP URLs is not always "limiting risk", it is changing the threat model by introducing a single point of failure as an obvious denial of service vector.  Depending on use case, may actually **increase the risks to the user**. 
Better to remove this statement.

```suggestion
Nodes may limit their complexity through a hard-coded list of known content
routers, essentially limiting their implementation to design section 3 of this
IPIP.  This comes at a price: (1) hard-coded routers become easy targets 
for denial of service attacks, decreasing the resilliency of the entire setup;
(2) nodes risk being out of date and to offer sub-optimal performance through their
failure to discover additional near-by content routing instances.
```

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+connected with.
+
+Conceptually, propagation of content routers will look like nodes gossiping
+their knowledge of router existance to each other. Initially, we expect that
+the current topology will look a bit more like a feedback loop over a
+bipartite graph - where one side of the graph is the set of general purpose
+IPFS nodes, and the other side are the bootstrap and core-infrastructural
+nodes with high connectivity in the network.
+
+### 1. content-routing as a libp2p protocol
```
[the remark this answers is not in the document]
lidel: I suspect the idea is for `/dnsaddr/bootstrap.libp2p.io` (or any other bootstrapper set by JS user, as long its  `/webtransport` or `/wss`)  to speak this new protocol, avoiding hardcoding anything new.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
+# IPIP 0342: Content Router Ambient Discovery
```
lidel, line 1: Since I can't expose a partial index without being punished, would this be more correct?

```suggestion
# IPIP 0342: IPNI Content Router Ambient Discovery
```

IPIP/0342-content-router-discovery.md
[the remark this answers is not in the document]
willscott: what prevents them from participating in this protocol as described?
browser nodes will need to contact to other existing nodes, as they do today. they would learn about the existence of content routers through those same channels via the new protocol, and could then make use of them.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+amount of querying before a client happens to run into a provider, leading to
+degraded experiences for most clients. The single global list that a provider
+can automatically add itself to leads to issues for how to mitigate an
+enumeration of all network participants by a malicious content router.
+
+Pros:
+* Network is already there, no need to create a new protocol to "provide" new providers instead of CIDs.
+* You could potentially associate a provider with a specific root CID content.
+Cons:
+* Nodes cannot drop use of the DHT / other content routing options always are 'second tier'.
```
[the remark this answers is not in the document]
willscott: the alternative of other content routers being found in the DHT, which is what this alternative is trying to describe, does mean that no ipfs node could be run without kad dht code for DHT lookups. Being a DHT participant is more complexity than just having libp2p code to be able to connect to other peers, and at least my understanding of what's being proposed in this alternative is that it is intertwined with the DHT and not equivalent to hardcoded bootstrap nodes through which content routers can be learned.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,267 @@
...
+nodes with high connectivity in the network.
+
+### 1. content-routing as a libp2p protocol
+
+IPFS nodes will advertise and coordinate discover of content routers using a
+new libp2p protocol advertised as "/ipfs/content-router-discovery/1.0.0".
+
+The protocol will follow a request-response model.
+A node will open a stream on the protocol when it wants to discover new
+content routers it does not already know.
```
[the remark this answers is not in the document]
willscott: updated in the section below around how a node should rank which other peers to query.

IPIP/0342-content-router-discovery.md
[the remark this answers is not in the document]
lidel: > what prevents them from participating in this protocol as described?

CORS. If the only type of router this protocol returns is HTTP URL, then by default JS-IPFS running on a website won't be able to read data via cross-origin requests to the discovered router due to [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) limitations. 

We have two ways of solving the problem:
1.  (easy spec fix) Add a paragraph that requires `https://` servers returned by this discovery protocol to to ALWAYS  have  `Access-Control-Allow-Origin: *`  etc set up 
    - cc @guseggert for visibility, as we should include note about CORS headers in https://github.com/ipfs/specs/pull/337 too
1.  (more involved) Create libp2p version of https://github.com/ipfs/specs/pull/337 that browser peers could use over existing `/wss` or `/webtransport` listeners. Another argument [Why IPFS needs Delegated Routing over libp2p](https://www.notion.so/protocollabs/wip-Why-IPFS-needs-Delegated-Routing-over-libp2p-b634a4ec905c427fa41a97a3b3ba804e).

IPIP/0342-content-router-discovery.md
```diff
@@ -32,6 +32,20 @@ is also insufficient long term because:
 
 ## Detailed design
 
+This spec is designed for the ability of IPFS nodes to automatically discover
+and make use of 'content routers'. Content routers are services which are able
+to fulfill libp2p's [ContentRouting](https://github.com/libp2p/go-libp2p/blob/master/core/routing/routing.go#L26)
```
lidel, line 37: nit: this is not libp2p API, it is there due to legacy reasons (libp2p being part of go-ipfs long time ago).
afaik libp2p team removed specs for this from their repo and we have a gap  in IPFS one (see https://github.com/ipfs/specs/issues/345).

```suggestion
to fulfill IPFS [ContentRouting](https://github.com/libp2p/go-libp2p/blob/master/core/routing/routing.go#L26)
```

IPIP/0342-content-router-discovery.md
```diff
@@ -32,6 +32,20 @@ is also insufficient long term because:
...
+API. These routers currently are considered to directly support queries using
+the protocols specified by
+[IPIP-337](https://github.com/ipfs/specs/pulls)
```
lidel, line 40:
```suggestion
[IPIP-337](https://github.com/ipfs/specs/pull/337)
```

IPIP/0342-content-router-discovery.md
```diff
@@ -32,6 +32,20 @@ is also insufficient long term because:
...
+and/or
+[IPIP-327](https://github.com/ipfs/specs/pull/327).
```
lidel, line 42: We need to either decide which one is the future, or document how client decides which one to use for sending requests.
IF we do the latter, including content type in response is the way to go: Reframe endpoint is  `application/vnd.ipfs.rpc[..]; version=n`.

willscott:
I updated to hopefully address your review, @lidel 
* There's a more concrete description of what the proposed protocol will look like
* I added a query tag to support discovery of more than just content routing through the mechanism
* in terms of 'is this only for ipni' / are you 'punished' if you're incomplete - i think it's probably too early to try to predict the dynamics here without either pretty extensive modeling or subsequent user data. The ranking / propagation of routers would be impacting by what users are after - so if there's a community or application that's focused on a narrow subset that's well addressed by a different router, that router could also be very viable with this discovery mechanism.

IPIP/0342-content-router-discovery.md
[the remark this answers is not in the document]
aschmahmann: I'm confused as to how the current draft addresses the issues here:

> will have knowledge of the entire CID space

This line (from the spec) seems problematic even in the IPNI case. IPNI != the entire CID space which starts to make implementations complicated. The idea that one routing system is going to cover every use case is IMO not the way to go (and also the reason why there's even discussion of a delegated content routing API rather than just an IPNI API).

---

An example issue is to say we have 4 routers:
1. cid.contact (IPNI)
2. FilSwan (IPNI)
3. routing.delegate.ipfs.io (proxies DHT + IPNI)
4. ipfsdht.delegate.ipfs.io (proxies DHT, this is the only one that doesn't exist today, but certainly could)

While router 3 provides strictly more information than routers 1 or 2 it's also likely to be slower than them. It seems optimal to either contact 3 or contact 1/2 + 4 in parallel. For a node running its own DHT client would contact routers 1 or 2 and never contact 4. However, a naïve implementation may just result in all requests going through router 3 as it has the most CIDs covered which is not good. Perhaps a classification algorithm would be able to tease-out the optimizations here without further protocol adjustments, but that seems like a lot of complexity that could be alleviated by a small protocol adjustment.

This case also seems more problematic than the one that's been resolved by flagging router "type" like `content-routing`, `peer-routing`, etc. since support for a given API (content/peer routing) can be discovered with a single query to the endpoint whereas this requires a bunch of code complexity.

This seems like it'd be largely resolvable based on allowing users to query and return a set of named routing systems, or just calling this the "IPNI discovery system" so that routers like 3 + 4 know not to participate. I'd rather the former, but understand the latter.

While I wouldn't be surprised if down the road we also ended up requiring some of that  ML-style classification code anyway I suspect walking down that path now is premature and likely to cause us problems.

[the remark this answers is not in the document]
willscott: * It seems like there's a lot of complexity in expressing this 'composition complexity' either directly or through classification. we don't have this problem today so I would prefer to defer this sort of grammar to a subsequent IPIP - you worry that 3 would do better than the others, but I would argue that would be incentive for the IPNI team to build what i think you have previously called 'radar' to incorporate DHT results into IPNI such that 1,2,3 are all equal :)

* As you say, "IPNI != the entire CID space". I think it's a mistake to limiting our framing of this to an "IPNI discovery system" when it is simply discovering 'the most complete' available content routers. We're trying to be inclusive/general here - and I don't see huge harm in calling this content routers.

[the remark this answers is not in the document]
aschmahmann: > we don't have this problem today

I guess it depends what you mean by "today". I would like libraries like https://github.com/libp2p/js-libp2p-delegated-content-routing to switch to the latest content routing API (or have alternative which have switched) at which point js-libp2p in browsers should be able to leverage both the DHT and IPNI to get data from any peer that speaks wss/webtransport/webrtc.

Ideally they could use this protocol for discovery rather than hardcoding a DHT resolution endpoint.

@lidel could probably speak more about desired timings here.

> but I would argue that would be incentive for the IPNI team to build what i think you have previously called 'radar' to incorporate DHT results into IPNI such that 1,2,3 are all equal :)

That's cool and would certainly resolve at least this use case 😄.

---

> We're trying to be inclusive/general here
❤️

>  I think it's a mistake to limiting our framing of this to an "IPNI discovery system" when it is simply discovering 'the most complete' available content routers.

That's an interesting framing. By pushing for the "most complete set" it seems like you're essentially trying to get routers to compete for attention and content and make it so that only a single request to a single system needs to happen for clients to get what they need. If this is how the system evolves this is very nice to client machines.

However, if routers try and cut costs or code complexity by serving more-specific data (e.g. only data advertised over a specific pubsub channel, only data put in the IPFS Public DHT, only data from the BitTorrent network, ...) then the client code could start becoming problematic as it tries to figure out who to ask, without spamming all the routers.

I'm more cautious in advocating for the latter, but could see this go either way 😅. As long as the more immediate case around DHT + IPNI data being available to browser nodes is covered I'm happy 😄.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,340 @@
...
+
+### 2. probing of the discovery protocol
+
+A node will probe it's connected peers for content routing updates in two
+situations:
+
+1. When it needs to perform a content routing query, and has not
+successfully performed a sync in over a day.
+2. When it's auto-nat status indicates it is eligible to be a DHT server, and
+it has not successfully performed a synce in over a day.
```
guillaumemichel, line 165:
```suggestion
it has not successfully performed a sync in over a day.
```

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,340 @@
...
+noticable impact.
+2. nodes will only propagate content routers they believe to work,
+limiting the spread of spam / unavailable content routers to the directly
+connected peers of an adversary.
+
+With the exception of LAN tables, the other connections made by IPFS
+nodes do not have geographic locality. As a result, performance is
+separated in the tracking of content routers because it will not be
+effective as a ranking factor in the non-geographically-aware
+gossip system described here. As an optimization, nodes may choose to
```
guillaumemichel, line 243: gossiping should be geographically-aware and happen only between peers that are geographically close to each other. Otherwise, I may share a content router that is geographically close from me, but it will be too slow for you, and you won't use it at all.
So sharing content routers with geographically far peers becomes irrelevant, as long as we had enough content routers and that they are distributed around the globe.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,340 @@
...
+a global list that all clients see new providers, or it takes an inordinant
+amount of querying before a client happens to run into a provider, leading to
+degraded experiences for most clients. The single global list that a provider
+can automatically add itself to leads to issues for how to mitigate an
+enumeration of all network participants by a malicious content router.
+
+Pros:
+* Network is already there, no need to create a new protocol to "provide" new providers instead of CIDs.
+* You could potentially associate a provider with a specific root CID content.
+Cons:
```
guillaumemichel, line 327:
```suggestion
* You could potentially associate a provider with a specific root CID content.

Cons:
```

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,340 @@
...
+
+##### b. Availability Attacks / failing to provide records
+
+* if list of records insufficient, client will get more from other providers in subsequent queries, leading to downranking
+
+#### 2. Exposure of IPFS Clients (enumeration of network participants)
+
+* a new provider is only visible to directly connected peers. they only forward it to peers asking them if it meets their bar
+for reliability.  This means propogation through the network is only posisble for routers that behave correctly.
+* because clients only propagate their 'top' routers, latency is also relevant, and with sufficient number of routers, the would only
```
guillaumemichel, line 297:
```suggestion
* because clients only propagate their 'top' routers, latency is also relevant, and with sufficient number of routers, they would only
```

IPIP/0342-content-router-discovery.md
willscott: do we believe ipfs nodes will generally have enough knowledge to ambiently identify which peers are geographically close to them?

Winterhuman: https://github.com/libp2p/specs/pull/413 (GossipSub v1.2) would probably solve this, since it's all about minimising latency

willscott: we could use the same modeling / structure as gossip sub, but this is meant to be pull-based rather than push based. I have concerns about dropping in GossipSub directly.

IPIP/0342-content-router-discovery.md
```diff
@@ -99,12 +107,50 @@ to content discovery queries. By default this will be 10. (This default is
...
+Protocol messages are encoded using *cbor*. The following protocol examples demonstrate
+the schemas of requests and responses if they were to be encoded with JSON.
+
+A query on the "/ipfs/router-discovery/1.0.0" protocol will look like:
+```json
+{
+  "router": "string",
+  "filter": "bytes of the bloom filter"
+}
+```
```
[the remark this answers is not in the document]
mxinden: > (we could have a separate type for discovering routers that support peer routing).

Thus they would serve the same use-case as a [rendezvous](https://github.com/libp2p/specs/commits/master/rendezvous/README.md) server?

> thoughts on the use case and the wire format here?

I can not think of a project outside of the IPFS realm that is in need for this.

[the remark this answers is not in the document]
willscott: > Thus they would serve the same use-case as a [rendezvous](https://github.com/libp2p/specs/commits/master/rendezvous/README.md) server?

It extends the rendezvous protocol in two ways:
1. not requiring the need for a single hard-coded rendezvous point
2. adding reputational gossip in addition to just directory listing in the rendezvous protocol

IPIP/0342-content-router-discovery.md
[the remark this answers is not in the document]
lidel: +1 that delegated  DHT  is the core use case for IPFS  on Web Platform (JS in HTML in web browser)

- The gist of https://github.com/libp2p/js-libp2p-delegated-content-routing story is that is uses `/api/v0/dht` from Kubo RPC and we want to move away from that model. 
- Switching to HTTP API at `routing.delegate.ipfs.io` (proxies DHT + IPNI) is an easy win, and we would want to do this ASAP.
- Having ambient discovery via bootstrap nodes talking `/wss` and `/webtransport` will allow for basic resiliency / redundancy

IPIP/0342-content-router-discovery.md
lidel: Just noting that routers identified with DNS names  (cid.contact) could use things like Anycast to ensure the client is routed to the closest one. (I believe we already do that for ipfs.io gateway..)

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,341 @@
...
+API. These routers currently are considered to directly support queries using
+the protocols specified by
+[IPIP-337](https://github.com/ipfs/specs/pulls/337)
+and/or
+[IPIP-327](https://github.com/ipfs/specs/pull/327).
+
+In addition, this protocol expects that content routers that may be considered
+for auto-configuration/discovery by IPFS nodes will have knowledge of the
+entire CID space - in other words a delegation to such a router may be
+considered 'exhaustive'.
```
guseggert, line 55: How does this happen? What kind of consistency SLAs should indexers have, and how can they achieve it?

willscott: I'd like to say 'that's outside of this direct IPIP' - in that if routers fail to be consistent they would risk loosing priority.

In practice:
* indexers follow the list of providers from other indexers so that the constituents they follow are consistent
* they gossip announcements they see to each other so new updates are propagated between them
* [in progress] they can come to snapshot consensus periodically over a vector of providers & latest advertisements.

BigLep:
This came up in https://pl-strflt.notion.site/2023-05-30-Content-Routing-WG-12-b2ed74834fe44e359bbcdd02740e2084

There is going to be implementation need for this in the next quarter or two.  As a result, want to get ahead before code gets written and decisions ossify.

Some next steps:
1. Update the IPIP to [the newest template](https://github.com/ipfs/specs/blob/main/ipip-template.md) standard with lessons learned the last 6 months
  - I assume @willscott or someone from PL EngRes Bedrock team will do this.
2. Review from spec maintainers
  - I assume @lidel will take the first pass here on the updated doc when it's ready.

Implementation notes:
"Lassie currently does not depend on boxo. It may be hard for us to prototype ambient-discovery in boxo until the boxo/rest-of-world dependency conflicts are resolved"

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,341 @@
...
+rank peers by how recently a content routing discovery query has been make
+to that peer, with tie breaking preference for LAN nodes and for nodes
+with explicit peering agreements.
+
+Other factors that may be considered include:
+* Reputation of the peer, including how long it has been connected and if it
+  has served useful content in the past.
+* Latency / ping time of the peer.
+
+### 3. selection of routers
```
lidel, line 181: Note to self: this section of the spec should be more specific about "bare minimum reputation system", and provide enough for implementer to do the right thing, and not say clients do "as they wish".

Expected behavior on non-client services like bootstrappers should be also elaborated.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,341 @@
...
+known content routers past the 'uncertain' threshold, the node may
+choose to issue content routing queries for a fraction of the DHT
+lookup queries it receives as a way to maintain a more accurate
+table of content  routers.
+
+## Test fixtures
+
+TK is a CID currently only available through the content routing system,
+and not through the IPFS DHT. This is a piece of content that can be queried
+to validate the presence of alternative content routing systems.
```
lidel, line 218: Can't hardcode CID in the spec, it will stop being useful the moment someone announces it on DHT.

```suggestion
An unique CID needs to be created every time and made only available through the tested router, 
and not through any of the other ones, including the IPFS DHT. 

This piece of content can be queried to validate the presence of alternative content routing systems.
```

willscott: pointed to a implementation of a provider that can generate unique cid and provide it only over a configured IPNI endpoint as a basis for testing.

IPIP/0342-content-router-discovery.md
```diff
@@ -0,0 +1,357 @@
...
+}
+```
+
+A response is a list of entries, which looks like:
+
+```json
+[
+  {
+    "peer": "multiaddr.MultiAddr",
+    "score": float
```
lidel, line 151: - single `score` may be too vague. based on Rhea/Saturn alone,  we may want to track "lookup" score and "retrieval" score separately
- if we want this API to be not limited to IPNI, then we could have `type` per result

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+## Detailed design
+
+This spec is designed for the ability of IPFS nodes to automatically discover
+and make use of 'content routers'. Content routers are services which are able
+to fulfill IPFS's [ContentRouting](https://github.com/libp2p/go-libp2p/blob/master/core/routing/routing.go#L26)
+API. These routers currently are considered to directly support queries using
+the protocols specified by
+[IPIP-337](https://github.com/ipfs/specs/pulls/337)
+and/or
+[IPIP-327](https://github.com/ipfs/specs/pull/327).
```
SgtPooki, line 66: Would it be worth calling out these names as well, so users unfamiliar with the ipip-### dont have to click through

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+
+In addition, this protocol expects that content routers that may be considered
+for auto-configuration/discovery by IPFS nodes will have knowledge of the
+entire CID space - in other words a delegation to such a router may be
+considered 'exhaustive'.
+
```
SgtPooki, line 72: Is it possible to have an exhaustive list of the entire CID space?

Also, the wording here is hard to follow. Maybe something like:

```suggestion
In addition, implementers of this protocol may specify default content routers, but this protocol comes with a few expectations for default content-routers:

Default content routers configured by those implementing this protocol:

* MUST know of the
entire CID space - in other words, a delegation to such a router may be
considered 'exhaustive'

```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+### 0. content-router discovery state tracking
+
+Nodes will conceptually track a registry about known content routers.
+This registry will be able to understand for a given content router two
+properties:
```
SgtPooki, line 77:
```suggestion
Nodes will conceptually track a registry of known content routers.
This registry will maintain two properties for a given content router:
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+* reliability - how many good vs bad responses has this router responded
+with. This statistic should be windowed, such that the client can calculate
+it in terms of the last week or month. This will in practice be stored as
+daily buckets of successful and unsuccessful queries against a router, where
+success indicates that the router was queried, and the data was subsequently
+retrieved from a node returned as a provider by that router.
```
SgtPooki, line 83: Is it possible for a content router who returns a peer that *should* have content, to know that they successfully provided that content?

1. A asks B for providers of bafy1
2. B finds C and tells A about it
3. A asks C for bafy1.

Is this spec requiring A to tell B the result of item 3 above? How is this tracking accomplished?

src/ipips/ipip-0342.md
guillaumemichel: I also doubt that any single/federated entity will know the entire CID space. It may be possible now, but this certainly doesn't scale.

If we shard the CIDs on different instances, we basically get a DHT (with O(1) lookup if all shards fit in memory). However, it inherits the same weaknesses that the current DHT has, when providing a fair amount of CIDs, you need to open a connection to every single shard. To mitigate this, we might consider introducing intermediary services tasked with CID-to-shard allocation, streamlining access and reducing the number of direct connections needed.

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+The node will request routers from the peer that it does not already know.
+To express what it does know, it will query with a bloom filter. The
+statistical data structure provides a minimal amount of deniability around
+the routers that the client already knows.
+
+* The size of the bloom filter is chosen by the client. It is sized such
+that it has a greater than 99% certainly that it will receive a useful
+response. The maximum size of a query may be capped by the server, but can be
+effectively considered to be under 10kb.
+* The client will hash it's known content routers into the bloom filter
```
guillaumemichel, line 118: typo

```suggestion
* The client will hash its known content routers into the bloom filter
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+to set bits in the filter at the locations to which these known routers
+hash.
+* The server will have a parameter for a number of servers it wants to return
+to content discovery queries. By default this will be 10. (This default is
+picked as the result of modeling router propagation). It will iterate through
+it's list of known content routers, hashing them against the bloom filter and
```
guillaumemichel, line 124:
```suggestion
its list of known content routers, hashing them against the bloom filter and
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+selecting the top routers that are not already known to the client. It will
+return this list, along with it's reliability score for each.
```
guillaumemichel, line 126:
```suggestion
return this list, along with its reliability score for each.
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+    "peer": "/dns4/dev.cid.contact/tcp/443/https",
+    "router": "ipni",
+    "score": 0.90
+  },
+]
+```
+
+### 2. probing of the discovery protocol
+
+A node will probe it's connected peers for content routing updates in two
```
guillaumemichel, line 182:
```suggestion
A node will probe its connected peers for content routing updates in two
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+situations:
+
+1. When it needs to perform a content routing query, and has not
+successfully performed a sync in over a day.
+2. When it's AutoNAT status indicates it is eligible to be a DHT server, and
```
guillaumemichel, line 187:
```suggestion
2. When its AutoNAT status indicates it is eligible to be a DHT server, and
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+
+Content routers meeting the good reliability threshold are ordered by
+performance. the top one is queried, as is an 'uncertain' router if
+one exists.
+
+These threshold values are maintained for a year for the purposes
+of local selection.
+They are maintained for a month for the purpose of admitting
+knowledge of routers to others - so a client will no longer set bits for
+routers it is aware of but which do not meet it's threshold for 'good'
```
guillaumemichel, line 222:
```suggestion
routers it is aware of but which do not meet its threshold for 'good'
```

src/ipips/ipip-0342.md
```diff
@@ -0,0 +1,366 @@
...
+
+### 2. probing of the discovery protocol
+
+A node will probe its connected peers for content routing updates in two
+situations:
+
+1. When it needs to perform a content routing query, and has not
+successfully performed a sync in over a day.
+2. When its AutoNAT status indicates it is eligible to be a DHT server, and
+it has not successfully performed a sync in over a day.
```
guillaumemichel, line 188: What is the reasoning here?

PR #309, opened 2022-08-12
IPIP-309: IPNS Lagrange Replication

Winterhuman:
An IPIP proposal based on https://github.com/ipfs/specs/issues/268.

Issue #268, opened 2022-02-23
Self-maintaining IPNS Record Replication


Winterhuman:
# Self-maintaining Record Replication V2

See edit history for Self-maintaining Record Replication V1.

## The Problem

IPFS has some challenging attributes which makes storing content for long periods of time without maintenance difficult:

- Most IPFS nodes don't stay online for long.
- Many IPFS nodes will clear their caches over time or will just disappear forever.
- All IPFS nodes will only store content they've explicitly requested, the only exception being DHT entries.

So, if we have an IPNS record, which we want to remain accessible to anyone even after the publisher has stopped republishing records under it's key, how do we achieve this?

Related Issues: https://github.com/ipfs/kubo/issues/8542 https://github.com/ipfs/kubo/issues/4435 https://github.com/ipfs/kubo/issues/3117 https://github.com/ipfs/kubo/issues/1958

## The Aim

The goal of this proposal is to achieve the following:

- Distribute IPNS records amongst a dynamic set of peers, this peer set will change as record holders disappear and become inaccessible, the overall effect is that the IPNS record never disappears completely.
- Let IPNS record holders be **optionally** responsible for updating this dynamic set of peers such that, once some lower threshold is reached, all responsible record holders will bring the number of record holders above the threshold again.
- Achieve all of the above without significantly increasing background bandwidth usage from record holders, even for those responsible for updating the peer set, this follows a similar principle to CircuitV2 to ensure this system is as widely used as possible.
- And, most importantly, remove the need for IPNS record expiration while still mitigating replay attacks as much as possible.

So, let's outline the scenario:

- Node A wants to publish a CID under `k51key`.
- Node A finds nodes `12D3KooWkay`, `12D3KooWkez`, and `12D3KooWbey` who are willing to add the IPNS record to their DHT. These nodes are now record holders.
- And now, Node A disappears; no one is able to publish any new IPNS records under `k51key`, however, the IPNS record should still be around for as long as possible.
- What can the record holders do?

## Record Holder Behaviour

Each IPNS record holder needs to consider the following two questions:

- What is the minimum number of record holders (lower threshold) that needs to be reached before I update the peer set?
- Which nodes should be in the dynamic set of peers and by what criteria?

Let's look at question 1...

For record holders which have not opted in to maintaining the peer set above the threshold, their lower threshold is `0` meaning they don't check the threshold and they don't maintain the peer set, this is analogous to `Routing: dhtclient`.

[...]

PR #349, opened 2022-11-27
IPIP-309: IPNS Lagrange Replication

Winterhuman:
Reopened from https://github.com/ipfs/specs/pull/309 in it's own branch

### DIFF
--- a/IPIP/0000-ipns-lagrange-replication.md
+++ b/IPIP/0000-ipns-lagrange-replication.md
@@ -74,7 +74,7 @@ IPNS records won't have anything set for the expiration field, this may interfer
 
 ### Security
 
-To prevent old IPNS records from being replicated, the optional 5th step stated in https://github.com/ipfs/specs/blob/main/naming/pubsub.md#protocol should be followed. This should make replay attacks difficult to execute.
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
The idea worth exploring, some initial comments below and inlined.

- formalizing how peers other than original publisher can keep records alive sounds good, but this IPIP should clearly state in "Compatibility" section when IPNS reproviding starts and when it stops
  - idea: mirror how regular provider records work:  if CID gets garbage-collected, my node stops providing it. same could be applied to reproviding IPNS record (e.g. if root CID from the record is not pinned, providing stops when CID that IPNS record poitns at is garbage-collected). 
- introduced threshold and interval knobs need some default values – ok to say "follow value from dht spec", but it needs to be explicitly stated
- removing expiration feels controversial and orthogonal to the rest of IPIP
  -   perhaps lifting expiration should be a separate IPIP (with detailed replay attack analysis)?

Winterhuman:
Closing for now, I don't really have the capability to move this pull further, replay attack analysis and an implementation are beyond what I can do, I'll leave https://github.com/ipfs/specs/issues/268 open as the idea

Issue #274, opened 2022-03-27
IPNS Record Owner Optimisation


Winterhuman:
Here's the current situation when fetching the latest IPNS record:

- Peer `12D3KooWowner` creates an IPNS record `k51key...` and distributes it through the network.
- Peer `12D3KooWresolver` wants to find the latest version of `k51key...`.
- `12D3KooWresolver` searches the network and finds `n` peers who advertise `k51key...`, they then fetch the record from `n` number of peers and check which record has the highest sequence value.

The problem with this system is that it's very slow needing multiple peers to verify that you've fetched the latest record, however, there is an optimisation for speeding this up for when one of the `n` peers happens to be the record owner. Here's how it might work:

- Peer `12D3KooWowner` creates an IPNS record `k51key...` and distributes it through the network.
- Peer `12D3KooWresolver` wants to find the latest version of `k51key...`.
- `12D3KooWresolver` searches the network and finds `n` peers who advertise `k51key...`. Instead of trying to fetch `n` number of records, they start by fetching a few at random (let's say three for this example).
- `12D3KooWresolver` reads the record to find PeerID `12D3KooWowner`, they then check to see if any of the `n` peers they found have PeerID `12D3KooWowner`.
- If one of the 3 random peers was the record owner, the search can immediately stop since the record owner's copy of `k51key...` is guaranteed to be the latest and has already been fetched.
- If one of the `n-3` peers is the record holder, `12D3KooWresolver` can attempt to fetch `k51key...` directly from them. If it succeeds, the search can immediately stop since the record owner's copy of `k51key...` is guaranteed to be the latest.
- If `12D3KooWowner` isn't any of the `n` peers, or the attempt to fetch `k51key...` from `12D3KooWowner` failed, `12D3KooWresolver` will continue to fetch all remaining records to compare them as before.

Because IPNS records expire, record owners are likely to be online most of the time in order to ensure the record doesn't disappear from the network, this means the chance of finding the record holder amongst the `n` peers will be relatively high in the current environment.

Winterhuman:
Turns out you actually mentioned something very similar to this about 4 years ago: https://github.com/libp2p/go-libp2p-kad-dht/issues/145#issuecomment-410831110

Winterhuman:
Having an `IpnsEntry.trusted` field for defining an array of PeerIDs to trust seems like a possible way forward, what do you think? @Stebalien @lidel

Context: https://github.com/ipfs/specs/pull/319

[...]

PR #350, opened 2022-11-27
Mention record owner optimisation

Winterhuman:
Related: https://github.com/ipfs/specs/issues/274

Changes IPNS.md to mention that the owner of a record is implicitly trusted to hold the most recent record, and implementations MAY skip the sequence value comparison in the case where the provider of the record is the record owner.


rinzler100 (approved):

Jorropo:
That an intresting thought to do, but does an implementation actually does this ?

I don't belive Kubo is that smart.

Jorropo:
And you would need to write an IPIP because there are side effects of such things that we need to discuss (multiple nodes using the same peerid):
https://github.com/ipfs/specs/blob/main/IPIP/0000-template.md

Winterhuman:
Closing as this needs an IPIP

Issue #345, opened 2022-11-15
IPFS DHT Specification is  missing


lidel:
This is a placeholder for missing specs.

## Problem

- No pecs for IPFS DHT exist.
- Kad DHT specs that live in https://github.com/libp2p/specs/tree/master/kad-dht and are currently focused on Peer routing only (scope limited to libp2p use of DHT).
- IPFS specifics such as how provider records and IPNS records are published  / resolved was removed from the mentioned specs, but should be still  in git history of that repo 

## TODO 
- [ ] see if we can salvage  any old content from https://github.com/libp2p/specs/tree/master/kad-dht
- [ ] decide if our spec is on top of libp2p one, or do we have our own, end-to-end specs 
- [ ] WRITE THE SPECS and get sign-off  from GO, JS and Rust impl.

SgtPooki:
@lidel will this IPIP also include AcceleratedDHTClient spec? or should we create a separate issue for that?

lidel:
tbd, not part of protocol per se, just an efficient way of architecting client.  

this is similar to Bitswap Sessions - could be an Appendix "Notes for implementers", or a separate file to avoid noise

PR #354, opened 2022-12-13
docs: validity follows expiration window

hacdias:


lidel (approved):
Thanks!  This is useful clarification. 

(in the future we will refer to expiration value from IPFS DHT specs instead, [once we have them](https://github.com/ipfs/specs/issues/345))

PR #355, opened 2022-12-15
IPIP-0000: IPFSClient API

markg85:
Draft for sure!

This spec is intended to specify how a "ipfsclient" works.
It relies on 3 different specs:
1. [IPIP-0280](https://github.com/ipfs/specs/pull/280) (gateways file)
2. IPIP-0000 (multi gateway client, spec isn't on github yet. This spec defines how gateways are determined to be used)
3. IPIP-0000 (using gateways, spec isn't on github yet. This spec defines how an application can find and use gateways)

The missing specs will be on github later. Will update this message when they are online.


markg85:
**Forget this one! I renamed the branch.**

PR #356, opened 2022-12-15
IPIP-356: IPFSClient API

markg85:
Draft for sure!

This spec is intended to specify how a "ipfsclient" works.
It relies on 3 different specs:
1. [IPIP-0280](https://github.com/ipfs/specs/pull/280) (gateways file)
2. [IPIP-0000 (multi gateway client. This spec defines how gateways are determined to be used and use them)](https://github.com/ipfs/specs/pull/359)
3. [IPIP-0000 (using gateways. This spec defines how an application can find and use gateways)](https://github.com/ipfs/specs/pull/360)

The missing specs will be on github later. Will update this message when they are online.

Issue #348, opened 2022-03-07
Create IPIP with Gateway spec for partial CAR exports


lidel:
## Context

https://github.com/ipfs/go-ipfs/pull/8758 adds support for CAR export via Gateway.
It exports entire dag as a CAR stream, which does not cover all use cases.

For example, thin clients may want to export unixfs directory root block + its immediate children, or progressively fetch a big DAG from multiple gateway endpoints.

### Why we need selector support

- Verifiable HTTP Gateway Responses (https://github.com/ipfs/in-web-browsers/issues/128) 
  - for mobile web browsers (content integrity without battery drain caused by full p2p)
    - mobile browser should be able to traverse huge  unixfs directory tree without having to fetch everything (only root block + root blocks of immediate children are needed for generating useful dir listing)
  - for IoT devices and other thin clients
    - fetching bigger DAGs progressively, load-balancing/falling back if some gateways are too slow/unreliable – makes HTTP more useful and pushes back the moment when an expensive p2p retrieval  has to be spawned 

## Scope

- [x] query param
- [x] HTTP header
- [ ] TBD configurable size budget for CAR stream + UnixFS downloads 
- [ ] TBD allow selectors everywhere? (UnixFS? dag-cbor/json?)

## Proposed design (A) :anger:

The go-car library  supports passing selectors, the idea is to add a parameter  to do just that.

We have to URL-escape selector somehow, either way,
so the choice is between encodeURIComponent and multibase encoding:

Text (JSON) representation:

```
/ipfs/{cid}?format=car&selector.json=encodeURIComponent({json serialization of selector})
```

Binary (CBOR) representation:
```
/ipfs/{cid}?format=car&selector.cbor=multibase({cbor serialization of selector})
```

## Proposed design (B)  :anger:  

```
/ipfs/{cid}?format=car&selector={cid2}
```

Here `{cid2}` is a CID representing selector data. It could be dag-cbor, dag-json.
Small ones could be inlined (with identity hash), bigger ones could be fetched once and reused efficiently.

## Proposed design (C) :pinching_hand: 

Over time, we realized this is the most useful and safest way.
No selector CIDs, only predefined, most useful "partial CAR export scope" parameters  for now:

```
/ipfs/{cid}/some/subpath/file?format=car&dag-depth=1&include-path=true
```

- `depth=1`  means "root+direct children only" – good for fetching UnixFS dir listing with file sizes / types, or splitting bigger DAGs into partial retrievals over multiple gateways / threads
[...]

PR #359, opened 2022-12-16
IPIP-359: Multi gateway client

markg85:
A spec to describe how multi gateway clients - formally known as racing gateways - should behave. This is very much a companion spec to https://github.com/ipfs/specs/pull/356.

To refine it's place.
This spec describes how multi gateway clients work and should be implemented.
https://github.com/ipfs/specs/pull/356 would have this in it's implementation.

### DIFF
--- a/integrations/MULTI_GATEWAY_CLIENT.md
+++ b/integrations/MULTI_GATEWAY_CLIENT.md
@@ -1,8 +1,8 @@
-# IPIP 0000: Multi gateway client
+# IPIP-359: Multi gateway client
 
 - Start Date: 2022-12-16
 - Related Issues:
-  - None
+  - https://github.com/ipfs/specs/pull/359
 - Relies on specs:
   - IPIP-0280
 


integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+- Relies on specs:
+  - IPIP-0280
+
+## Summary
+
+A defined way for getting a list of usable gateways.
+
+## Motivation
+
+When developing an application with IPFS functionality you'd ideally want more then 1 gateway and distribute the requests among N gateways. This spec relies on IPIP-0280 (gateways file).
```
meandavejustice, line 15: Leaving a note to remind you to link to IPIP-0280 once it is merged

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+      G --> H[Store gateway];
+```
+
+The `200ms` threshold here is arbitrarily picked. From a decentralized point of view, 200ms allows you to go roughly halfway across the globe assuming your internet connection is stable. From a data retrieval point of view 200ms can be slow but can be just fine too. For example, if a site loads with 1 connection at a time with each connection having a 200ms latency then you will experience that site to "load slow". But if you load the same site with multiple concurrent connections where "some" might hit the 200ms threshold then you won't see much difference.
+
+Once a gateway, in the above flow, reaches the `Store gateway` point then it should internally in this `multi gateway client` implementation be stored in some array.
+
+### Keeping the usable gateway list fresh in the background
+
+Getting this list of gateways and maintaining if they should be used can take quite some time. The adviced approach here is to run each request in an async matter where the async flow follows the same flow as the above flowchart.
```
meandavejustice, line 71: adviced => advised*

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Configuration options
+
+`concurrent requests` Defaults to 10. There must be a way to specify how many concurrent requests the `multi gateway client` does per IPFS request.
```
meandavejustice, line 75: I think default should be 6 requests here per => https://docs.diffusiondata.com/cloud/latest/manual/html/designguide/solution/support/connection_limitations.html#:~:text=Most%20modern%20browsers%20allow%20six,with%20any%20server%20or%20proxy.

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+`max simultaneous cids` Defaults to 5. There must be a way to define how many simultaneous IPFS requests the `multi gateway client` can handle at any given time.
+
+`max total gateways in use` Defaults to 25. There must be a way to specify how many total gateways can be used for the `multi gateway client` as a whole.
```
meandavejustice, line 79: What is the reason for having a limit here? Seems to me that more would always be better

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+`racing` Defaults to false. There must be a way to specify if `racing` should be used. Racing means the `multi gateway client` will ask at most the number of `concurrent requests` to all download the same data. The one who downloads it first if the one whose output is used, the rest is ignored.
+
+`verify raw` Defaults to true. This tells the `multi gateway client` implementation to verify RAW data as wel as CAR data. Setting this option to true (the default) means the `multi gateway client` is guaranteed to only give back valid data. If this option is set to false then raw data is returned as-is, unverified.
```
meandavejustice, line 83: typo: wel => well

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+These options are set on a `multi gateway client` level and apply to each request. If no options are specified the defaults as listed should be applied.
+
+### Request method
+
+There must be a method to allow IPFS data retrieval. The input for this method must be an IPFS url in these forms: `ipfs://<cid>` and `ipns://<cid>`.
+
+The data retrieval for a given CID must adhere to the configuration options.
+
+There must be an async way to get the data represented by that CID. While the `multi gateway client` can handle any CID data, in it's default settings all data is being verified. If `verify raw` is set to false then raw data is passed back as-is. CAR data is always verified.
```
meandavejustice, line 93: Is there any usecase to have CAR data returned without verifying? Probably not, but if so we should include an option for that as well

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Keeping the usable gateway list fresh in the background
+
+Getting this list of gateways and maintaining if they should be used can take quite some time. The adviced approach here is to run each request in an async matter where the async flow follows the same flow as the above flowchart.
+
+### Configuration options
+
+`concurrent requests` Defaults to 10. There must be a way to specify how many concurrent requests the `multi gateway client` does per IPFS request.
+
+`max simultaneous cids` Defaults to 5. There must be a way to define how many simultaneous IPFS requests the `multi gateway client` can handle at any given time.
```
meandavejustice, line 77: Is there a reason "5" was chosen here? It may make sense to set to 6 also for the reasons [above](https://github.com/ipfs/specs/pull/359/files#r1055893571)

meandavejustice:
I think there should be more info on the defined behavior when `racing` is set to false, especially since it is the default.

meandavejustice:
Leaving a note so that we remember to capture https://github.com/ipfs/specs/pull/356#issuecomment-1355385624

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+Users, applications using IPFS, get a defined way to find gateways to use.
+
+### Compatibility
+
+N/A
+
+### Security
+
+N/A
```
lidel, line 153: - Note that clients must verify received blocks before using them, and discard ones which do not match expected CID.
- If `ipns://` is to be supported, note if  / how  to handle DNSLinks

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+## Design rationale
+
+### User benefit
+
+Users, applications using IPFS, get a defined way to find gateways to use.
+
+### Compatibility
+
+N/A
```
lidel, line 149: Refer to 
- existing gateway specs, namely [one that returns block and car](https://github.com/ipfs/specs/blob/main/http-gateways/TRUSTLESS_GATEWAY.md), which is implemented by Kubo >=0.13.
- CARv1 specification (if you use it)

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+As an aside, a request is what the user puts in. For example:
+
+```cpp
+<client>::request("ipfs://bafyA");
+<client>::request("ipfs://bafyB");
+```
+
+Is 2 requests. These count at `max simultaneous cids` where the default is 5 maximum. If there are more then `max simultaneous cids` then those that don't get handled will be put on a queue to be handled as soon as a slot becomes available.
+
+Internally that CID is represented by N different CIDs (each block). Say `bafyA` consists of 100 blocks (simplified depiction):
```
lidel, line 104: Does this mean the client  always sends the first request for a single block, deserialize it, and then send CAR request for its branches? This is fine for MVP i guess, but it is hard to make a good decision when to swith from block to CAR for a deeper DAG.

Hannah made a demo during MTB5 and had some good ideas about adding option to fetch CAR with non-leave blocks first (metadata), and then fetching leaves with actuald ata at the end  – wrote some notes in https://github.com/ipfs/specs/issues/348#issuecomment-1387627152. It also included byte range requests, which are important for use cases like video seeking.

I feel we should strongly consider adding these parameters to CAR  requests, before this IPIP is finalized.

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
...
+
+### Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
+
+## Open questions
+
+### CAR verification file
+
+Besides verifying for response headers, we should also define which blob we actually expect. Like a "Hello world" or "Hello IPFS".
```
lidel, line 167: for a quick heartbeat check, a CAR with single root for a zero-length block will be enough, and won't waste much bandwidth

integrations/MULTI_GATEWAY_CLIENT.md
lidel: Maybe mention https://github.com/ipfs/specs/blob/main/http-gateways/PATH_GATEWAY.md#only-if-cached-head-behavior as mechanism for prioritizing gateways which already have the data? Shotgunning fetch request to 5 gateways is wasteful.

integrations/MULTI_GATEWAY_CLIENT.md
```diff
@@ -0,0 +1,175 @@
+# IPIP 0000: Multi gateway client
+
+- Start Date: 2022-12-16
+- Related Issues:
+  - None
```
lidel, line 5:
```suggestion
  - https://github.com/ipfs/specs/pull/359
```

PR #370, opened 2023-01-24
Reduce the scope of IPIP-337 by excluding write operations

masih:
Reduce the scope of IPIP-337 by temporarily excluding write operation. The write operations are to be added in a separate PR once the IPIP-337 is merged.

See:
 - https://github.com/ipfs/specs/pull/337

### DIFF
--- a/routing/DELEGATED_CONTENT_ROUTING_HTTP.md
+++ b/routing/DELEGATED_CONTENT_ROUTING_HTTP.md
@@ -3,6 +3,7 @@
 ![wip](https://img.shields.io/badge/status-wip-orange.svg?style=flat-square) Delegated Content Routing HTTP API
 
 **Author(s)**:
+
 - Gus Eggert
 
 **Maintainer(s)**:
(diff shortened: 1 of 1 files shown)


lidel (approved):
Thank you @masih, merging, let's continue in https://github.com/ipfs/specs/pull/337

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -182,61 +158,12 @@ Schema: `bitswap`
...
-```json
-{
-    "Protocol": "transport-bitswap",
-    "Schema": "bitswap",
-    "Signature": "<signature>",
-    "Payload": "<payload>"
-}
-```
-
-- `Signature`: a multibase-encoded signature of the sha256 hash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` JSON
```
lidel, line 197: > - `Signature`: a multibase-encoded signature of the sha256 hash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` JSON. Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -182,61 +158,12 @@ Schema: `bitswap`
...
-}
-```
-  - `Keys` is a list of the CIDs being provided
-  - `Timestamp` is the current time
-  - `AdvisoryTTL` is the time by which the caller expects the server to keep the record available
-    - If this value is unknown, the caller may use a value of 0
-  - `ID` is the peer ID that was used to sign the record
-  - `Addrs` is a list of string-encoded multiaddrs
-
-A 403 response code should be returned if the signature check fails.
```
lidel, line 216: > A [400 Bad Request](https://httpwg.org/specs/rfc9110.html#status.400)  response code should be returned if the signature check fails.

lidel (commented):
@masih  when you open IPIP with these PUTs specs, remember to re-apply below suggestions originally proposed on IPIP-337:

masih:
Thanks @lidel I'll open the `PUT` IPIP after #337 is merged.

PR #372, opened 2023-01-25
feat: remove reframe

guseggert:
This predated the IPIP process and as a result it doesn't really fit here since it doesn't have any current users.


lidel (approved):
ok-crying

cc @aschmahmann for visibility, in case we ever need to dig this out from the git archive

PR #374, opened 2023-02-01
fix typo in IPNS.md ‘IpfsEntry’ -> ‘IpnsEntry’

gobengo:
afaict ‘IpfsEntry’ (with the ‘f’) is ambiguous, as it appears nowhere else in the document. This was probably a typo?

😅


lidel (approved):
yes, its a typo, thank you @gobengo :pray:

PR #378, opened 2023-02-13
Reintroduce provide records via `PUT` to HTTP delegated routing

masih:
Previous work reduced the scope of IPIP-337 to read operations only. The changes here re-introduce the write operations originally written by @guseggert back into IPIP-337.

The specification documents the ability to provide Bitswap records over `PUT` requests with advisory TTL. An implementation of the specification is already present in go-libipfs.

See:
 - https://github.com/ipfs/specs/pull/370


routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +190,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+- `AdvisoryTTL` is the time by which the caller expects the server to keep the record available
+  - If this value is unknown, the caller may use a value of 0
+- `ID` is the peer ID that was used to sign the record
+- `Addrs` is a list of string-encoded multiaddrs
+
+A [400 Bad Request](https://httpwg.org/specs/rfc9110.html#status.400)  response code should be returned if the signature check fails.
+
+Note that this only supports Peer IDs expressed as identity multihashes. Peer IDs with older key types that exceed 42 bytes are not verifiable since they only contain a hash of the key, not the key itself.
+Normally, if the Peer ID contains only a hash of the key, then the key is obtained out-of-band (e.g. by fetching the block via IPFS).
+If support for these Peer IDs is needed in the future, this spec can be updated to allow the client to provide the key and key type out-of-band by adding optional `PublicKey` and `PublicKeyType` fields, and if the Peer ID is a CID, then the server can verify the public key's authenticity against the CID, and then proceed with the rest of the verification scheme.
```
lidel, line 230: Any reason why we don't include `PublicKey` field from the start?
If we make it an optional, opaque bytes field, and say it should be deserialized as the protobuf from [libp2p peerid](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#keys) specs, we have both solved RSA problem, and don't need to maintain `PublicKeyType` registry.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +190,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+
+The `Payload` field is a string, not a proper JSON object, to prevent its contents from being accidentally parsed and re-encoded by intermediaries, which may change the order of JSON fields and thus cause the record to fail validation.
```
lidel, line 232: If we have to mangle `Payload` JSON anyway, what  is the benefit of introducing it as a new struct?
It was probably discussed before elsewhere, but why can't we use exactly the same protobuf and signature DHT uses?

Reusing DHT record payload would save CPU time on IPFS nodes that publish to both DHT and IPNI  (signature would be generated only once, not twice).

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -118,7 +160,7 @@ limits, allowing every site to query the API for results:
 
 ```plaintext
 Access-Control-Allow-Origin: *
-Access-Control-Allow-Methods: GET, OPTIONS
+Access-Control-Allow-Methods: GET, PUT, OPTIONS
 ```
 
 ## Known Transfer Protocols
```
lidel, line 166: :thought_balloon:  this section grows way faster than the rest of the spec.

Would it be ok to move "Known Transfer Protocols" to a separate file (`KNOWN_TRANSFER_PROTOCOLS.md`)?

lidel (changes requested):
Quick first pass with some questions.

Also, please add quick memo to `IPIP/` directory based on  https://github.com/ipfs/specs/blob/main/IPIP/0000-template.md (fine to keep its sections  as TODO, we can fill content later)

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: > what is the benefit of introducing it as a new struct?

I suspect the reason for this is better human readability? The beginning of this document (merged as part of IPIP-337 states "human-readable encodings of types are preferred".

> save CPU time on IPFS nodes

We can make the same argument about HTTP delegated routing itself. Unless there is evidence to suggest that we should be heavily optimising this, my vote would be to favour human readability.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: I recommend doing this in a separate PR to reduce lines changes what are irrelevant to what this PR is trying to introduce.

If this is not a blocker for this PR, I'd be happy to take this as a TODO and get it done once this PR is merged.

masih:
>  please add quick memo to IPIP/

Thanks @lidel for the review. I have added a `.md` file under `/IPIP`; is that what you had in mind?

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: > Reusing DHT record payload

@lidel can I ask where I can find the DHT record payload specification? I think we can add this to the spec via a specific request `Content-Type` media type, similar to #379.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+    "Protocol": "transport-bitswap",
+    "Schema": "bitswap",
+    "Signature": "<signature>",
+    "Payload": "<payload>"
+}
+```
+
+- `Signature`: a multibase-encoded signature of the sha256 hash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` JSON. Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
+
+  - Servers may ignore this field if they do not require signature verification.
```
lidel, line 205: @masih is cid.contact doing verification of this?

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
 The server should respect a passed `transport` query parameter by filtering against the `Addrs` list.
 
+#### Bitswap Write Provider Records
+
+```json
+{
+    "Protocol": "transport-bitswap",
+    "Schema": "bitswap",
+    "Signature": "<signature>",
+    "Payload": "<payload>"
```
lidel, line 199: :thought_balloon:  this creates one `bitswap` schema for `GET` and the other `bitswap` for `PUT`. Both under the same name, but with different fields. Bit nasty and error prone.

@masih is this ossified, or is there still time to fix this and use different schema for PUTs?

```suggestion
    "Schema": "bitswap-announce",
    "Signature": "<signature>",
    "Payload": "<payload>"
```

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+```
+
+- `Signature`: a multibase-encoded signature of the sha256 hash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` JSON. Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
+
+  - Servers may ignore this field if they do not require signature verification.
+- `Payload`: a string containing a serialized JSON object which conforms with the following schema:
+
+  ```json
+   {
+       "Keys": ["cid1", "cid2"],
```
lidel, line 210: @masih what is the max length of this list?

lidel (commented):
@masih I'm picking up the work here, thank you for your patience, we had some resourcing issues. 

This PR needs to be rebased on top of `./src/routing/http-routing-v1.md` and the updated [IPIP template](https://github.com/ipfs/specs/raw/main/ipip-template.md) and moved to `./src/ipips/` but that can be done later.

For now,  quick  questions below :pray:

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+       "Timestamp": 0,
+       "AdvisoryTTL": 0,
+       "ID": "12D3K...",
+       "Addrs": ["/ip4/..."]
+   }
+   ```
+
+- `Keys` is a list of the CIDs being provided
+- `Timestamp` is the current time
```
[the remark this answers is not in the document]
BigLep: Agreed on following standardized format that is easily human-parseable and doesn't have ambiguity.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+- `AdvisoryTTL` is the time by which the caller expects the server to keep the record available
```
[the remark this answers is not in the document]
BigLep: If a duration, maybe get that into the name.
Also embed the units in the parameter name so there is no ambiguity?

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+  - If this value is unknown, the caller may use a value of 0
+- `ID` is the peer ID that was used to sign the record
```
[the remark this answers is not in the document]
masih: No concerns 👍

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: There is still time. Changing thanks

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: cid.contact does not support any PUT over http delegated routing.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
masih: > having dedicated PUT for this one protocol feels odd, PUT for HTTP will look exactly the same

I agree :+1:

masih (commented):
@lidel this PR has aged long enough that it seems it needs to be rewritten, specially to accommodate transport agnostic publication of CIDs. There is the missing DHT record specs issue but i don't think that's a blocker for now while we iterate over this?

If you agree I revise this PR entirely, considering the original was extracted from IPIP-337 written by Gus.

Thoughts?

lidel:
@masih yes fine to revise this :pray:  (either same PR, or close this one and open a new one using https://github.com/ipfs/specs/raw/main/ipip-template.md to start with clean slate if that's easier)

lidel:
@masih just checking-in, how does the future of this IPIP look like?  (asking in context of https://github.com/ipni/index-provider/issues/403#issuecomment-1749308418)
Something you or someone from IPNI is willing to DRI, or is it ok to park till 2024 or until someone from other team have bandwidth to take over?

(fwiw we want to support PUTs in `/routing/v1` exposed by Kubo, but this is most likely a 2024 roadmap item)

masih:
> park till 2024 or until someone from other team have bandwidth to take over

Yes please. I don't have bandwidth to work on this just now unfortunately.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
```diff
@@ -148,6 +189,60 @@ Specification: [ipfs/specs/BITSWAP.md](https://github.com/ipfs/specs/blob/main/B
...
+```json
+{
+    "Protocol": "transport-bitswap",
+    "Schema": "bitswap",
+    "Signature": "<signature>",
+    "Payload": "<payload>"
+}
+```
+
+- `Signature`: a multibase-encoded signature of the sha256 hash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` JSON. Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
```
lidel, line 203: I assume we mean multihash here, and DAG-JSON for determinism across languages and to avoid quirks of JSON libraries?
```suggestion
- `Signature`: a multibase-encoded signature of the sha256 multihash of the `Payload` field, signed using the private key of the Peer ID specified in the `Payload` DAG-JSON. Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
```

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
lidel: After [IPIP-417](https://github.com/ipfs/specs/pull/417) we are in a better place now. 

I think we should avoid protocol-specific schema for bitswap and use peer schema from [IPIP-417](https://github.com/ipfs/specs/pull/417). It allows for passing optional protocol-specific metadata for things other than bitswap, allowing client to do single PUT with all info.

routing/DELEGATED_CONTENT_ROUTING_HTTP.md
willscott: I'm okay with either of these paths.
@masih  are you okay with this suggestion?

masih: Seems fine to me 👍

src/routing/http-routing-v1.md
```diff
@@ -106,17 +106,27 @@ Each object in the `Providers` list is a record conforming to a schema, usually
...
+
+Each object in the `Providers` list is a *write provider record* entry.
+
+Server SHOULD accept  representing writes is [Announcement Schema](#announcement-schema).
+
+:::warn
+
+TODO: is below a sensible limit?
+
+There SHOULD be no more than 100 `Providers` per request.
```
willscott, line 127: this distinction implies we won't want to use this same protocol to sync overall routing tables between nodes - if there are more than 100 known providers for data, would we expect a different protocol to be used?

src/routing/http-routing-v1.md
```diff
@@ -130,9 +140,8 @@ Each object in the `Providers` list is a *write provider record*.
 
 - `ProvideResults` is a list of results in the same order as the `Providers` in the request, and the schema of each object is determined by the `Protocol` of the corresponding write object
   - This may contain output information such as TTLs, errors, etc.
-  - It is undefined whether the server will allow partial results
+  - It is undefined whether the server will allow partial results <!-- TODO: maybe add Error field to `announcement` schema and use it in responses ?  -->
```
willscott, line 143: I'd defer specification of partial errors until we have implementations encounter that problem?

src/routing/http-routing-v1.md
```diff
@@ -318,82 +355,104 @@ the case, the field MUST be ignored.
...
+      "CID": ["cid1", "cid2", ...],
+      "Scope": "block",
+      "Timestamp": "YYYY-MM-DDT23:59:59Z",
+      "TTL": 0,
+      "ID": "12D3K...",
+      "Addrs": ["/ip4/...", ...],
+      "Protocols": ["foo", ...],
+      "Metadata": "mbase64-blob",
+    },
+    "Signature": "mbase64-signature"
```
willscott, line 376: signatures over blocks of announcements end up being awkward, because the intermediate server can't provide a compact provider record onwards that a subsequent node can verify.

I guess the use of double-hashed content routing records gives us a better base for partial authenticity of individual records.


In what cases are we hoping for transient verifiability of blocks of http put routing records?

src/routing/http-routing-v1.md
lidel: We could:
- Raise limit on the server
- Ask client to make multiple PUTs
  - or use streaming PUT with `application/x-ndjson`  (see https://github.com/ipfs/specs/pull/378#discussion_r1413292096)

src/routing/http-routing-v1.md
lidel: Yes, tbd. 
This will be something we have to specify if we want to implement `application/x-ndjson` for streaming  PUTs in https://github.com/ipfs/boxo/tree/main/routing/http/client

src/routing/http-routing-v1.md
```diff
@@ -277,6 +355,81 @@ the case, the field MUST be ignored.
...
+  - `ID` is Peer ID of the node that provides the content and also indicates the `libp2p-key` that SHOULD be used for verifying `Signature` field.
+  - `Addrs` is an a list of string-encoded multiaddrs without `/p2p/peerID` suffix.
+  - `Protocols` is a list of protocols supported by `ID` and/or `Addrs`, if known upfront.
+  - `Metadata` is a string with multibase-encoded binary metadata that should be passed as-is
+- `Signature` is a string with multibase-encoded binary signature that provides integrity and authenticity of the `Payload` field.
+  - Signature is created by following below steps:
+    1. Convert `Payload` to deterministic, ordered [DAG-JSON](https://ipld.io/specs/codecs/dag-json/spec/) map notation
+    2. Prefix the DAG-JSON bytes with ASCII string `PUT /routing/v1 announcement:`
+    3. Sign the bytes with the private key of the Peer ID specified in the `Payload.ID`.
+       - Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
```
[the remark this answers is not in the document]
masih: No concerns other than the usual pitfalls of verifying signature of JSON object, which a reference implementation should clarify.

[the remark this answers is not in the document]
willscott: looks reasonable to me

src/routing/http-routing-v1.md
```diff
@@ -93,6 +93,56 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
...
+  {
+      "ProvideResults": [
+          { ... }
+      ]
+  }
+  ```
+
+- `ProvideResults` is a list of results in the same order as the `Providers` in the request, and the schema of each object is determined by the `Protocol` of the corresponding write object
+  - This may contain output information such as TTLs, errors, etc.
+  - It is undefined whether the server will allow partial results <!-- TODO: maybe add Error field to `announcement` schema and use it in responses ?  -->
```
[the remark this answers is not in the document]
hacdias: I think streaming would be great. From what I searched, it seems that it should be technically possible to stream the request and respond before the request is fully read. However, most places mention that that is quite uncommon and some clients and proxies might not be happy with it. So my suggestion for now would be:

1. **Clients** _SHOULD_ batch their requests. Not just one announcement per request, but also not 20 million. Batch them for an optimal value.
2. **Servers** _SHOULD_ be able to either stream back or just plain JSON (similar to `GET`s - follow `Accept` header). With this, include an `Error` field to the schema. I think that, unless the whole batch fails, we should not set a "bad" status code. I think it should be per item.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +355,81 @@ the case, the field MUST be ignored.
...
+or if server policy is to provide TTL different than the requested one.
+
+```json
+{
+  "Schema": "announcement",
+  "Payload": {
+    "TTL": 17280000
+  }
+}
+```
```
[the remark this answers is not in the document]
hacdias: @lidel I think it is fine to use the same one, and mention that when getting it as response, there may be an `Error` field too for https://github.com/ipfs/specs/pull/378/files#r1418770825.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +355,81 @@ the case, the field MUST be ignored.
...
+  - Client SHOULD sign every announcement.
+  - Servers SHOULD verify signature before accepting a record, unless running in a trusted environment.
+    - ED25519 and other small public keys MUST be inlined inside of the `ID` field with the identity multihash type.
+    - Key types that exceed 42 bytes (e.g. RSA) SHOULD NOT be inlined, the `ID` field should only include the multihash of the key. The key itself SHOULD be obtained out-of-band (e.g. by fetching the block via IPFS) and cached.
+      If support for big keys is needed in the future, this spec can be updated to allow the client to provide the key and key type out-of-band by adding optional `PublicKey` fields, and if the Peer ID is a CID, then the server can verify the public key's authenticity against the CID, and then proceed with the rest of the verification scheme.
+  - A [400 Bad Request](https://httpwg.org/specs/rfc9110.html#status.400)  response code SHOULD be returned if the `Signature` check fails.
+
+:::warn
+
+TODO: what should be the  limits? Max number of CIDs per `announcement` ?
```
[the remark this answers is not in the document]
willscott: I would say that the most realistic bound is to say that the dag-cbor-serialized form of an announcement needs to be under 2MB

src/routing/http-routing-v1.md
lidel: I've pushed 30a5e3bf3603f9c3207668254a672704d1c3c2e0 which removes the concept of singing more than one CID.

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
hacdias: @lidel what do we do with empty fields? Are they not set? Are they set as null? Are they set as empty value (depending on their value, e.g., empty array)?

src/routing/http-routing-v1.md
```diff
@@ -131,6 +181,53 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
...
+      ...
+    }
+  ]
+}
+```
+
+Each object in the `Providers` list is a *write provider record* entry.
+
+Server SHOULD accept writes represented with [Announcement Schema](#announcement-schema)
+objects with `CID` list.
```
hacdias, line 202: `with CID list`? This is for peers.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +374,80 @@ the case, the field MUST be ignored.
...
+
+- `Schema`: tells the server to interpret the JSON object as announce provider
+- `Payload`: is a DAG-JSON-compatible object with a subset of the below fields
+  - `CID` is the CID being provided.
+    - This field is not presend when used for `PUT /routing/v1/peers`
+  - `Scope` is an optional hint that provides semantic meaning about announced identifies:
+    - `block` announces only the individual block (this is the implicit default if `Scope` field is not present).
+    - `entity` announces CIDs required for enumerating entity behind the CID (e.g.: all blocks for UnixFS file or a minimum set of blocks to enumerate contents of HAMT-sharded UnixFS directory, only top level of directory tree, etc).
+    - `recursive` announces entire DAGs behind the CIDs (e.g.: entire DAG-CBOR DAG, or everything in UnixFS directory, including all files in all subdirectories).
+  - `Timetamp` is the current time, formatted as an ASCII string that follows notation from [rfc3339](https://specs.ipfs.tech/ipns/ipns-record/#ref-rfc3339).
```
hacdias, line 406:
```suggestion
  - `Timestamp` is the current time, formatted as an ASCII string that follows notation from [rfc3339](https://specs.ipfs.tech/ipns/ipns-record/#ref-rfc3339).
```

hacdias: @lidel  String or byte encoded? IPNS does bytes for validity.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +374,80 @@ the case, the field MUST be ignored.
...
+  - `TTL` is  caching and expiration hint informing the server how long to keep the record available, specified in milliseconds.
+    - If this value is unknown, the caller may skip this field, or use a value of 0. The server's default will be used.
+  - `ID` is Peer ID of the node that provides the content and also indicates the `libp2p-key` that SHOULD be used for verifying `Signature` field.
+  - `Addrs` is an a list of string-encoded multiaddrs without `/p2p/peerID` suffix.
+  - `Protocols` is a list of protocols supported by `ID` and/or `Addrs`, if known upfront.
+  - `Metadata` is a string with multibase-encoded binary metadata that should be passed as-is
```
hacdias, line 412: @lidel  String or byte encoded?

hacdias, line 412: @lidel important question: are CID, Timestamp, ID, and Metadata going to be string encoded, or DAG-JSON bytes encoded? This is very important for deserialization in the implementation. If it's just strings, it's trivial to use the native JSON decoder. Otherwise, we'll have to do a very manual IPLD lookup and parsing.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +374,80 @@ the case, the field MUST be ignored.
...
+    },
+    "Signature": "mbase64-signature"
+  }
+```
+
+- `Schema`: tells the server to interpret the JSON object as announce provider
+- `Payload`: is a DAG-JSON-compatible object with a subset of the below fields
+  - `CID` is the CID being provided.
+    - This field is not presend when used for `PUT /routing/v1/peers`
+  - `Scope` is an optional hint that provides semantic meaning about announced identifies:
```
hacdias, line 402: This field should also not be present for `PUT /routing/v1/peers`, right?

src/routing/http-routing-v1.md
```diff
@@ -131,6 +181,53 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
...
+```json
+{
+  "Providers": [
+    {
+      "Schema": "announcement",
+      ...
+    }
+  ]
+}
+```
```
hacdias, line 197: Consistent with the GET?

```suggestion
```json
{
  "Peers": [
    {
      "Schema": "announcement",
      ...
    }
  ]
}
```
```

File: src/routing/http-routing-v1.md [hunk 23]
```diff
@@ -131,6 +181,53 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
 
 Each object in the `Peers` list is a record conforming to the [Peer Schema](#peer-schema).
 
+### `PUT /routing/v1/peers`
+
+#### `PUT` Request Body
+
+```json
+{
+  "Providers": [
```
hacdias, line 190: Consistent with GET?

```suggestion
  "Peers": [
```

src/routing/http-routing-v1.md
lidel: I would go with String for every field, for readability (we already do String for CID and multiaddrs).

TODO: 
- [ ] update spec to make it clear these are all strings

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
lidel: Empty fields should not be present, just to remove ambiguity on the default value.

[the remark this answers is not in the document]
lidel: I realized going with DAG-JSON introduces inconsistency across specs.
IPNS Records do the same signature dance but normalize key-value map with DAG-CBOR.

Let's switch to DAG-CBOR, this way there is no ambiguity, and signature does not require JSON representation in non-JSON context.

```suggestion
- `Signature` is a string with multibase-encoded binary signature that provides integrity and authenticity of the `Payload` field.
  - Signature is created by following below steps:
    1. Convert `Payload` to deterministic, ordered [DAG-CBOR](https://ipld.io/specs/codecs/dag-cbor/spec/) map notation (same normalization as Data field in IPNS Records)
    2. Prefix the DAG-CBOR bytes with ASCII string `PUT /routing/v1 announcement:`
    3. Sign the bytes with the private key of the Peer ID specified in the `Payload.ID`.
       - Signing details for specific key types should follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types), unless stated otherwise.
```

src/routing/http-routing-v1.md
hacdias: I don't think it's necessary. I think the only potential footgun is the validation. What do we do with empty fields? https://github.com/ipfs/specs/pull/378#discussion_r1445850659

src/routing/http-routing-v1.md
```diff
@@ -93,6 +93,56 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
 
 Each object in the `Providers` list is a record conforming to a schema, usually the [Peer Schema](#peer-schema).
 
+### `PUT /routing/v1/providers`
```
lidel, line 96: TODO: look into HTTP semantics for `PUT` and `POST`.

Quick brainstorm: I think the way we do announcements here is closer to `POST`  (we are not replacing old ones, old ones, if they exist, will expire according to TTL set in announcement). This means we should switch to `POST` here to avoid confusing web people, who have certain expectations around `PUT` vs `POST`

achingbrain: Using POST is sensible.

For example if nodeA and nodeB ask nodeC to publish provider records on their behalf for the same CID at the same time, if they send different providers in the request, `PUT` could imply a last-update-wins style overwrite of the underlying resource when in reality both sets of provider records should be published.

lidel: Thanks! I've switched `/routing/v1/providers` and `/routing/v1/peers` to POST in 7b32f53.

src/routing/http-routing-v1.md
lidel: Thanks, will remove.

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
lidel: Added in f6f0d9b

src/routing/http-routing-v1.md
lidel: Clarified in 03475cf

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
lidel: Switched to DAG-CBOR in 05dceba

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
lidel: Ack. Maybe overthinking this IPIP, but I am trying to keep some future doors open. 

This IPIP  is an opportunity to create a signed routing record convention that could be reused outside of HTTP API. For example, Amino DHT could start accepting signed announcements from this spec in addition to routing puts where PeerID matched one from libp2p connection, allowing light clients to delegate DHT puts.

I think with switch to DAG-CBOR done in https://github.com/ipfs/specs/pull/378/commits/05dceba203cbd7da79bea536ec3e6876a52b98b8 it is generic enough to use it for both (a signed DAG-CBOR map similar to IPNS).

But for errors, it felt like mixing abstractions, so I've added explicit `error` schema in ccbc085.

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
lidel: Ok, feels like a rabbit hole, with bunch of caveats that we may want to skip for now.

Let's keep the initial version of POST limited to what already is used (https://github.com/ipni/index-provider/issues/403 uses application/json requests), we can add streaming POST in future, if needed. 


Error schema added in ccbc085

lidel:
Pushed some editorial changes based on feedback, highlights:
- /providers and /peers switched from PUT to POST
- `Signature`  generation uses DAG-CBOR instead of DAG-JSON
- /peers envelope field renamed from `Providers` to `Peers`
- 2MiB limit per announcement 

There is wip work on reference implementation in https://github.com/ipfs/boxo/pull/539

src/ipips/ipip-0378.md
```diff
@@ -0,0 +1,111 @@
...
+explicit signature. That is why the `POST` APIs proposed here introduce a new
+DAG-CBOR data format for specifying provider records, and mechanism for signing
+them.
+
+##### Reframe
+
+The Reframe API was superseded by `/routing/v1`.
+See [IPIP-337/Backwards Compatibility](https://specs.ipfs.tech/ipips/ipip-0337/#backwards-compatibility).
+
+#### Forwards Compatibility
```
lidel, line 86: TODO

As this IPIP is right now, the routing announcement signature is used only for the remote HTTP server to decide if announcement should be accepted or not. This blocks announcing with PeerID that you don't own.

However, when user asks for providers, they won't get original payload+signature back, and won't be able to verify provider records themselves. 

**We don't have to solve or require it in this IPIP,** but we should add section to "Forwards Compatibility" explaining if/how signature type introduced in this IPIP could be stored and returned to end clients when someoneone wants that functionality.

If we don't write it down, we will be having multiple competing conventions, because people who want end-to-end routing integrity will invent something.

---- 

Initial thought on this is that we already have a concept of opaque  metadata fields for opaque protocols in [Peer Schema](https://specs.ipfs.tech/routing/http-routing-v1/#peer-schema). 

If we reuse it here, this section could hint at doing something like:

```json
{
  "Schema": "peer",
  "ID": "bafz...",
  "Addrs": ["/ip4/..."],
  "Protocols": ["transport-bitswap", "signed-routing-v1"],
  "signed-routing-v1": {
    "Payload": "[mbase64-dag-cbor-blob]", 
    "Signature": "[mbase64-blob]"
  }
}
```

Any concerns  @masih @willscott @hacdias ? 
This is just for "Forwards Compatibility" section, but if we have it, we could also implement it in [someguy](https://github.com/ipfs-shipyard/someguy), allowing people end-to-end integrity when they self-host own router (e.g. in private swarms).

masih: No immediate concern on my part 👍

MarcoPolo (commented):
Took a look at this. Overall seems reasonable. I'd push for using JWS for authentication of the payload instead of the proposed DAG-CBOR strategy.

src/routing/http-routing-v1.md
```diff
@@ -277,6 +371,98 @@ the case, the field MUST be ignored.
...
+
+  - `Metadata` (optional) is a string with multibase-encoded binary metadata that should be passed as-is
+
+#### Announcement Signature
+
+- `Signature` is a string with multibase-encoded binary signature that provides integrity and authenticity of the `Payload` field.
+
+  - Signature is created by following below steps:
+    1. Convert `Payload` JSON to deterministic, ordered [DAG-CBOR](https://ipld.io/specs/codecs/dag-cbor/spec/) map notation
+       - Specification intention here is to use similar signature normalization as with DAG-CBOR `Data` field in IPNS Records, allowing for partial code and dependency reuse.
```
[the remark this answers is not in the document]
gobengo: +1 RFC9421 because it is content-type agnostic. (ActivityPub/Mastodon use this)

And by `use JWS` do you mean this JSON serialization of it? (iiuc a JSON object with `payload` and `signatures` properties)
https://datatracker.ietf.org/doc/html/rfc7515#section-7.2.1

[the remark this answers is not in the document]
MarcoPolo: Whatever serialization makes the most sense. Why +0 to the idea?

src/routing/http-routing-v1.md
```diff
@@ -93,6 +93,54 @@ The client SHOULD be able to make a request with `Accept: application/x-ndjson`
...
+      "Schema": "announcement",
+      ...
+    }
+  ]
+}
+```
+
+Each object in the `Providers` list is a *write provider record* entry.
+
+Server SHOULD accept  representing writes is [Announcement Schema](#announcement-schema).
```
SgtPooki, line 113:
```suggestion
Server SHOULD accept representing writes as [Announcement Schema](#announcement-schema).
```

src/routing/http-routing-v1.md
```diff
@@ -277,6 +371,98 @@ the case, the field MUST be ignored.
...
+    - This field is not present when used for `POST /routing/v1/peers`
+  - `Scope` (optional) is a string hint that provides semantic meaning about CID (`/routing/v1/providers` only):
+    - `block` announces only the individual block (this is the implicit default if `Scope` field is not present).
+    - `entity` announces CIDs required for enumerating entity behind the CID (e.g.: all blocks for UnixFS file or a minimum set of blocks to enumerate contents of HAMT-sharded UnixFS directory, only top level of directory tree, etc).
+    - `recursive` announces entire DAGs behind the CIDs (e.g.: entire DAG-CBOR DAG, or everything in UnixFS directory, including all files in all subdirectories).
+
+  - `Timestamp` is the current time, formatted as an ASCII string that follows notation from [rfc3339](https://specs.ipfs.tech/ipns/ipns-record/#ref-rfc3339).
+
+  - `TTL` is caching and expiration hint informing the server how long to keep the record available, specified as integer in milliseconds.
+    - If this value is unknown, the caller may skip this field or set it to 0. The server's default will be used.
```
SgtPooki, line 408: In this scenario, are we expecting clients to POST all the records they want reprovided after TTL expires?

src/routing/http-routing-v1.md
```diff
@@ -277,6 +371,98 @@ the case, the field MUST be ignored.
...
+  "Error": "error in case there was error",
+  "TTL": 17280000
+}
+```
+
+- `Error` is a string representing the error that might have happened when announcing.
+
+- `TTL` in response is the time at which the server expects itself to drop the record
+  - If less than the `TTL` in the request, then the client SHOULD repeat announcement earlier, before the announcement TTL expires and is forgotten by the routing system
+  - If greater than the `TTL` in the request, then the server client SHOULD save resources and not repeat announcement until the announcement TTL expires and is forgotten by the routing system
```
SgtPooki, line 463:
```suggestion
  - If greater than the `TTL` in the request, then the client SHOULD save resources and not repeat announcement until the announcement TTL expires and is forgotten by the routing system
```

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
gobengo: +0 meaning 'no objection' and + because I trust you have more context than I do so I'm happy to defer to what you think is best. But personally it seems like choosing RFC9421 it would provide similar functionality for JSON but also every other media type, and then maybe wouldn't need to do RFC7515 at all.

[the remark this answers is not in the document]
MarcoPolo: Gotcha. I don't have a preference on RFC 7515 or RFC 9421. But I do have a preference for using an IETF standard here instead of the proposed DAG-CBOR solution.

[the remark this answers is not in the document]
lidel: Thank you for feedback. 

Additional context is that in the future, a light client may want to use /routing/v1 POST to delegate announcements of signed records to routing systems like Amino DHT. This means we want to implement a signature type that would be easy to implement and support by Amino DHT nodes as well.

A hesitation around using something like RFC9421 is that it brings more complexity: users of IPFS stack likely already have libp2p crypto due to PeerIDs and dag-cbor due to things like IPNS.
In contrast, the RFC9421 is relatively new and requires custom implementation (not in go yet: https://github.com/golang/go/issues/41046).

It feels we don't want to rush signatures until we have time to work on delegated Amino DHT, making sure we have end-to-end, real world use case for them.

Thoughts?

My current thinking is reducing scope of this IPIP, moving signatures to separate IPIP, and shipping basic POST here without signatures specified, as a DRAFT. It will still be useful, there are use cases where delegated routing is done in a controlled environment, where signatures introduce unnecessary penalty.

Issue #377, opened 2023-02-11
Specify how API from IPIP-337 can support new protocols


lidel:
Extracted loose ends from IPIP-337 (to unblock it), but we need to address below at some point.
cc @guseggert @aschmahmann  @masih 

## Open questions

- How do we deal with plain text labels from [table.csv](https://github.com/multiformats/multicodec/blob/master/table.csv) changing in the future?
  - It happened before: [go-cid had invalid mapping](https://github.com/ipfs/go-cid/releases/tag/v0.2.0), and [we changed label for at least 4 codes](https://github.com/multiformats/multicodec/pull/283/files)
  - Can we allow numeric values instead (indicate them in hex, start with `0x`) 
- How do we allow for novel protocols that do not have multicodec registered?
  - https://github.com/ipfs/specs/pull/337/files#r1044055388
- How does the use of Schema interact with systems like IPNI?
  - https://github.com/ipfs/specs/pull/337/files#r1044060034

## Idea 1

> - `Protocol` is the multicodec name of the transfer protocol or an opaque string
>   - for protocols with multicodec, a numeric code starting with `0x` can be used instead of name
>   - for protocols without multicodec registered, codes from [private range](https://github.com/multiformats/multicodec#private-use-area) `0x300000` – `0x3FFFFF` can be used
> 

## Idea 2

-  state `Protocol` and `Schema`  are opaque strings, and write convention how to return opaque records as base-encoded bytes.
   -   there is some conceptual ovelap with https://github.com/ipni/specs/pull/4

lidel:
Worth being precise, because as of 2026 Q3, only half of this is actually extensible.

- **`Protocols` is fine.** [Routing V1](https://specs.ipfs.tech/routing/http-routing-v1/) treats transfer protocol names as opaque strings, max 63 characters, with the multicodec registry as convention rather than requirement. Label churn cannot break clients, and an application-level protocol with no registered code just picks a name. The numeric `0x` idea was never needed for this part.

- **`Addrs` is not.** Those are multiaddrs, and both stacks reject what they do not recognise, dropping the whole record rather than the one address:

  - **Go / boxo**: `types.Multiaddr.UnmarshalJSON` calls `multiaddr.NewMultiaddr`, which errors on an unknown component. The error surfaces from the record's `json.Unmarshal`, so the entire `PeerRecord` fails to parse, and `readProviderResponses` in `routing/http/contentrouter` logs a warning and continues.
[...]

Issue #376, opened 2023-02-05
IPNS: allow lean V2-only records


lidel:
## Where we are today (2023 Q1)

We are still living with the legacy of decisions made in 2021 and require creation of hybrid V1+V2 records, but we only care about V2 fields during the validation.

I've researched and documented the current state in https://github.com/ipfs/specs/pull/319, and obviously it is not the best.

Gist:
- [Created records](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-creation) have both v1 and V2 signatures, and we have duplicated values in both top level protobuf AND `data` CBOR field 
- [Record Validation](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-verification) only cares about V2 signature, but still requires fields related to V1 to be always present in a record and match values from CBOR in `data` field, for the record to be considered valid
  - This means many fields MUST be duplicated, even when both ends use a modern client that only cares about signatureV2 that guards CBOR field. 

We've been producing and expecting these hybrid V1+V2 records [since 2021](https://github.com/ipfs/js-ipns/pull/121).

## Proposed move to V2-only IPNS

It is time to clean up IPNS ecosystem and stop requiring V1 fields when there is no `signatureV1`.

For modern IPNS, the outer `IpnsEntry` protobuf should effectively only have two required fields:  `data` and its `signatureV2`, and such record, as long signature is valid, should be accepted as valid.

We can get to that future in two steps:

1. [Keep producing V1+V2 records](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-creation) as backward-compatible default, but [relax validation](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-verification)
   - Only check/require fields to be duplicated in top level protobuf IF `signatureV1` is present. IF there is no `signatureV1` and only `signatureV2` and `data` fields are present and valid, the record should be considered valid.
   - This will allow people to build V2-only systems that produce records that are considered valid.
2. (at some point in the future) stop producing V1+V2 and switch to publishing V2-only records that are protobuf with only two fields: Data CBOR+signatureV2
   - we have to do this in two steps, because we've rushed the way V2 was introduced in 2021, and STILL require field copying, even when `signatureV1` is missing :upside_down_face: and we could only care about CBOR versions from the start :(

## 
```[tasklist]
### Migration plan for V2-only IPNS
[...]

PR #379, opened 2023-02-15
IPIP-379: Delegated IPNS HTTP API

### LINKED ISSUE #343: IPIP: Delegated IPNS over HTTP
This is a follow-up from https://github.com/ipfs/specs/pull/337, see the comment here: https://github.com/ipfs/specs/pull/337/files#r1022991131

We need to be able to delegate IPNS over HTTP in a similar way as with content routing. One starting use case is for network indexers, from @masih:

> On the Network Indexer side I am working on a proposal to support IPNS via Network Indexers for example.

masih:
Propose HTTP APIs for delegated naming system that enables offloading of naming system to other processes or servers.

Relates to:

- https://github.com/ipfs/specs/pull/377

Fixes #343


lidel (changes requested):
Thank you for submitting this @masih  :pray: 

Dropped quick feedback around JSON and Signatures, and including canonical serialization to `ipns-record`  protobuf format.

I think we need to support canonical serialization because entire ecosystem expects that.

We could also have JSON as default for quick eyeballing, don't feel strongly about it nor find much utility: it feels like a distraction/security footgun but  lmk if you had some specific use cases that depend on introducing this additional JSON serialization format

routing/DELEGATED_NAMING_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+      "EOL": {
+        "Timestamp": 0,
+        "AdvisoryTTL": 0
+      }
+    },
+    "PublicKey": "<optional-public-key>",
+    "ExtendedData": {}
+  }
+}
+```
```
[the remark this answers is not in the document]
masih: `v2` means something in the iteration of protobuf format; i am not clear why it should leak here. It's a detail that's IMHO is irrelevant from the perspective of the users whom would be using this API :) WDYT?

routing/DELEGATED_NAMING_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+- `ExtendedData` represents the extensible data as arbitrary JSON object.
+
+## Versioning
+
+The path predix `/v1` donates the version number of the HTTP API. Backwards-incompatible change must
+increment the version number.
+
+## API
+
+### `GET /naming/v1/records/{ipns-name}`
```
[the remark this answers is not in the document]
masih: I thought about this at the time of writing; the potential issue I see is that `ipns-name` has `/ip(n|f)s/`prefix. That would make the URI look something like this`/routing/v1/ipns/ipns/123...` which seems slightly odd. Thoughts?

IPIP/0379-delegated-naming-http-api.md
```diff
@@ -0,0 +1,83 @@
...
+## Design rationale
+
+The rationale for delegated naming over HTTP APIs closely follows the reasoning listed in [IPIP-337](0337-delegated-routing-http-api.md#design-rationale).
+
+The document proposes the following:
+- Use of HTTP semantics for publication and resolution of naming records.
+- Preference for human-readable request and response encoding, such as JSON format
+- Optional backward compatibility support for the existing ProtocolBuffer format using `Content-Type: application/x-protobuf; messageType="IpnsEntry"`
+- Use of extra headers in `OPTIONS` response to communicate the supported cpabilities and limitations, namely:
+  - `X-Ipns-Allow-Max-Size` -- to signal maximum supported IPNS record size
```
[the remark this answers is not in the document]
masih: The size limit in the spec specifies the maximum. But in reality IPNS records are typically much smaller. The rationale in this spec is to offer an opportunity to reduce the maximum size supported with means to communicate that to the user. If no such header is returned on `OPTIONS` call then the user may assume that the max size is as dictated in the IPNS spec.

Forcing the server to offer support for maximum of 10 KiB regardless seems less than optimum in a world where we want to: 1) retire/discourage old RSA keys, 2) limit oddities caused by arbitrary size IDENTITY CIDs, and 3) facilitate large scale, efficient and fast naming delegation.

IPIP/0379-delegated-naming-http-api.md
```diff
@@ -0,0 +1,83 @@
...
+##### Serialization Format
+
+IPNS records use [ProtocolBuffer serialisation format](../ipns/IPNS.md#record-serialization-format).
+This format is widely in use in IPNS over PubSub and DHT routing systems.
+One of the motivations of this document is to introduce simple to use HTTP APIs and ultimately reduce barrier for interaction across alternative systems.
+Further, interoperability across the existing and HTTP APIs is also desirable in order to reduce the barrier for adoption of the delegated HTTP APIs.
+
+The specification here maintains backwards compatibility in terms of record serialisation, with preference for human-readable formats such as JSON.
+To maintain backwards compatibility, the implementers of the HTTP API should accept requests with following `Content-Type` header:
+- `application/x-protobuf; messageType="IpnsEntry"`
```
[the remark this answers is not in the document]
masih: I'd like to push for this please:
* I think it is fair to say that the point of specs like this is to introduce extensibility. IPNS records creation delegated via this spec will most likely be unpacked. Because the whole point is to offer alternative naming system management to the DHT. So keeping the records format untouched such that records as they are could also be put onto the existing DHT interface seems unnecessary.
* The spec proposed here offers the best of both worlds: old records are optionally accepted by the server, while the new format offers an opportunity to not only improve UX when it comes to assembling IPNS records but also improve readability and cross implementation extensibility.

routing/DELEGATED_NAMING_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+  e.g. `ipns/{ipns-key}`, `/ipns/example.com`, `/ipfs/baf...`, etc.
+- `Sequence` represents the current version of the record starting from `0`.
+- `Validity` captures the mechanism by which the record is validated. Each validity type reserves a
+  field key under this object.
+  - `EOL` donates that the validity type is EOL, containing:
+    - `Timestamp` represents the time in the future at which the record expires with nanoseconds
+       precision represented as an ASCII string that follows notation
+       from [RFC3339](https://www.ietf.org/rfc/rfc3339.txt).
+    - `AdvisoryTTL` represents an optional field that hints at how long the record should be
+       cached.
```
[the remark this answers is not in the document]
masih: > Someone will see resolved path and will use Value without checking signature.

* It is the delegated servers reponsibility to validate the records. Happy to spell that out if it's not obvious from the specification.
* The rationale is that the server should also offer enough information for the client to be able to verify records independently should the client choose to.

> We should remove these clear text fields, nudging people to use validation libraries that return a value only if the signature is valid.

This is certainly a way to do it but seems extreme at the cost of human readability, cross language extensibility. I need your help to understand why binary format would encourage people to call `.Validate` instead of decoding and using fields straight up?

routing/DELEGATED_NAMING_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+Where:
+
+- `Signature` is the multibase-encoded signature of the sha256 hash of the `Payload` field, signed
+  using the private key that corresponds to the `PublicKey` in the `Payload` if present. And
+  Otherwise, the private key associcated to the IPNS record key. Signing details for specific key
+  types should
+  follow [libp2p/peerid specs](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#key-types),
+  unless stated otherwise.
+- `Payload` is the content of the IPNS record as specified
+  by [IPNS Record](../ipns/IPNS.md#ipns-record) specification:
```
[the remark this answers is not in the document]
masih: I see this as an opportunity to reduce cognitive cycles for users that are trying to make sense if IPNS records. I am not sure I understand the necessity to carry over identical binary formatting onto specifications that are written to _delegate_ services.

> introduce computation or storage cost (either you store CBOR and generate JSON, or you store both)
> Do we need JSON for IPNS?

Dlegated IPNI servers will most likely be unpacking the records before storage because the chances are they won't be using the vanilla DHT for storage. Aside from that, the specification proposed here offers backwards compatibility with the old binary format, while allowing the plain text human readable format as well. 

Mixing both `COBR` and `JSON` in a request response mechanism that clearly supports `Content-Type` mechanism to communicate formatting seems less than optimum. As I elaborated in other comments, it seems extreme to carry on with the existing binary format regardless when native HTTP apis can support both with added benefits of better UX, human readability and cross language/platform integration (e.g. inserting IPNS records from the browser over webtransport as an example).

We could have used the binary format used in DHT for all of the delegated routing HTTP APIs. But we did not for good reason: human readability, ease of use across multiple platforms/environments, easier debugging and so on. It seems like an opportunity missed to enforce the old formats in HTTP APIs where native support for extensible formats are built in. 

Sure we can make support for old formats mandatory instead of optional but it is unclear why that should be the _only_ format supported?

IPIP/0379-delegated-naming-http-api.md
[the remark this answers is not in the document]
Winterhuman: I think decreasing the maximum size wouldn't be great, mainly for two reasons:

1. While RSA keys are being deprecated, newer quantum-resistant algorithms are appearing and they have significantly larger keys than ed25519.
2. IPNS record can have custom fields, and not just an identity CID as the `Value`, which means records can meaningfully use the extra space for metadata.

routing/DELEGATED_NAMING_HTTP.md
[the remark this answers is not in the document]
lidel: What happens, when we decide to evolve IPNS in the future, and introduce signatureV3, and have both V2 and V3  for a while, to allow clients to do graceful upgrade?
Will you break existing users by suddenly swapping the value in `signature` field?
I don't think there is a good or bad answer here, just trade-offs. 

Would be good to add a paragraph to the IPIP saying "Signature field in JSON notation will always point at the latest version of signature from IPNS Record"


After thinking about it, I don't feel strongly about the details of JSON format, as long we also require endpoint to support binary `application/vnd.ipfs.ipns-record`.  Clients who care about signature verification will always use the latter anyway, because they would recreate most of it for signature verification anyway, so it is less expensive to work on it from start, than convert JSON  protobuf/cbor.

IPIP/0379-delegated-naming-http-api.md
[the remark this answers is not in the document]
lidel: Having both sgtm, as long both have the same signature bytes.
IPNI is free to take `application/vnd.ipfs.ipns-record`, and store each field once.

routing/DELEGATED_NAMING_HTTP.md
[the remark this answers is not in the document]
lidel: I fully agree we want to minimize cognitive overhead, and  you are right, mixing JSON and CBOR  in `application/json` response will be against that. I am concerned about having deserialized, UNVERIFIED value right there in JSON payload, but if IPNI is okay with that, I am fine with it as well, as long clients can opt-in to verifiable binary format.

My _reductio ad absurdum_ failed to get the point about signatures across, so I'll try to rephrase:
All I want, is to make sure we have understanding that **we can't invent a new signature type for IPNS**.
So far, this IPIP  states that `Signature` in JSON response type invents new signature type:

> `Signature` [in JSON response from IPNI] is the multibase-encoded signature of the sha256 hash of the Payload field, signed using the private key that corresponds to the PublicKey in the Payload if present. And Otherwise, the private key associcated to the IPNS record key. Signing details for specific key types should follow libp2p/peerid specs, unless stated otherwise.

The way `signatureV2` is generated is defined in [IPNS spec](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-creation) and all clients expect this. Inventing anything new here means you force clients to now support two competing IPNS specs, and do CPU work twice by creating two different signatures, and to know when use which. 

This is harmful to the ecosystem, where we already struggle to clean up and [finish transition from V1 to V2](https://github.com/ipfs/specs/issues/376) by removing requirement for V1 fields. We want to have a single signature, and a simple format. 

The only path forward is to update description of `Signature` field in JSON representation and state it should match `IpnsEntry.signatureV2` from [IPNS spec](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-creation).

routing/DELEGATED_NAMING_HTTP.md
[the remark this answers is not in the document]
lidel: > why binary format would encourage people to call .Validate instead of decoding and using fields straight up?

Cleartext means the fastest way to make things work is to do the wrong thing – use Value from JSON response and don't care about signature validation.

Binary requires an extra step, which  means someone had to find go/js-ipns library, read its API/docs, and that increases the chance they will see `.Validate` function and at least THINK for a second why it is there.

But as I said, not feeling strongly, if IPNI feels JSON should maximize human readability, I can live with that, as long we have mandatory  `application/vns.ipfs.ipns-record` too.

> It is the delegated servers reponsibility to validate the records

My understanding is the only value delegated server validation provides is during PUTs: it protects IPNI from storing invalid records, and informs client about typos/errors.

During GET, clients can't blindly trust HTTP server returned truth to them, because they delegate verification to at least two entities (Protocol Labs AND the CA company that signed TLS cert for HTTPS). 

Using IPNS GET result from IPNI without verifying signature is as bad as not verifying CID against received block or using a random public HTTP gateway.

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Security
+
+All interaction over the APIs should use TLS to protect against third-party observation and tampering.
+Additionally, the IPNS records are signed by the publisher's identity and contain sequence number to avoid replay attacks.
+
+To avoid Denial of Service attack, maximum IPNS record size of `10 KiB` applies.
+Implements are permitted to set a lower limit. If lower than the default maximum, the limit should be discoverable via `OPTIONS` request with header key `X-Ipns-Allow-Max-Size` with value specified as the number of bytes.
+
+Similarly, a client may check if a server supports ProtocolBuffer formatted records by checking the `X-Ipns-Allow-Protobuf` header key in response to `OPTIONS` request. If present the header value must be either `true` or `false` the absence of the header indicates that ProtocolBuffer formatted records are not supported.
```
lidel, line 69: Since `application/vnd.ipfs.ipns-record` is mandatory, should we simplify spec and remove `Ipns-Allow-Protobuf`?

```suggestion
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+Privacy in delegated IPNS is out of scope for this work.
+- The usual JSON parsing rules apply. To prevent potential Denial of Service (DoS) attack, clients should ignore responses larger than 100 providers and introduce a byte size limit that is applicable to their use case.
```
lidel, line 72: I think this is copy-paste from older IPIP  – mind rephrasing it to refer to Max Size or something?

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Alternatives
+
+See:
+- [Libp2p-based IPNS](https://github.com/ipfs/go-ipns).
```
lidel, line 77: There is no libp2p there – records can be published and resolved over any medium, including sneakernet.

```suggestion
- [go-ipns](https://github.com/ipfs/go-ipns).
```

routing/DELEGATED_IPNS_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+- `Validity` captures the mechanism by which the record is validated. Each validity type reserves a
+  field key under this object.
+  - `EOL` donates that the validity type is EOL, containing:
+    - `Timestamp` represents the time in the future at which the record expires with nanoseconds
+       precision represented as an ASCII string that follows notation
+       from [RFC3339](https://www.ietf.org/rfc/rfc3339.txt).
+    - `AdvisoryTTL` represents an optional field that hints at how long the record should be
+       cached.
+- `PublicKey` represents the optional public key used to sign the record. This field is only
+  required if it cannot be extracted from the IPNS name, e.g. in the case of legacy RSA keys.
```
lidel, line 73: When a future implementer reads this, it may not be obvious why this is important, so let's clarify:

```suggestion
  required if it cannot be extracted from the IPNS name, e.g. in the case of legacy RSA keys.
  The idea here is that IPNS Record should always have all information neccessay
  for signature verification in offline mode, without additional lookups to retrieve the public key.
```

routing/DELEGATED_IPNS_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+
+**Path Parameters**
+
+- `ipns-name` the IPNS name to resolve.
+
+**Response Status Codes**
+
+- `200` (OK): indicates that the response body containing the IPNS record that corresponds to the
+  IPNS name.
+- `404` (Not Found): indicates that no matching records are found.
```
lidel, line 93: Similar here, HTTP spec states that 404s are cacheable by default, so it is in our own interest to specify some rule of thumb for how long 404 should be cached (`Cache-Control: max-age=n`)

routing/DELEGATED_IPNS_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+- `ExtendedData` represents the extensible data as arbitrary JSON object.
+
+## Versioning
+
+The path predix `/v1` donates the version number of the HTTP API. Backwards-incompatible change must
+increment the version number.
+
+## API
+
+### `GET /naming/v1/records/{ipns-name}`
```
lidel, line 83: Discussed this before, but in general squatting multiple top level paths for the same purpose of delegated routing is adding unnecessary work and complexity for devops (multiple rules in nginx, bigger surface for bugs, friction for adoption – updating software is not enough, you need to add new rules).

Let's keep all "delegated routing" APIs under `/routing/v1`

```suggestion
### `GET /routing/v1/ipns/{ipns-name}`
```

routing/DELEGATED_IPNS_HTTP.md
```diff
@@ -0,0 +1,176 @@
...
+        "AdvisoryTTL": 0
+      }
+    },
+    "PublicKey": "<optional-public-key>",
+    "ExtendedData": {}
+  }
+}
+```
+
+### `PUT /naming/v1/records/{ipns-name}`
```
lidel, line 123:
```suggestion
### `PUT /routing/v1/ipns/{ipns-name}`
```

lidel (changes requested):
@masih  apologies for slow turnaround on this one, wanted to give it a propser thought when i have more time, but it never happened :(

To move things forward, I gave it another pass.


I am ok with making JSON as human-readable as you want – we already made that call in existing `/routing/v1/providers`, but we can't compromise on signature.

My proposal is to reuse signature and [Record Verification steps from the IPNS spec](https://github.com/ipfs/specs/blob/main/ipns/IPNS.md#record-verification) and avoid inventing new things to keep the number of things IPNS implementer needs to do at minimum – details inline.

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+- Use of extra headers in `OPTIONS` response to communicate the supported capabilities and limitations, namely:
+  - `X-Ipns-Allow-Max-Size` -- to signal maximum supported IPNS record size
+  - `X-Ipns-Allow-Protobuf` -- to signal whether the server supports ProtocolBuffer formatted records.
+- Streaming interaction is not supported.
+
+### User benefit
+
+The ability of offload naming onto another process or server via an idiomatic and simple HTTP API opens up an opportunity to not only reduce the burden of naming system management, but also introduce alternative naming systems across other routing systems, namely IPNI.
+
+### Compatibility
```
lidel, line 39: TODO(before we ratify this):
- [ ] clients use `application/vnd.ipfs.ipns-record` ([IANA](https://github.com/ipfs/in-web-browsers/issues/209), [multicodec](https://github.com/multiformats/multicodec/pull/312)) no matter what router type is used for GET/PUT of IPNS record
- [ ] clients can use trustless gateway and  /routing/v1 endpoints  interchangeably, and a single service can share IPNS record cache across all supported routers DHT/Pubsub/HTTP

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+#### Backwards Compatibility
+
+##### Serialization Format
+
+IPNS records use [ProtocolBuffer serialisation format](../ipns/IPNS.md#record-serialization-format).
```
lidel, line 45:
```suggestion
IPNS records use [`application/vnd.ipfs.ipns-record` protobuf serialization format](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format).
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+This format is widely in use in IPNS over PubSub and DHT routing systems.
+One of the motivations of this document is to introduce simple to use HTTP APIs and ultimately reduce barrier for interaction across alternative systems.
+Further, interoperability across the existing and HTTP APIs is also desirable in order to reduce the barrier for adoption of the delegated HTTP APIs.
+
+The specification here maintains backwards compatibility in terms of record serialisation, with preference for human-readable formats such as JSON.
+To maximize interoperability with existing ecosystem, the canonical [IPNS record serialization format](../ipns/IPNS.md#record-serialization-format) (`0x0300`) can be requested with content type `application/vnd.ipfs.ipns-record`.
```
lidel, line 51:
```suggestion
To maximize interoperability with existing ecosystem, the canonical [IPNS record serialization format](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format) (`0x0300`) can be requested with content type `application/vnd.ipfs.ipns-record`.
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+##### Reframe
+
+See [IPIP-337/Backwards Compatibility](0337-delegated-routing-http-api.md#backwards-compatibility).
+
+#### Forwards Compatibility
+
+See [IPIP-337/Forwards Compatibility](0337-delegated-routing-http-api.md#forwards-compatibility).
```
lidel, line 59:
```suggestion
#### Forwards Compatibility

See [IPIP-337/Forwards Compatibility](https://specs.ipfs.tech/ipips/ipip-0337/#forwards-compatibility).
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+### Security
+
+All interaction over the APIs should use TLS to protect against third-party observation and tampering.
+Additionally, the IPNS records are signed by the publisher's identity and contain sequence number to avoid replay attacks.
```
lidel, line 64:
```suggestion
Additionally, the IPNS records must be [validated according to the rules stated in the IPNS spec](https://specs.ipfs.tech/ipns/ipns-record/#record-verification) before further processing.
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+
+To avoid Denial of Service attack, maximum IPNS record size of `10 KiB` applies.
```
lidel, line 66:
```suggestion
To avoid Denial of Service attack, [maximum IPNS record size defined in IPNS spec](https://specs.ipfs.tech/ipns/ipns-record/#record-size-limit) applies.
```

IPIP/0379-delegated-ipns-http-api.md
lidel, line 69: Unnecessary complexity, as explained in the other comment.
```suggestion
Implements are permitted to set a lower limit. If lower than the default maximum, the limit should be discoverable via `OPTIONS` request with header key `X-Ipns-Allow-Max-Size` with value specified as the number of bytes.

Similarly, a client may check if a server supports ProtocolBuffer formatted records by checking the `X-Ipns-Allow-Protobuf` header key in response to `OPTIONS` request. If present the header value must be either `true` or `false` the absence of the header indicates that ProtocolBuffer formatted records are not supported.
```

IPIP/0379-delegated-ipns-http-api.md
lidel, line 72: I think this is copy-paste from older IPIP  – removing.
```suggestion
- The usual JSON parsing rules apply. To prevent potential Denial of Service (DoS) attack, clients should ignore responses larger than 100 providers and introduce a byte size limit that is applicable to their use case.
```

IPIP/0379-delegated-ipns-http-api.md
```diff
@@ -0,0 +1,82 @@
...
+Similarly, a client may check if a server supports ProtocolBuffer formatted records by checking the `X-Ipns-Allow-Protobuf` header key in response to `OPTIONS` request. If present the header value must be either `true` or `false` the absence of the header indicates that ProtocolBuffer formatted records are not supported.
+
+Privacy in delegated IPNS is out of scope for this work.
+- The usual JSON parsing rules apply. To prevent potential Denial of Service (DoS) attack, clients should ignore responses larger than 100 providers and introduce a byte size limit that is applicable to their use case.
+
+### Alternatives
+
+See:
+- [Libp2p-based IPNS](https://github.com/ipfs/go-ipns).
+- Reframe; find out more on [IPIP-337 Design Rationale](0337-delegated-routing-http-api.md#design-rationale).
```
lidel, line 78:
```suggestion
- Reframe; it was deprecated and now removed. Historical reasons can be found in [IPIP-337 Design Rationale](https://specs.ipfs.tech/ipips/ipip-0337/#design-rationale).
```

lidel (commented):
@masih i'd like to pick this up and move forward. Are you ok with changes suggested above? 

Especially my concerns around signatures in JSON (https://github.com/ipfs/specs/pull/379#discussion_r1139426741) which lack useful spec or use case (who will be using JSON?). Perhaps we could remove JSON from this IPIP for now, and add it in future IPIP, if anyone has use case / need for it?

FYSA  me and @hacdias plan to support IPNS GET and PUT in Kubo if  cid.contact  intents to support it.
-  Kubo will use the original  `application/vnd.ipfs.ipns-record` for both GETs or PUTs, we will never use JSON.
   - The [ipns-record protobuf from the IPNS spec](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format) is what other already implemented IPNS routers already use (DHT, [ipns over pubsub](https://specs.ipfs.tech/ipns/ipns-pubsub-router/)), and it is true for the rest of ecosystem too:  [`?format=ipns-record` on Gateways](https://specs.ipfs.tech/ipips/ipip-0351/), we also have [multicodec](https://github.com/multiformats/multicodec/pull/312) and  [IANA registration pending](https://github.com/ipfs/in-web-browsers/issues/209) -- all use the well-defined `application/vnd.ipfs.ipns-record`.

IPIP/0337-delegated-routing-http-api.md
```diff
@@ -60,7 +60,7 @@ To understand the design rationale, it is important to consider the concrete Ref
 So this API proposal makes the following changes:
 
 - The Delegated Content Routing API is defined using HTTP semantics, and can be implemented without introducing Reframe concepts nor IPLD
-- There is a clear distinction between the RPC protocol (HTTP) and the API (Deleged Content Routing)
+- There is a clear distinction between the RPC protocol (HTTP) and the API (Delegated Content Routing)
```
lidel, line 63:
```suggestion
- There is a clear distinction between the [Kubo RPC](https://docs.ipfs.tech/reference/kubo/rpc/)  and the vendor-agnostic Routing V1 HTTP API (introduced in [IPIP-377](https://specs.ipfs.tech/ipips/ipip-0337/).
```

masih:
@lidel please go ahead we can iterate on remaining issues in separate IPIPs thanks 🙏

lidel:
Ok! I'll circle back to this when @hacdias  is back next week and we will make it limited to  `application/vnd.ipfs.ipns-record` + have reference implementation in boxo and kubo.  Having end users publish IPNS records  to both DHT and IPNI by default is our priority.

hacdias:
@lidel I rebased the PR. I would highly suggest that this new contents should be included into `http-routing-v1.md` instead of a new specification. They are the same API, so I don't see we would have different specifications.

routing/DELEGATED_NAMING_HTTP.md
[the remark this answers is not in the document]
lidel: Resolving this, as we've decided to reduce the scope of this initial IPIP and only support [application/vnd.ipfs.ipns-record](https://www.iana.org/assignments/media-types/application/vnd.ipfs.ipns-record).

lidel (approved):
Applied final editorial tweaks, this looks ready from my end.
Moving to "ready for final review" stage. 

Added to the agenda of [IPFS Implementers WG call on 2023-06-01](https://pl-strflt.notion.site/IPFS-Implementers-Working-Group-f102a74802b34529a759ffbc3ea20303?p=52fbb1d6546e420699afb07f6396a619&pm=s) but don't expect issues, since we've reduced the scope to the bare minimum:

- `/routing/v1/ipns` `GET` and `PUT` of [serialized IPNS record in existing format](https://specs.ipfs.tech/ipns/ipns-record/#record-serialization-format), no special headers, no JSON, no streaming

hacdias:
I opened a Boxo implementation PR here: https://github.com/ipfs/boxo/pull/333

lidel:
No concerns raised during final ratification during [IPFS-Implementers-Sync-2023-06-15](https://www.notion.so/pl-strflt/IPFS-Implementers-Sync-2023-06-15-b8556243125442f2b0bf48c5db90d056), ratified, merging.

PR #380, opened 2023-02-27
Return 500 on resolution errors

hacdias:
See https://github.com/ipfs/go-libipfs/issues/185


http-gateways/PATH_GATEWAY.md
```diff
@@ -679,7 +679,8 @@ Gateway MUST respond with HTTP error when it is not possible to traverse the req
 - [`404 Not Found`](#404-not-found) should be returned when the root CID is valid and traversable, but
 the DAG it represents does not include content path remainder.
   - Error response body should indicate which part of immutable content path (`/ipfs/{cid}/path/to/file`) is missing
-- [`400 Bad Request`](#400-bad-request) can be used for remaining traversal errors caused by invalid or unsupported DAG types.
```
hacdias, line 682: AFAIK, 400 is removed for path resolution purposes. There are still 400s, but for other use-cases (invalid headers, invalid format requested, etc).

lidel (approved):

PR #383, opened 2023-03-09
IPIP-0383: Compact Denylist Format

hsanjuan:
This IPIP introduces a line-based denylist format for content blocking on IPFS
focused on simplicity, scalability and ease-of-use.

Related:
- https://github.com/hsanjuan/nopfs – an implementation of this IPIP which add supports for content blocking to the go-ipfs stack and particularly to Kubo.


hsanjuan:
Just checkpointing, I need to finish it tomorrow probably.

BigLep:
@hsanjuan :  thanks for creating.  Can you please also make sure there is visibility to this in https://github.com/ipfs/specs/pull/340 for visibility?

hsanjuan:
Note to self: There's a rule type I forgot to ask in the form of : `*/path/to/content`, which blocks paths regardless of mhash.

hsanjuan:
Note to self: double-hashes could be multibase encoded. Thinking if they could actually be a multibase encoded multihash, or directly a raw CID. Current list format would have to be a special case.

This removes the need for Hints indicating base/hash function.

hsanjuan:
Note to self: remove "allow block items", as desired effect can be mostly achieved with negated rules, and it is more efficient.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,338 @@
...
+- The hash of an IPNS path `/ipns/*`.
+- The hash of a CIDv0[+path]
+
+Blocking layer recommendation: Resolver + BlockService.
+
+In order to check for a matching rule, the Resolver should:
+
+- IPFS path: convert the CID to v1base32 and hash the path without the `/ipfs/` prefix.
+- IPFS path: convert the CID to v0 and hash the path without the `/ipfs/` prefix.
+- IPNS path: hash the path "as is".
```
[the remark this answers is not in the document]
hsanjuan: yes, this is correct.

hsanjuan:
I have done one more pass to the spec, explaining a bit better complex things like double-hashing and making a distinction between NameSystem and PathResolver. I think I'm pretty happy with the current form.

hsanjuan:
Test fixtures are missing... A list of tests will come next.

An implementation of a content blocker with support to this format exists now in https://github.com/hsanjuan/nopfs, and the Badbits list is also available in compact form: https://badbits.dwebops.pub/badbits.deny.

IPIP/383-compact-denylist-format.md
```diff
@@ -0,0 +1,112 @@
...
+* It can be understood by just looking at it.
+* It can be edited by hand.
+* Implementations can choose to support different aspects (i.e. blocking but no optional hints).
+* Denylist parsers are easy and stupid.
+
+### Compatibility
+
+The current Protocol Labs denylist format
+https://badbits.dwebops.pub/denylist.json can be easily converted into the
+proposed compact format. This is shown at https://badbits.dwebops.pub/denylist.json.
```
olizilla, line 101:
```suggestion
proposed compact format. This is shown at https://badbits.dwebops.pub/badbits.deny.
```

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,458 @@
...
+
+Our recommendation is that /mime/ rules automatically set IPFS clients into a
+"unixfs only" mode where only unixfs (+raw blocks) are allowed at the
+BlockService layer, and content type is checked at the Unixfs layer, as the
+blocks get assembled into an actual files. That should cover gateway usage.
+
+#### Allow (or negated) rules
+
+Block items can be prepended by `+`, that items matching the rule are to be allowed
+rather than blocked.
```
olizilla, line 385: Maybe `!` for negation, like a .gitignore

https://git-scm.com/docs/gitignore#_pattern_format

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,458 @@
...
+When users want to block by multihash directly, they must base58btc-encoded
+multihashes. This rule does not block subpaths that start at this CID, only
+the CID itself.
+
+Blocking layer recommendation: BlockService.
+
+##### `/ipfs/CID/PATH`
+
+IPFS-Path-Rule: Blocks the exact ipfs path that is referenced from the
+multihash embedded the CID before attempting to resolve it. It does not block
```
olizilla, line 256:
```suggestion
multihash embedded in the CID before attempting to resolve it. It does not block
```

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,458 @@
...
+    - The b58-encoded-multihash extracted from an IPNS key when the IPNS key
+      is a CID.
+  - An IPFS-Path: `b58-encoded-multihash/P/A/T/H` where the multihash is
+    extracted from the CID in `/ipfs/CID/P/A/T/H` (The multihash and the CID
+    are the same in the case of CIDV0). The `/P/A/T/H` component is optional
+    and should not have a trailing `/`.
+
+The latter form allows blocking by double-hash using any hashing function of
+choice. Implementations will have to hash requests using all the hashing functions
+used in the denylist, so we recommend sticking to one.
```
olizilla, line 332: Do we need this yet? Can we specify use of a multihash for self-describing future-proof goodness but require that only sha256 is valid for now to simplify things? For context: I'm looking at how i would add support for this spec to e-ipfs and w3s.link.

I acknowledge that we already have to accept providers can only do best effort enforcement of these lists given we already have to deal with path based rules which can't enforced at the BlockService level e.g in e-ipfs which is bitswap only... so i could just ignore entries that are hashed with a fn i don't support yet, but i'd prefer to be able to say "we support all block items that can be applied by the blockservice".

hsanjuan: If your denylist only includes one type of hashing function, then you only hash using that one to check the double-hashes. So it is up to who makes the denylist... I would totally discourage using multiple hashing functions for items in the same denylist.

But other than that, since they are multihashes there's no reason to support anything. In a way I'm deferring the discussion of "what is the best hashing function for this" until later, because sha256 is probably not it. Mostly you just want something very very fast, and doesn't necessarily need to be a cryptographic one as it just meant to obfuscate.

COMPACT_DENYLIST_FORMAT.md
bumblefudge: In all the examples, negations follow supersets of the negated content without a blank line in between.  this is probably an unintended coincidence, but given that `\n-` and `\n+` are used in some syntaxes like markdown to indent list items or derivative/dependent items, this choice of negation syntax *could* confuse some readers who interpret/assume a rule that you can't negate something unless its a prior entry or subset thereof.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,458 @@
...
+# Block some subpaths (equivalent rules)
+/ipfs/Qmah2YDTfrox4watLCr3YgKyBwvjq8FJZEFdWY6WtJ3Xt2/test*
+/ipfs/QmTuvSQbEDR3sarFAN9kAeXBpiBCyYYNxdxciazBba11eC/test/*
+
+# Block some subpaths with exceptions
+/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/blocked*
++/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/blockednot
++/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/blocked/not
++/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/blocked/exceptions*
+
```
bumblefudge, line 151:
```suggestion

# Last-wins resolution of conflicts: 

/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/

!/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/allowedsubset

/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/allowedsubset/disallowedsubset

```
(note, I used `!` instead of `+` as negation prefix!)

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,458 @@
...
+detail.
+
+In particular, a [Blocker implementation validator](https://github.com/ipfs-shipyard/nopfs/tree/master/tester) is provided in Go, and can be adapted to other languages if needed.
+
+### Security
+
+This proposal takes into account security:
+
+- Denylist headers and line-length limits are well specified to avoid malformed lists to cause things like large memory usage while parsing.
+- Supported type of blocks have been though out to avoid amplified consumption of resources or side effects (i.e. downloading of additional dag-blocks) during the implementation.
```
bumblefudge, line 433:
```suggestion
- Supported type of blocks have been thought out to avoid amplified consumption of resources or side effects (i.e. downloading of additional dag-blocks) during the implementation.
```

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I'm fine changing the negation operator to `!`, (or supporting them all), but then it should be changed across the doc and not in one place.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I'm not sure this is correct, the last line should probably be `/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/*` as otherwise it won't match `/ipfs/QmUboz9UsQBDeS6Tug1U8jgoFkgYxyYood9NDyVURAY9pK/allowedsubpath` and the path will be allowed (if I'm not mistaken).

I added:

```
# Block some subpaths with exceptions: last-matching-rule wins (!)
```

I think it is illustrative for a showcase of the syntax.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I have switched to "!"  now.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+
+# Block IPNS domain name and path
+/ipns/domain2.example/path
+
+# Block IPNS key - blocks wrapped multihash.
+/ipns/k51qzi5uqu5dhmzyv3zac033i7rl9hkgczxyl81lwoukda2htteop7d3x0y1mf
+
+# Block all mime types with exceptions
+/mime/image/*
+!/mime/image/gif
```
darobin, line 163: Is this something that a denylist format needs to support? It feels like it belongs in a local configuration format rather than with something that will get distributed. What's more, media type information is unreliable and rarely available on IPFS as it tends to rely on sniffing.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+```
+
+#### Header
+
+The list header is a YAML block:
+
+- Must be valid YAML
+- Fully optional
+- 1KB maximum size
+- Delimited by a line containing `---` at the end (document separator)
```
darobin, line 211: If the header is optional, is the document separator required or are implementations expected to detect that a line beginning with a block item to know that there is no header?

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+<block_item2> [hint_list]
+...
+```
+
+#### Header
+
+The list header is a YAML block:
+
+- Must be valid YAML
+- Fully optional
```
darobin, line 209: If it's optional but has mandatory syntax, what is the expected behaviour if the syntax is incorrect? Does the implementation ignore the entire list or just the header?

COMPACT_DENYLIST_FORMAT.md
bumblefudge: right, I'd break this into two questions that should ideally be decided before pouring cement:
1.) is the system layered enough that local configs and/or processing of lists after merging is a better place for this to economize at higher distro/merging layer, and
2.) is this a required capability or an optional/extension feature (to allow economizing compute at any layer if implementor needs to avoid sniffing anything)

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+- 1KB maximum size
```
darobin, line 210: This seems very short. A description alone could run over. What is the expected processing behaviour in case the size is exceeded?

COMPACT_DENYLIST_FORMAT.md
bumblefudge: as above, the tradeoffs are expressiveness of list-sharing for compute-obligation for conforming processors of the lists.  here too deciding layering before committing to a degree of expressiveness (or the core/extension, required/optional corrolary) might be preferable to deciding this in advance of commitments to implement and conform at scale.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+- Delimited by a line containing `---` at the end (document separator)
+
+Known-fields (they must be lowercase):
```
darobin, line 213: Is the intent that these are the only fields that are accepted or are implementations expected to ignore unknown fields? If they ignore unknown fields, is there a way of reserving future standardised additions? If they don't ignore unknown fields, what is the expected processing?

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+
+- `version`: the denylist format version. Defaults to 1 when not specified.
```
darobin, line 215: There doesn't seem to be specific processing happening based on this? Are implementations expected to reject versions they aren't aware of?

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+
+Lines should not be longer than 2MiB including the "\n" delimiter.
+
+#### Block item
+
+A block item represents a rule to enable content-blocking:
+
+- `PATH` elements are expected to be %-encoded, per [RFC 3986, section 2.1](https://developer.mozilla.org/en-US/docs/Glossary/percent-encoding).
+- `CID` elements represent a CID (either V0 or V1).
+- `CIDv0` are for us equivalent to baseb58btc-encoded sha256 multihashes although they are not the same thing (a CIDV0 carries implicit codec (dag-pb) and multibase information (b58btc). When we say a b58-encoded multihash needs to be extracted from the CID, this usually is a no-op in case of CIDv0s.
```
darobin, line 240: Is there specific processing that implementations are expected to carry out for items they don't recognise? Skipping is more resilient to format evolution in a distributed system but I would worry that operators might think they're properly protected by tracking a given list with legal implications when in fact their deployment is skipping over some items it doesn't get.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+A *hint* is a key-value duple associated to the denylist as a whole (part of the header), or to a specific \<block_item\>.
+
+Header hints can be used to set denylist-wide options or information that
+implementations can choose to interpret or not.
+
+#### List body
+
+A denylist is made of lines which are made by a *block items* followed by zero or more space-separated hints.
+
+Lines should not be longer than 2MiB including the "\n" delimiter.
```
darobin, line 232: Is the expectation that the implementation will bail on longer lists? Or ignore after a cut-off?

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Mime-blocking will likely have to rely on sniffing and likely on stuff that can be interpreted as a file (so unixfs). The concern is warranted. It would also mostly apply to gateways, as it is the main thing turning blocks into files again.

I would be fine striking this off. If someone wants a text-only gateway then it is probably better done with nginx.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: How I do it is to read a max of 1KB looking for `---`. If not there, assume no header and start parsing line by line.

The header is fully optional but if it exists it needs to have the ---.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: We can increase the max size to whatever is sensible.

If the denylist does not comply with the spec, then you can consider it invalid and stop processing it as a general rule.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: It has known fields that you can expect to find on a header. Missing fields should just be ignored I think.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Implementations can expect to find these fields. They may or not be there. There might be other fields too and implementations can do whatever they want with those.

With regards to reserving fields, we could add a rule that any custom fields should start with `x-*` and that implementations should ignore unknown fields otherwise. I don't know if it matters a lot.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Yes, as it would mean they don't know how to support them, I guess. i.e. `version: 2` is not something that is specified, yet.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Implementation-specific? I'm not sure the specification should say how implementations react to things outside the spec. You could kill everything, or you could log an error message, or you could log a warn message. Perhaps we can recommend logging an error.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: It's a limit per line. There is an issue here: you can always have valid IPLD paths that are longer than whatever the limit it, so those would be unblockable. I'm not sure what to make of that. On the other side, having a 50GB line might be a way to exhaust memory. 2MiB seemed like a compromise, but could be higher and be fine too I guess.

COMPACT_DENYLIST_FORMAT.md
darobin: The reason I'm asking all these annoying questions is because I think that this is a spec for which interoperability is vital. If it were just a convenient way of sharing some configuration or something like an adblock list where the worst-case failure is that you see an ad you don't want to see, then the details of implementation behaviour wouldn't matter much. But the failure impact here is that people could end up going to jail, being fined, or having their business blocked, put on the SafeBrowsing list, etc.. As an operator, if I subscribe to a CSAM or DMCA or phishing blocklist, I need to _know_ that it works.

With 10^2 implementations running over 10^6-10^7 nodes (and more!), knowing that it works means we have to be painfully finicky about processing rules. If your implementation reads 1000 bytes before giving up looking for `---` and mine reads 1024, we might get different lists. Or mine assumes UTF-8 (there's no specification of encoding for the format) but you're Japanese and so used UTF-16, we might get wildly different results, or might be counting characters and bytes differently, etc.

I think you specified 1kb here because that's how far you scan for `---`? I think that that's likely to cause problems over time (more metadata gets added leading to needing more space, but can it interoperate with deployed code, version indicators work poorly in distributed environments, etc.). Might an approach like Markdown frontmatter (having a `---` both before and after) not work better?

More generally, would you be okay if I PRed your PR to make it tighter from an interop perspective? It'll be annoying, and it'll make it longer and less pleasant to read, but I'm afraid that we need it for this context.

hsanjuan: > I need to _know_ that it works.

The spec cannot ensure this. The implementation may end up failing even if spec is 100% tight on details.



I agree it is good to be more accurate though. What is a reasonable header size?

UTF-8 encoding should also be specified.



> Might an approach like Markdown frontmatter (having a `---` both before and after) not work better?

I don't quite see the point as it doesn't make anything easier with regards to parsing. If `---` is found at the beginning, then you can assume that here is a header but then have to deal with parsing that header. There might still not be a header at all if the denylist is malformed, so you will still have to handle that. You still need to scan until the `---` at the end is found or hit the limit etc.

It is simpler imho the way it is.

It is more interesting to discuss about failure modes: should a malformed entry abort the full list? be ignored? what happens if a header is not found within the limits but the first entry is not a parsable blocking rule (i.e. could signal that the header is too long).

For the the implementation should attempt to block what it can understand and skip over what it cannot understand (logging errors as needed). We should not be blocking or breaking the application due to malformed lists.

COMPACT_DENYLIST_FORMAT.md
bumblefudge: Wait, does this mean that the *block items*  (referents of the various kinds of entries) need to be under 2MiB, or that lines with 1.9MiB of annotations are valid?

hsanjuan: The line of text should be less than 2MiB. How that is divided between block_item and annotations is not specified. IMHO, it is a bit easier for a parser implementation to enforce limits on the basic unit of a line, rather than having separate limits for the block_items or allocations, even if that would perhaps be more "correct".

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Added a note

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Added a note that implementations should not parse unsupported versions.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I added a note about custom fields starting with `x-`. Custom information can also be added in the hints map. I clarified that header hints are like writing hints for each and every item in the list.

An implementation may return an error when an item is blocked, and ideally, the rule that caused the blocking, and that includes the hints for that rule, both the specific and the header hints.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I have clarified/added details on how to parse the header. I have increased the max size to 1MiB and specified exactly the number of bytes. I have made explicit that the list is UTF-8 encoded and newlines are `\n` characters.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I added a note saying that failure to parse the header as YAML when it is present (`---` is found) should abort and signal error. A header too-large gets processed as rules, and will error depending if the implementation errors on bad rules or ignored if the implementation ignores bad rules.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Removed mime blocking from spec.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I am fine prominently recommending a hashing function. Which one should it be though?

What is:

* A cryptographic hash function (we need this in the end as pre-image and collision protection are important).
* Well supported in all languages/platforms
* Very fast

Is it blake3? sha256? Are we ok with md5 (sure there are collisions but need to be carefully crafted). ? I'll ask around.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,484 @@
...
+
+Doublehash-Rule: Blocks using double-hashed item, which can be:
+
+- The sha256-hex-encoded hash of `CIDV1_BASE32/PATH`: this is the legacy
+  badbits block anchor format. It can only block by CID and not by
+  multihash. When no path present, the trailing slash must be kept
+  (`CIDV1_BASE32/`).
+- A b58-encoded multihash (a.k.a CIDV0), corresponding to the Sum() of:
+  - An IPNS-Path:
+    - `/ipns/IPNS` when the IPNS name is NOT a CID.
```
lidel, line 356:
```suggestion
    - `/ipns/name` when the IPNS `name` is NOT a CID.
```

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,484 @@
...
+    - The b58-encoded-multihash extracted from an IPNS key when the IPNS key
+      is a CID.
```
lidel, line 358:
```suggestion
    - The b58-encoded-multihash extracted from an IPNS name when the IPNS name
      is a CIDv1 with `libp2p-key` codec.
```

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,484 @@
...
+
+- IPFS path: convert the CID to v1base32 and hash `CIDV1BASE32/PATH` with the
+  hashing functions used in the denylist. Match against declared double-hashes.
+- IPFS path: convert the CID to CIDv0 and hash `CIDV0/PATH` without trailing `/` with the hashing functions used in the denylist. Match against declared double-hashes.
+- IPNS path:
+
+The NameSystem should:
+
+- If NAME is a domain name: Hash `/ipns/NAME` with the hashing functions used in the denylist. Match against declared double-hashes.
+- If NAME is a CID, extract the multihash, encoded with baseb58btc and hash it with the hashing functions used in the denylist. Match against declared double-hashes.
```
lidel, line 401:
```suggestion
- If NAME is a CID, extract the multihash, encode it with baseb58btc and hash it with the hashing functions used in the denylist. Match against declared double-hashes.
```

COMPACT_DENYLIST_FORMAT.md
olizilla: Would you consider defining it as _only_ sha2-256 for the first version of the spec? That restriction could be loosened later if folks need it.

hsanjuan: I would like more input. I think sha2-256 may impose larger perf impact than necessary.

lidel: Here is an idea/compromise: we explicitly define the hash functions that should be used in the YAML header under `double-hash` key. If the key is not present, we assume SHA256 and clients only use that (backward-compatible). 

This way, we future-proof the format to switch to better functions as the time goes by, but we don't force everyone to waste resources on hashing with every function if the list only uses one.

Example: if we want to limit users to Blake3 `double-hash: 0x1e` or  allow SHA256 and Blake3: `double-hash: 0x1e, 0x12`

hsanjuan: Clients don't need to hash with every function. They only hash with the functions that the list uses, and they know which ones are those because of the multihash-prefix, so as you parse the list you know which ones to double-hash with (if any).

The future-proofness is ensured by using multihashes already.

The question is whether we require those multihashes to be sha256 only or something else.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,457 @@
...
+```
+
+Blocking layer recommendation: NameSystem + PathResolver + BlockService.
+
+In order to check for a matching rule, the PathResolver should:
+
+- IPFS path: convert the CID to v1base32 and hash `CIDV1BASE32/PATH` with the
+  hashing functions used in the denylist. Match against declared double-hashes.
+- IPFS path: convert the CID to CIDv0 and hash `CIDV0/PATH` without trailing `/` with the hashing functions used in the denylist. Match against declared double-hashes.
+- IPNS path:
```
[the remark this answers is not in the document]
hsanjuan: No, I think it is just necessary to delete that line. The PathResolver never receives an unresolved IPNS path. When I wrote this I didn't know and must have left it like this. I will fix it.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: Fixing this by hand.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: I don't understand the CIDv1 addition. IPNS names can be CIDv0, in which case there is nothing to do, and CIDv1s? Or should I add CIDv0 as a different line?

I didn't care to research if you can have IPNS names using non b58 encoding for CIDv0 or non libp2p-key for CIDv1...

hsanjuan: In practice, you try to parse NAME as a CID and if that flies you get the b58-multihash out of it and double hash that.

If not, you double hash `/ipns/name`.

lidel: This is just a precaution  – better to make spec very explicit and avoid the problems around the fact that PeerID != CIDv0 (CIDv0 only starts with `Q` while PeerID can also start with `1`, see https://cid.ipfs.tech/#12D3KooWHj48PbuAPUf9LeovBJZEXn2A4rseW5z9wbTBwvM82nAq and https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#decoding)

COMPACT_DENYLIST_FORMAT.md
lidel: For pragmatic reasons, our options are probably limited to either sha2-256 or blake3 (ipfs impls. will have both).
Between sha2 and blake3, the latter should make a faster default (faster than md5 too). Maybe benchmark?

ps. The purpose behind the explicit list of hashes in the YAML header was to protect the node operator from silently exploding the list of hashes to check when the intention is to only use one (or two) functions.  But maybe it is not a real DoS risk, these lists will come from somehow trusted sources.
If we decide that we only use a single hash function per list, it still feels like something that should be explicitly defined in YAML header, and not hard-coded in the spec (future-proofing without having to change the specs).

COMPACT_DENYLIST_FORMAT.md
lidel: @hsanjuan should we remove it from excamples here too?

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,486 @@
...
+# Blocking by CID - blocks wrapped multihash.
+# Does not block subpaths.
+/ipfs/bafybeihvvulpp4evxj7x7armbqcyg6uezzuig6jp3lktpbovlqfkuqeuoq
+
+# Block all subpaths
+/ipfs/QmdWFA9FL52hx3j9EJZPQP1ZUH8Ygi5tLCX2cRDs6knSf8/*
+
+# Block some subpaths (equivalent rules)
+/ipfs/Qmah2YDTfrox4watLCr3YgKyBwvjq8FJZEFdWY6WtJ3Xt2/test*
+/ipfs/QmTuvSQbEDR3sarFAN9kAeXBpiBCyYYNxdxciazBba11eC/test/*
```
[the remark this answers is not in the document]
hsanjuan: Say you are running a gateway only, perhaps the only place you want to block is at the namesys and resolver levels.

So if you put `/ipfs/cid`  in the denylist, it would still be fine as long as you  are not blocking in the BlockService-level.

Our gateways/kubo need more comprehensive blocking (as, I don't want to "fetch" a blocked CID in the first place), but I can imagine applications which are ok only blocking by path and not by block, and allow to fetch blocks to resolve paths, even if paths to those specific blocks are blocked.

Also note that if "/ipfs/cid" is blocked, we see this issue with subpaths. But if we are talking of "/ipfs/cid/path" and "/ipfs/cid/path/other", the latter might still work, depending on how resolution is implemented. So `/*`  has a point there.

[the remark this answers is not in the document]
lidel: Thanks, this makes sense, I think the problem here is that this is useless example of the path blocking:

```
# Block all subpaths
/ipfs/QmdWFA9FL52hx3j9EJZPQP1ZUH8Ygi5tLCX2cRDs6knSf8/*
```


We have better example below, with actual subpath:

```
# Block some subpaths (equivalent rules)
/ipfs/Qmah2YDTfrox4watLCr3YgKyBwvjq8FJZEFdWY6WtJ3Xt2/test*
```

I think we should just remove the `/ipfs/cid/*` example as it is only mudding the water.
`/ipfs/cid/subdpath/*` is way better.

COMPACT_DENYLIST_FORMAT.md
```diff
@@ -0,0 +1,486 @@
...
+# and QmVTF1yEejXd9iMgoRTFDxBv7HAz9kuZcQNBzHrceuK9HR etc. by multihash
+//QmX9dhRcQcKUw3Ws8485T5a9dtjrSCQaUAHnG4iK9i4ceM
+
+# Double hash Path block using blake3 hashing
+# base58btc-blake3-multihash(gW7Nhu4HrfDtphEivm3Z9NNE7gpdh5Tga8g6JNZc1S8E47/path)
+# Blocks /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
+# /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
+# /ipfs/f01701e20903cf61d46521b05f926ba1634628d0bba8a7ffb5b6d5a3ca310682ca63b5ef0/path etc...
+# But not /path2
+//QmbK7LDv5NNBvYQzNfm2eED17SNLt1yNMapcUhSuNLgkqz
```
lidel, line 187: @hsanjuan `QmbK7LDv5NNBvYQzNfm2eED17SNLt1yNMapcUhSuNLgkqz` uses sha256, not blake3, blake3 is only used in the root CID of th epath that is being blocked (`bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a`) is this intended?

Would it make sense to switch this to multihash which is also blake3, to demonstrate one can use something other than sha256?


below generated with `printf "gW7Nhu4HrfDtphEivm3Z9NNE7gpdh5Tga8g6JNZc1S8E47/path" | ipfs add --raw-leaves --only-hash --hash blake3 --quiet | ipfs cid format -f '%M' -b base58btc`: 

```suggestion
# Double hash Path block using blake3 hashing
# base58btc-blake3-multihash(gW7Nhu4HrfDtphEivm3Z9NNE7gpdh5Tga8g6JNZc1S8E47/path)
# Blocks /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
# /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
# /ipfs/f01701e20903cf61d46521b05f926ba1634628d0bba8a7ffb5b6d5a3ca310682ca63b5ef0/path etc...
# But not /path2
//gW813G35CnLsy7gRYYHuf63hrz71U1xoLFDVeV7actx6oX
```

COMPACT_DENYLIST_FORMAT.md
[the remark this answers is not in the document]
hsanjuan: I have removed the example and added NOTEs explaning and warning about the behaviour here.

COMPACT_DENYLIST_FORMAT.md
hsanjuan: ok...  The spec equates base58btc with CIDv0 but blake3 multihash should work too.

This is wrong in the implementation. It assumes the double-hash is a CIDv0.

src/ipips/ipip-0383.md
```diff
@@ -0,0 +1,123 @@
...
+  - Lists support CIDs
+  - Lists support CIDs+path (explicit)
+  - Lists support CIDs+path (implicit - everything referenced from CID)
+  - Lists support double-hashed multi-hashes
+  - Lists support double-hashed cid+path (current badbits format)
+  - Lists can be edited by hand on a text editor
+  - Lists are ipfs-replication-friendly (adding a new entry does not require downloading more than 1 IPFS block, to sync the list).
+  - Lists support comments
+  - Lists support gateway http error hints (i.e. type of block)
+  - `echo "/ipfs/cid" >> ~/.config/ipfs/denylists/custom` should work
```
lidel, line 79: :information_source: there is a bug in [spec-generator](https://www.npmjs.com/package/spec-generator) tool that renders `>>` incorrectly – ignore for now.

src/ipips/ipip-0383.md
```diff
@@ -0,0 +1,123 @@
...
+IPFS implementations should support content moderation, particularly when it
+comes to deployments of publicly-accessible infrastructure like gateways.
+
+The first step in a larger strategy to enable decentralized content moderation
+in IPFS setups is to agree in a denylist format that different implementations
+can rely on and share.
+
+## Detailed design
+
+See :cite[compact-denylist-format].
```
lidel, line 38: :information_source:  This reference will start working once we have the spec published at https://specs.ipfs.tech/compact-denylist-format

src/compact-denylist-format.md
```diff
@@ -0,0 +1,504 @@
...
+  of the block item is, and therefore avoids making the denylist a directory
+  of *bad* content. This is particularly useful for harmful content, where
+  solely accessing it is bad.
+- Double-hashing does not exclude adding additional context via comments or hints
+- The presence of a single double-hashed block item makes necessary that the implementation hashes every CID and CID+path that needs to be checked, which has a performance impact.
+- In general, it is good that users can inspect the nature of the content blocked if they wish to, so we recommend not using double-hashing by default as it helps transparency (i.e. blocking due to copyright claims).
+
+## Implementations
+
+- [NoPFS](https://github.com/ipfs-shipyard/nopfs): An implementation of IPIP-383 which add supports for content blocking to the go-ipfs stack and particularly to Kubo.
```
lidel, line 494: :thought_balloon:  TODO: update once/if we know this ships built-on  in Kubo 0.24

src/compact-denylist-format.md
```diff
@@ -0,0 +1,504 @@
...
+
+The latter form allows blocking by double-hash using any hashing function of
+choice. Implementations will have to hash requests using all the hashing functions
+used in the denylist, so we recommend sticking to one.
+
+Conveniently, the latter form allows using a b58-encoded sha256 multihashes
+(usual form of CIDv0 - `Qmxxx...`), so that double-hashes can be like:
+
+```
+$ printf "QmecDgNqCRirkc3Cjz9eoRBNwXGckJ9WvTdmY16HP88768/my/path" | ipfs add --raw-leaves --only-hash --quiet | ipfs cid format -f '%M' -b base58btc
```
lidel, line 375: :thought_balloon: TODO(editorial):  make "Example" section with `:::note`

COMPACT_DENYLIST_FORMAT.md
lidel: Assuming yes. Removed remaining mentions in 7233af2
We can always add them in future IPIP.

src/compact-denylist-format.md
```diff
@@ -0,0 +1,498 @@
...
+# and QmVTF1yEejXd9iMgoRTFDxBv7HAz9kuZcQNBzHrceuK9HR etc. by multihash
+//QmX9dhRcQcKUw3Ws8485T5a9dtjrSCQaUAHnG4iK9i4ceM
+
+# Double-hash Path block using blake3 hashing
+# base58btc-blake3-multihash(gW7Nhu4HrfDtphEivm3Z9NNE7gpdh5Tga8g6JNZc1S8E47/path)
+# Blocks /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
+# /ipfs/bafyb4ieqht3b2rssdmc7sjv2cy2gfdilxkfh7623nvndziyqnawkmo266a/path
+# /ipfs/f01701e20903cf61d46521b05f926ba1634628d0bba8a7ffb5b6d5a3ca310682ca63b5ef0/path etc...
+# But not /path2
+//QmbK7LDv5NNBvYQzNfm2eED17SNLt1yNMapcUhSuNLgkqz
```
lidel, line 160: TODO: we need to update this example after https://github.com/ipfs-shipyard/nopfs/pull/28/files lands, and have double-hash entry that is Multihash that is something other than sha256 (e.g. `gW813G35CnLsy7gRYYHuf63hrz71U1xoLFDVeV7actx6oX` from the linked PR)

lidel (commented):
Rebased and applied editorial fixes to publish both spec and IPIP on https://specs.ipfs.tech/ website.
I'll be revisiting this as we iron remaining kinks in reference implementation https://github.com/ipfs/kubo/pull/10161 (+ related PRs), but the content is ready for final reviews from the wider audience of node operators and implementers.
I'll mention this during next IPFS Implementers Call.

src/ipips/ipip-0383.md
lidel: workaround fix in https://github.com/ipfs/spec-generator/pull/21, will update this PR to get that

COMPACT_DENYLIST_FORMAT.md
olizilla: yes, i  strongly agree that the header yaml should define the hash fn, and i am asking that the spec, in it's initial wording, require only sha-256 be used.

we can even add a note in the spec that folks should raise a PR here if that is causing them real perf problems.

adding support for more hash functions would have to take into account that there is no baseline requirement of what hash functions an ipfs impl must support. web3-storage does not verify blake3 today. surprising but true https://github.com/web3-storage/car-block-validator/blob/cb121b780c86724c0266f797b5a4badffad77140/src/index.js#L1-L17

hsanjuan: You don't need to list hash functions in the header. As you parse the list, you see the multihash function used in the multihashes that you encounter and that is the list of hashing functions you have double-hash with. Putting the list at the top is redundant.

I think the issue here is a matter of compatibility, and sha256 being the best thing that is nicely and universally supported everywhere. I'm ok with the spec requiring sha256 for double-hashing, otherwise leave things open as they are. If people use other hashes (which remains possible), they would be breaking the spec, but that's up to them.

bumblefudge: I actually disagree that listing which hash functions are used in a list is necessary and always _redundant_, because I am focused on a use-case where more lists from more unfamiliar authorities get added and subtracted more often, making the list itself worth annotating.  If managing and federating many denylists is being optimized for, all kinds of (optional!) frontmatter fields could be quite helpful in deciding _which ones_ to federate/follow before parsing them, particularly for software that builds on CIDs selectively (which might need to know in advance if a denylist includes entries it cannot, or does not need to, parse).

In any case, I would argue that an optional frontmatter field for declaring which hash functions were/are used to populate a list is good, actually.  I don't have strong opinions, but some basic questions, in case enough people agree that a section is worth adding to the IPIP itself:
1. should the values be `name` entries from the [`multicodec` table](https://github.com/multiformats/multicodec/blob/master/table.csv), or the actual tags?
2. what should a consumer do/think/feel if an entry uses a tag not announced up top?
3. should list hosters ADD entries to the frontmatter declaration over time if they add functions over time, or create a new list at a new URL when these "env var"s change? Considering many subscribers have to unload and reload the list for any changes except EOF appends, this could complicate the spec a lot.

hsanjuan: 1. name for sanity
2. With or without hash-function annotations at the top, you're in the same situation if you find a hash that you don't support while parsing. Annotation at the top would only help to reject a list altogether before parsing it. But why would you do that and not try to work with the entries that you can support in any case? Anyways, you can use any optional hints you want. We can standardize an optional hint name for this as long as it's optional.
3. I think you're finding the subtle issues. "whoops now my list includes also blake3 double-hashes" is bad to communicate in any way. "List A is done and you can follow up again on list B" is also not considered. It would be easier if we specify what the recommended behavior is when finding an unknown hashing function.

Also, we are talking a lot about hashing functions pertaining to the double-hash entries that are enconded as multihashes. You could also have normal `/ipfs/<cid>` entries, where the CID itself is not something that your underlying IPFS implementation understands: maybe you don't know the ipld codec, maybe the hash-function is unknown. In this cases, it would not be the "compact denylist parser component/blocker" the one that needs to be aware, but the underlying "ipfs" implementation. etc. but some implications might be similar.

So my practical approach is, 1) everything goes 2) try to parse what you can 3) strong recommendation or obligation to only use sha256 for double-hashing, everything else is non-spec-compliant (for now).

kdenhartog:
Would it be possible to include metadata tags about categories of why a line has been added? This would be useful for displaying in Brave as an enhancement to the messaging that normally gets displayed with safe browsing which seems like the best way to fit this in.

lidel (approved):
This has been ready for ratification since December, but took bit longer due to holidays and other work. No concerns were raised, merging.

Thank you @hsanjuan  for driving this, on both spec and implementation front, and for being patient :heart: 
Thank you to everyone who provided feedback and tested the reference implementations.

FYSA we shipped implementation in [Kubo 0.24](https://github.com/ipfs/kubo/blob/master/docs/changelogs/v0.24.md#support-for-content-blocking) and is also included in [Rainbow](https://github.com/ipfs/rainbow/).

anthonybudd:
I've just posted a question on r/IPFS and I think it would be better suited here.

---
I just added a file to IPFS that was a .txt file of the output of `printenv` from inside a docker container, when I went to the file’s address I was greeted by this 401 screen, I have been out of the IPFS community for a while and I have not seen this feature before.

The error message strikes me as strange and I have some questions about this feature

Who controls the list of blocked content? Is this list centralized?
Who decides what qualifies as “legal, abuse, malware or security reasons”?
Given the "interplanetary" nature of IPFS, what nation's laws are being used? For example I could share an image of Jon Venables in the USA but this would be illegal in the UK. If the UK court ruled this was illegal but the server was in the US, what would happen to my content in this case?
What safeguards are in place to prevent this feature being abused? For example, say a whistleblower published documents on IPFS, would it be possible for a government to leverage this feature to prevent it from being shared?
How does this feature work technically? If IPFS is a peer-to-peer protocol how is this (presumably) centralized list of banned content shared between nodes?


On the face of it

hsanjuan:
> I've just posted a question on r/IPFS and I think it would be better suited here.

This is not really the best place. Ask on the discuss.ipfs.io forums if you still wondering about these questions.

PR #384, opened 2023-03-15
docs: return 400 Bad Request on Invalid CIDs

hacdias:
See https://github.com/ipfs/go-libipfs/pull/205


hacdias:
ping @lidel

lidel (approved):
sgtm, make sure it is not lost and re-applied to https://github.com/ipfs/specs/pull/382 :pray:

PR #386, opened 2023-03-18
IPIP-0386: Subdomain Gateway Interop with _redirects

suicide:
This PR proposes a small change to how gateways handle redirects from a path URL to a subdomain URL.

Currently, if a resource in a path is not found, the gateway returns a 404 from the path URL. However, if the same path is requested on the subdomain URL, the gateway could use the `_redirects` file to find the requested resource.

I suggest that if a subdomain gateway provides interoperability with path gateway URLs it should prefer to redirect to the subdomain URL before resolving the resource.

This change would help Single Page Apps on IPFS when using native URLs.


aschmahmann:
triage: implementation of this has shipped in boxo and in kubo v0.21.0. This is in progress to be shipped, but needs to be rebased on `main` and ratified at the subsequent implementers sync.

@lidel will rebase when he can if nobody else beats him to it (help appreciated).

src/ipips/ipip-0386.md
```diff
@@ -0,0 +1,167 @@
+---
+title: "IPIP-0386: Subdomain Gateway Interop with _redirects"
+date: 2023-03-18
+ipip: proposal
```
lidel, line 4:
```suggestion
ipip: ratified
```

lidel (approved):
Rebased and made cosmetic editorial changes.

The fix proposed in this IPIP is already implemented in `boxo/gateway` and Kubo 0.22.

Will mention this during IPFS Implementers call as ready for ratification, but this is just a formality, we will most likely merge next week.

lidel:
No concerns raised during ratification period, merging!

Thank you @suicide  for taking time to write down rationale why this is important for long-term interop :pray:

PR #396, opened 2023-04-03
chore: fix broken link in IPIP-337

TorfinnOlsen:
Propose to update Delegated Content Routing Spec IPIP-337 to include up to date information regarding the deprecation of reframe in alignment with @guseggert efforts documented [here](https://github.com/ipfs/specs/pull/372).


lidel (changes requested):
@TorfinnOlsen thank you for submitting this. IPIPs are meant to be historical documents, they explain decision and state made at specific point in time. We want reader to be able to see Reframe looked at the time.

We do not want to rewrite history books here. At the time, afaik, we did not plan to fully kill Reframe, we just wanted a simpler API until Reframe gets proper resourcing.
As it turned out, nobody at PL, nor wider IPFS ecosystem cared to pick up Reframe, so the removal decision was made later ([specs](https://github.com/ipfs/specs/pull/372), [Kubo](https://github.com/ipfs/kubo/issues/9479)).

If you want to fix the broken link, perhaps change this PR to be just below?

```diff
- The [Reframe spec](https://github.com/ipfs/specs/blob/main/reframe/REFRAME_PROTOCOL.md) for delegated content routing is an experimental attempt at this,
+ The [Reframe spec](https://github.com/ipfs/specs/tree/ed02df474c18eb9fa1d5bb2aeecb05209102c373/reframe) for delegated content routing is an experimental attempt at this,
```

Issue #375, opened 2022-03-11
Need for IPIP: Data Onboarding via HTTP POST (and future ipfs:// POST|PUT)


lidel:
## Problem statement

HTTP Gateways are the most successful way for retrieving content-addressed data. Successful use of HTTP for retrieval use cases proves that IPFS does not replace HTTP, but augment it by providing variability and resiliency. IPFS over HTTP brings more value than the sum of its parts.

Removing the need for implementation specific RPC APIs (like one in Kubo) allowed not only faster adoption of CIDs on the web, but enabled alternative implementations of IPFS (like Iroh in Rust) to test compliance and benchmark thenselves against each other.

 While we have HTTP Gateways as a standard HTTP-based answer to the retrieval of data stored with IPFS (including verifiable [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw) and [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car) responses), the data onboarding over HTTP is currently done with vendor-specific APIs. 

The status quo at 2023 Q1 is pretty bad from the end user/developer’s perspective: every IPFS implementation, including online services providing storage and pinning services, exposes custom opinionated HTTP API for onboarding data to IPFS.

## Why we need IPIP for HTTP Data Onboarding

To illustrate, some prominent examples (2022 Q4):

<details>
  <summary>Click to expand :see_no_evil: </summary>

- Implementations
    - Kubo RPC  (AKA legacy /api/v0/..)
        - Is often used as a “standard HTTP API upload template” because it has commands for all onboarding needs:
            - [https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-add](https://web.archive.org/web/20221201011916/https://docs.ipfs.tech/reference/kubo/rpc/#api-v0-add)  – files and directories
                - FLAG: it uses custom form-data handling that requires special library for directory upload, which is an awful papercut for someone expecting simple upload with “curl” ([http://web.archive.org/web/20221201011916/https://docs.ipfs.tech/reference/kubo/rpc/#request-body](http://web.archive.org/web/20221201011916/https://docs.ipfs.tech/reference/kubo/rpc/#request-body))
                - FLAG: Kubo RPC was never designed to be used in browser context, and there are known bugs around the way it handles uploads (example: [https://github.com/ipfs/kubo/issues/5168](https://github.com/ipfs/kubo/issues/5168))
[...]

PR #401, opened 2023-04-17
IPIP-401: Data Onboarding with HTTP Semantics

hacdias:
This is the first part of #375, where we introduce data onboarding via the `POST` method.

- [ ] Write specification
- [ ] Write IPIP
	- [ ] Based on https://www.notion.so/protocollabs/wip-IPIP-Data-Onboarding-with-HTTP-POST-4c394b8ebb774f2d87d34466019257fc with @lidel in editors.
- [ ] Write implementation

### DIFF
--- a/src/http-gateways/writable-gateway.md
+++ b/src/http-gateways/writable-gateway.md
@@ -1,0 +1,46 @@
+---
+title: Writable Gateway Specification
+description: >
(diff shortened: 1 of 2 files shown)


lidel:
This never happened, and the draft had no feedback / lack of political / devgrant backing.
Let's close and continue in https://github.com/ipfs/specs/issues/375 if anyone is interested to see this.

PR #385, opened 2023-03-17
Publish IPIPs to specs.ipfs.tech

hacdias:
Based on #394.

A preview of what moving the IPIPs can look like. Nice things to get:

- [x] Rebase and update once https://github.com/ipfs/specs/pull/408 is merged.
- [x] Merge https://github.com/ipfs/specs/pull/394
- [x] "Related Issues" on the frontmatter, such that it is displayed in a similar way to the editors (depends on https://github.com/ipfs/spec-generator/pull/6)
- [x] Separate styling, maybe a different accent colour: yellow? orange? We should use something we already have.
- [x] Add IPIPs link in the main index.
- [x] Update the IPIP process to include right paths.
- [x] Decide on URL structure (done in https://github.com/ipfs/specs/pull/385#discussion_r1143876565)
- [x] Simplified listings by using a common include with the list. This way they don't get out of sync (thanks @darobin)

PR #402, opened 2023-04-17
IPIP-402: Partial CAR Support on Trustless Gateways

hannahhoward:
# Goals

Improve the capabilities of a trustless HTTP gateway by adding additional capabilities to requests for CAR files.

The goal is to enable a client capable of translating/decoding CAR files to make a single request to a trustless gateway that in most case allows them to render the same output generated via a request to a trusted gateway. (or if not in a single request, as few requests as possible)

# Additions That Are Non-Breaking

Adds two query parameters, <del>`car-scope`</del> `dag-scope=block|entity|all` and <del>`bytes`</del> `entity-bytes=from:to`, specific to CAR file requests only, that allow a more precise specification of the type of DAG to download at the terminus of a content path (CID+path). The default values for these parameters match current behavior, such that if they are not specified, the return CAR file matches currently specified behavior.

# Breaking changes

In order to serve requests for content paths other than just a CID root in a trustless manner, we are requiring gateways to return intermediate blocks from the CID root to the path terminus as part of the returned CAR file. This was not previously well specified, but most major HTTP Gateway implementations are currently only returning blocks starting at the end of the content path and will need to add intermediate blocks.

closes https://github.com/ipfs/specs/issues/348  cc https://github.com/ipfs/in-web-browsers/issues/128


lidel (commented):
Thank you for opening this @hannahhoward!

I don't think this is a breaking change, since by default, the change is to return MORE blocks, than requested.
There may be perf. penalty due to unexpected blocks, but there should not be any breakage. 

Initial thoughts in comments inline.

Once we are happy with the API, the next steps will be:
- [ ] dogfood this in Rhea for few weeks
- [ ]  implement it in `boxo/gateway` so `bifrost-gateway`, `kubo`, and probably  `boost` too, all benefit and expose the same API
  - tracked in https://github.com/ipfs/boxo/pull/303
- [x] add  IPIP  document  to this PR and include description of text fixtures we need
- [ ] write [conformance tests](https://github.com/ipfs/gateway-conformance) that use the same test fixtures, allowing others (e.g. https://github.com/web3-storage/freeway/issues/34) to test compatibility

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
...
+
+The following query parameters are only available for requests made with either a `format=car` query parameter or an `Accept: application/vnd.ipld.car` request header. These parameters modify shape of the IPLD graph returned within the car file
+
+### `car-scope` (request query parameter)
+
+Optional, `car-scope=(block|file|all)` with default value 'all', describes the shape of the dag fetched the terminus of the specified path whose blocks are included in the returned CAR file after the blocks required to traverse path segments.
+
+`block` - Only the root block at the end of the path is returned After blocks required to verify the specified path segments.
+
+`file` - For queries that traverse UnixFS data, `file` roughly means return blocks needed to verify the end of the path as a filesystem entity. In other words, all the blocks needed 'cat' a UnixFS file at the end of the specified path, or to 'ls' a UnixFS directory at the end of the specified path. For all queries that do not reference non-UnixFS data, `file` is equivalent to `block`
```
willscott, line 218:
```suggestion
`file` - For `{path}` segments traversed which are enumerable, only the minimum blocks needed for traversal should be returned. The final referenced object should be returned fully, but links should not be followed. In the specific instance of UnixFS data, these semantics are equivalent to the use of an [InterpretAs](https://ipld.io/specs/selectors/) [unixfs](https://github.com/ipfs/go-unixfsnode/blob/f4d6da0dbdf3614269b3580eeebe2bb5245f09bc/signaling.go#L47-L48) for the path, and as "unixfs-preload" for the terminal object.
```

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
...
+
+`all` - Transmit the entire contiguous DAG that begins at the end of the path query, after blocks required to verify path segments
+
+### `bytes` (request query parameter)
+
+Optional, `bytes=x:y` with default value `0:*`. When the entity at the end of the end of the specified path can be intepreted as a contingous array of bytes (such as a UnixFS file), returns only the blocks required to verify the specified byte range of said entity. Put another way, the `bytes` parameters can serve as a trustless form of an HTTP range request. If the entity at the end of the path cannot be interpreted as a continguous array of bytes (such as a CBOR/JSON map), this parameter has no effect. Allowed values for `x` and `y` are positive integers where y >= x, which limit the return blocks to needed to satify the range [x, y]. In addition the following additional values are permitted:
```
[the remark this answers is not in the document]
willscott: this seems reasonable.
i'd probably push to wait to attempt to define semantics in a non-car setting until we have a more concrete use case and can better understand what we need there.

alanshaw (changes requested):

src/http-gateways/trustless-gateway.md
```diff
@@ -8,28 +8,36 @@ editors:
...
 
-## `HEAD /ipfs/{cid}[?{params}]`
+Path parameters are only permitted for requests that specify CAR format. For RAW requests, only `GET /ipfs/{cid}[?{params}]` is supported
+
+## `HEAD /ipfs/{cid}[/{path}][?{params}]`
 
 Same as GET, but does not return any payload.
 
+Path parameters are only permitted for requests that specify CAR format. For RAW requests, only `HEAD /ipfs/{cid}[?{params}]` is supported
+
```
lidel, line 40: ## Proposal for feature detection via HTTP OPTIONS

Old gateways will ignore graph parameters and return CAR with the entire DAG.
We need a light mechanism for clients to detect which gateway supports partial CARs.

There is a good prior art for this in IPNI, where `OPTIONS` method returns headers when a specific feature is supported, and the lack of header means no support. 

This way light clients can adjust retrieval strategy and if graph is not available, do brute-force retrieval like the one in RAPIDE.


```suggestion

## `OPTIONS /ipfs/[{cid}[/{path}]]`

A client may discover the list of additional features supported by the trustless gateway server via sending `OPTIONS` request.

In response, the server may include HTTP headers:

- `Ipfs-Gateway-Format-Car-Verifiable-Paths` set to `true` indicates `/ipfs/cid/dir/file.jpg?format=car` will return a CAR with minimal set of blocks required for verifying the entire content path.
- `Ipfs-Gateway-Format-Car-Scopes` header key, with value as the comma separated list of `car-scope` parameters supported. 
  - For example, `Ipfs-Gateway-Format-Car-Scopes: block, all` means `/ipfs/cid/dir/file.jpg?format=car&car-scope=block` can be used to return all blocks for parent directories and only the root block for the `file.jpg`. 
- `Ipfs-Gateway-Format-Car-Bytes` set to `true` to indicate range requests via `?format=car&bytes=from:to` are supported.


The absence of a header implies that the server does not offer support  specific feature and that it will always return the entire DAG behind the requested CID.
```

src/http-gateways/path-gateway.md
lidel: @willscott I am afraid linking/referring to implementation details from  `go-unixfsnode` source code is not something we can do in specs. 

If there is no easier way, we need to define these selectors in the spec before refering to them.

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: Correct.  

`?format=car&bytes=` is concerned only with narrowing down the number of blocks. You use these blocks the same way you would if they arrived over bitswap or from local block store—so a different abstraction.

src/http-gateways/trustless-gateway.md
lidel: Would appreciate feedback on ergonomics of doing feature-detection this way from  potential client/server implementers  (cc @hannahhoward @guanzo @olizilla  @hacdias @John-LittleBearLabs @fabricedesre   for visibility)

src/http-gateways/path-gateway.md
lidel, line 218: ## Is there a better name for this `car-scope` than `file` ?

Now is the last call to propose something better (cc @aschmahmann)
(some things i've heard: `blob`, `adl`, `bytes`, `data`)

src/http-gateways/path-gateway.md
```diff
@@ -567,7 +591,10 @@ The following response types require an explicit opt-in, can only be requested w
 - Raw Block (`?format=raw`)
   - Opaque bytes, see [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw).
 - CAR (`?format=car`)
-  - Arbitrary DAG as a verifiable CAR file or a stream, see [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car).
+  - A CAR file or a stream that contains all blocks required to trustless verify the expressed query, see [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car).
+    - Must contain, as the very first block, the block that hashes to the content root CID
+    - Must contain, immediately following the root block, all blocks encountered while traversing the expressed path in the order they were traversed
+    - Must contain, immediately following traversed path blocks, appropriate blocks in depth first traversal order required to verify the query expressed at the terminus of the path in [query parameters](#query-parameters-for-car-requests)
```
aschmahmann, line 597: I think we should clarify here whether it is expected that responses here should be byte-wise identical. I'd like us to explicitly make this a yes/no rather than having people guess.

Some ways in which you could have a CAR file not contain exactly the same bytes despite having a consistent traversal order:
- Different CAR version (e.g. CARv1 vs CARv2)
- Adding extra CAR features (e.g. indices, or if we add a solution for signaling EOF in CARs as a remedy for https://github.com/ipfs/specs/pull/332 while still calling it a CAR, etc.)
- How many and which CIDs are in the header

My inclination is to explicitly call out that `format=car` **does not reproduce deterministic results**. If people want reproducible results (e.g. for whatever DAG House is doing with CAR CIDs, or what Filecoin is doing with converting CARs to Piece CIDs) then we should add more parameters that describe exactly what it is they want so that we can change the defaults later.

This is similar to the situation we're in with `ipfs add` in kubo.

src/http-gateways/trustless-gateway.md
```diff
@@ -8,28 +8,36 @@ editors:
...
 
-- data is requested by CID, only supported path is `/ipfs/{cid}`
-- no path traversal or recursive resolution, no UnixFS/IPLD decoding server-side
 - response type is always fully verifiable: client can decide between a raw block or a CAR stream
+- no UnixFS/IPLD decoding server-side
+- for CAR files:
+  - the behavior is identical to :cite[path-gateway]
+- for raw blocks:
+  - data is requested by CID, only supported path is `/ipfs/{cid}`
+  - no path traversal or recursive resolution
```
aschmahmann, line 23: This seems confusing. It seems like the minimal implementation is just to support blocks and not CARs. CARs is going to support some IPLD or UnixFS decoding server side.

aschmahmann (changes requested):

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
 - `format=json` → `Accept: application/json`
 - `format=cbor` → `Accept: application/cbor`
 
+## Query Parameters for CAR Requests
+
+The following query parameters are only available for requests made with either a `format=car` query parameter or an `Accept: application/vnd.ipld.car` request header. These parameters modify shape of the IPLD graph returned within the car file
```
hacdias, line 210:
```suggestion
The following query parameters are only available for requests made with either a `format=car` query parameter or an `Accept: application/vnd.ipld.car` request header. These parameters modify shape of the IPLD graph returned within the car file.
```

src/http-gateways/trustless-gateway.md
guanzo: > This request if often sent by web browser anyway, so we would not be introducing much overhead.

The OPTIONS request/response isn't accessible to userland JS, so browser clients can't see these resposne headers and know what features are supported.

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
...
+
+Optional, `car-scope=(block|file|all)` with default value 'all', describes the shape of the dag fetched the terminus of the specified path whose blocks are included in the returned CAR file after the blocks required to traverse path segments.
+
+`block` - Only the root block at the end of the path is returned After blocks required to verify the specified path segments.
+
+`file` - For queries that traverse UnixFS data, `file` roughly means return blocks needed to verify the end of the path as a filesystem entity. In other words, all the blocks needed 'cat' a UnixFS file at the end of the specified path, or to 'ls' a UnixFS directory at the end of the specified path. For all queries that do not reference non-UnixFS data, `file` is equivalent to `block`
+
+`all` - Transmit the entire contiguous DAG that begins at the end of the path query, after blocks required to verify path segments
+
+### `bytes` (request query parameter)
```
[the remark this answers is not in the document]
jbenet: (disregard this comment **_iff_** the only things that can be retrieved from this endpoint are unixfs files)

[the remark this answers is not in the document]
lidel: Ack, we probably need to prefix this with the same word we use in `car-scope=` (currently `file`, but we discuss alternatives in https://github.com/ipfs/specs/pull/402/files#r1174018410

BigLep:
Related tracking items that would be good to add here:
- dag.house work
- Lassie work
- Boxo work
- ipfs/gateway-conformance work

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
...
+
+### `car-scope` (request query parameter)
+
+Optional, `car-scope=(block|file|all)` with default value 'all', describes the shape of the dag fetched the terminus of the specified path whose blocks are included in the returned CAR file after the blocks required to traverse path segments.
+
+`block` - Only the root block at the end of the path is returned After blocks required to verify the specified path segments.
+
+`file` - For queries that traverse UnixFS data, `file` roughly means return blocks needed to verify the end of the path as a filesystem entity. In other words, all the blocks needed 'cat' a UnixFS file at the end of the specified path, or to 'ls' a UnixFS directory at the end of the specified path. For all queries that do not reference non-UnixFS data, `file` is equivalent to `block`
+
+`all` - Transmit the entire contiguous DAG that begins at the end of the path query, after blocks required to verify path segments
```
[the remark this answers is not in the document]
rvagg: Good question, if we're first-classing unixfs in all of this then it's not unreasonable to suggest we should interpret "all" via a unixfs lens. This would require more complexity though, we don't have enough pieces implemented via go-unixfsnode to make this workable just yet I think.

Another variation might be a unixfs link pointing to a non-unixfs graph, should that be followed, result in an error, be ignored?

src/http-gateways/path-gateway.md
```diff
@@ -567,7 +591,10 @@ The following response types require an explicit opt-in, can only be requested w
 - Raw Block (`?format=raw`)
   - Opaque bytes, see [application/vnd.ipld.raw](https://www.iana.org/assignments/media-types/application/vnd.ipld.raw).
 - CAR (`?format=car`)
```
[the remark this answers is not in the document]
rvagg: ... but that currently such an error case must be determined by the consumer of the CAR in lieu of a good [error signalling mechanism](https://github.com/ipfs/specs/pull/332).

src/http-gateways/path-gateway.md
rvagg: Are you suggesting that block ordering and block inclusion could be non-deterministic too? If we could lock that in then it satisfies a whole lot of concerns around incremental verifiability.

But on the other ways that a CAR could be non-deterministic—I think the main problem could be the possibility of abuse in being able to craft CARs that are excessively large and wasteful. CARv2 has optional padding on both sides of the payload, so you can make them very large. A malicious server could begin a CARv2 with a very large data padding and just drip-feed feed random bytes. Or an index padding that's excessively long—which can be protected against on the client side (when CARv2, stop reading after CARv1 payload), but it's yet more logic to be inserted into this process just to protect against a class of misbehaviour.

rvagg: I'm considering turning this doc into an extension to the CARv1 spec for "deterministic CARs", similar to how some data format specs, such as CBOR, have sections with recommendations for deterministic forms: https://github.com/filecoin-project/lassie/blob/main/docs/CAR.md, it could serve as an explication of the 3 dot-points listed here.

rvagg: (and I'd like to get a good set of fixtures for these too, which we have partly in our Lassie test suite, but they need to be more portable and cover a few more cases)

src/http-gateways/path-gateway.md
```diff
@@ -567,7 +591,10 @@ The following response types require an explicit opt-in, can only be requested w
...
-  - Arbitrary DAG as a verifiable CAR file or a stream, see [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car).
+  - A CAR file or a stream that contains all blocks required to trustless verify the expressed query, see [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car).
```
John-LittleBearLabs, line 594: I wonder if it might be reasonable to have an optional feature where one can request a CAR file containing as large a subset of this that can be fetched immediately (for some reasonable definition of immediate, perhaps whatever is in your local store). If one is combining results from multiple gateways this might be a reasonable compromise between number of round trips and the length of any one request.

src/http-gateways/trustless-gateway.md
John-LittleBearLabs: I think this seems like a fantastic idea.

I don't know if HTTP headers are easier/better than doing something like JSON in the body, but it would certainly have worked for me up to this point. I suppose JSON (or similar) would have the option to arrange the data differently, e.g. `"Formats": [ "car", "raw", "dag-cbor", "dag-json" ]`. I guess the downside is that right now I think it's theoretically possible to have a useful client that doesn't actually know how to parse JSON, and headers might be a more primitive thing to read.

Would we want to use a feature-specific prefix? Like `Ipfs-Gateway-Feature-` so a client could _note_ the presence of features it's unaware of (distinguishing it from headers that aren't feature announcements)? I'm not sure exactly what one could do with that other than log it, but... :shrug: doesn't cost much.

I think my main feedback would be to add more features to what can be detected. At the moment it may seem redundant to have older features, since software versions tend to be rather linear and this request would be a newer feature than e.g. ?format=raw. But having optional features that are not assumed to be present by the rest of the ecosystem lowers the barrier to entry for new gateway implementations.

willscott:
> I think my main feedback would be to add more features to what can be detected. 

are you specifically interested in feature detection of what codecs are supported by the gateway, or rather if the gateway is able to support e.g. the car scope's being introduced here?

src/http-gateways/path-gateway.md
John-LittleBearLabs: Alternatively, a server-side timeout, to give the gateway an idea of how much effort it should put into collecting all the things. Something like `?timeout=1` implying that the response might be incomplete or, in the case of IPNS, expired.

src/http-gateways/path-gateway.md
rvagg: 3 fairly separate bits of feeback related to this, I hope you don't mind me doing it here:

1. Specifically on this piece, the language I've adopted for explaining this is "unixfs entity"; but "entity" is the more generic form which translates to: if unixfs, then the _thing_ that's here: file or directory, all blocks of sharded file or directory, _or_ if not unixfs then just this block. I see the language in this doc has "entity" too so I'd suggest leaning in to that.

2. I've been wondering what this looks like if/when we escape our unixfs prison and have other ADLs at play. What if WNFS shows up with an ADL we can run that returns >1 block and it makes sense to include at this point. I think "entity" makes sense for that but I'm wondering how else we could make this more future-proof. Embedding the implicit unixfs seems like one problem we have and that special-case only shows up for "entity" and pathing and we do have an opportunity in this specific spec to do something about that—e.g. requiring an additional parameter to opt-in to unixfs special-casing, which would also make room for other ADLs. `bafyfoobar/path/to/thing?adl=unixfs&car-scope=entity`.

3. I'm not a fan of `car-scope`, because I don't see why we have to lock it in for "car". We're likely to have additional block containers in the future that we may not call a "car" (there's been >1 proposals for a CAR successor that are different enough to not naturally be a CARv3). But also, what the parameter is really doing is specifying a traversal depth and style. Even `dag-scope` would be a better name. I'd combine one of `dag`, `graph` or `traversal`, with one of `scope`, `type` or `mode`. "car" is the one that doesn't fit and it's already specified in the format.

src/http-gateways/path-gateway.md
rvagg: Pulling in @John-LittleBearLabs's comment above re responses _"containing as large a subset of this that can be fetched immediately"_: I think we could make that allowable as long as they are still in the right traversal order. What I want to be able to do is traverse the graph on the client side of this to confirm that the blocks are as expected. I _could_ error when the response contains a surprise block that's not the one I expected, alternatively I could skip over the ones I expect, exhausting the traversal, assuming that the block I was given might be later in the traversal because the server had dead-ends that it couldn't serve. It would be a shame to bail at the first sign of a missing block if the server has a lot more of the graph available.

What I don't want to do is have a memory pool of random blocks I was given that _may or may not_ be related to the graph I care about. I could do this with a single block while I run through a traversal looking for where the gap in blocks ends and where to pick up again.

That nicely leaves open the option to go fishing elsewhere for other parts of the graph that are missing, but maximising what a single server can provide.

(In many ways we're defining a dumber graphsync with this protocol so can learn lessons from it: https://github.com/ipfs/go-graphsync/blob/1988c8f1920ba9e558dd85ceaf84c0c76e0f46de/graphsync.go#L129)

John-LittleBearLabs: To push back on myself slightly, it might be worth considering the interplay weirdness between allowing gaps to exist (partial response) and a byte range (`bytes=x:y`). To be on the safe side, perhaps it should be considered an invalid request if you try to do both at the same time.

rvagg: Well, this is where client validation comes in - it's still an erroneous response as far as a client is concerned. Any "verifiable CAR" validation tools we write should say that "this is not a verifiable CAR". But it is a useful bundle of blocks we can use to work toward a verifiable CAR.

Until we have better error signalling mechanisms then we're a bit stuck with either "terminate" or "keep going and try our best" on the server side.

aschmahmann: > Are you suggesting that block ordering and block inclusion could be non-deterministic too? If we could lock that in then it satisfies a whole lot of concerns around incremental verifiability.

My main ask here is to define very explicitly if responses should be bytewise identical, and then if not what constraints and freedoms are implementers allowed. There are some people in our ecosystem who have been encoding CAR files in a very specific way and then assuming/hoping that other tooling behaves the same way or continues to behave exactly as it does today. This leads to similar problems as with people assuming kubo's `ipfs add` will always produce the same UnixFS graph by default.

I explicitly gave a few examples that had nothing to do with incremental verifiability or anything IPLD traversal related to demonstrate the potential flexibility here.

However, yes even in the realm of incremental verifiability there are multiple plausible ways to do block ordering and inclusion. For example:
- Instead of doing a specific kind of traversal we could have any valid traversal traversal order (e.g. for a UnixFS HAMT enumeration the blocks could come in any order rather than just a depth first one)
- Duplicate blocks could or could not be included. Duplicate blocks trade off more (although gzip might help here) data transfer in exchange for clients needing less memory if they're streaming outputs out. It's not unfathomable that both are desirable options for different use cases

> I'm considering turning this doc into an extension to the CARv1 spec for "deterministic CARs"

Sure, we can nail down both a spec for a deterministic CARv1 and a specific deterministic response type here (they might be the same or come with extra conditions around say what goes in the headers) if people would like this to be deterministic.

While I don't feel strongly about names inevitably someone will come around with a different way of creating the CAR (they want CARv2s, they want CARv2s with different indices, they want/don't want duplicate blocks, etc.) and will need a way to signal that. I don't know if `format=car` should mean "one of the many ways to encode this graph into a CAR that was blessed in mid 2023" or if perhaps we want a more explicit signal indicating that the format requested is the precise deterministic format being specified rather than just a valid CAR.

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
aschmahmann: In this particular case the consumer of the CAR should be able to determine this as well so that they can't be lied to. For example, if you ask for `/ipfs/bafyDir/file` and it exists but I return to you an error saying it doesn't you should know I'm wrong and be able to ask someone else for the data.

src/http-gateways/path-gateway.md
aschmahmann: > I hope you don't mind me doing it here

Not at all, and I also tend to agree with all of the points you've made.

> entity

Sure, not attached to `shallow`, `node`, etc. I just want something that isn't obviously misleading.

> I'm not a fan of car-scope ... I'd combine one of dag, graph or traversal, with one of scope, type or mode. "car" is the one that doesn't fit and it's already specified in the format.

I like your style 😄 and agree with your assessment that CAR feels weird here because what's being described is a graph-ish operation not a "how are the bytes stored/transmitted" operation. The only counterpoint I could see here is that this feels format specific because the way we're expressing the various entities when describing how the graph should be sent to us in its rendered format uses HTTP Headers rather than query parameters which makes this a little bit format related.

For example `/ipfs/bafyfoo/path/to/directory?dag-scope=all` is invalid rather than say implying `format=tar` unless an alternative format like `car`, `iso`, `zip`,... was specified.

> I've been wondering what this looks like if/when we escape our unixfs prison and have other ADLs at play. What if WNFS shows up with an ADL we can run that returns >1 block and it makes sense to include at this point.

I too would very much like to escape UnixFS prison. To that extent if there is a good word meaning approximately "single logical unit of data" without any recursive properties it'd be nice to use that so that people are familiar with it if/when support for non-UnixFS multiblock data structures emerges.

However (and somewhat off topic, although I'd be interested to engage with it further if you'd like):
- I don't think `?adl=` is going to be what saves us here since you might have multiple ADLs to apply. Instead I suspect we end up with something a little closer to https://github.com/ipfs/specs/pull/293 but in some situations implemented as closer to `ipfs://bafyfoo.... rest of ipld uri semantics?path=ipld` so that we can have things like escape characters, multiple transformations in a given path, etc.
- I suspect that given the opportunity to try things again (e.g. with IPLD paths) that something like `ipfs://bafy-unixfs-dir;adl=unixfs/subdir?path=ipld&format=car` would by default have `dag-scope=entity` rather than `dag-scope=all` since that matches what we do in the non-car case.

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
aschmahmann: > we don't have enough pieces implemented via go-unixfsnode to make this workable just yet I think.

I can't speak to that, although it seems a bit weird if true given that we can do this with what was formerly go-unixfs. I suspect it's probably doable with go-unixfsnode but if anything the difficulty would arise in expressing this via selectors which I think might be doable as well.

> Another variation might be a unixfs link pointing to a non-unixfs graph, should that be followed, result in an error, be ignored?

Yeah, that's perhaps an even better example. Mostly we should clarify what's going on here. We might be able to punt on what the "ideal" model is here since IIUC for non-malformed UnixFS "all as IPLD" or "all as UnixFS" are the same. However, this probably means more pressure on tools like https://github.com/ipfs/gateway-conformance to make sure we handle malformed UnixFS uniformly by erroring.

My understanding (from watching a several hundred message long Slack thread that has long since been GC'd because people didn't use public forums 😢) is that not erroring on undefined edge cases is the reason why you can do `/ipfs/some-dag-cbor-thing/path/to/unixfs-dir/file`. So I'd like to keep these cases covered so that we don't get spec expansion by accident (e.g. people trying to hookup non-unixfs graphs to unixfs ones and asking gateways to traverse them).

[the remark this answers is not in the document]
rvagg: ACK on the ADL thing, we shouldn't drop that line of thinking and you're right that something in the region of @rangermauve's experiments are probably the path forward here. I'd like to mature one of my side experiments to the point where a stacked ADL makes sense so we can push on this a bit.

lidel:
I've resolved conflicts and rebased this on top of #385. Will finish and add initial version of the IPIP document tomorrow.

I also plan to extract `OPTIONS`-based feature detection to a separate IPIP, just to make things easier to review and discuss, as it will cover features beyond this IPIP.

hannahhoward:
Want to summarize some thoughts about determinism.

Byte-identical responses seems like a very high bar for a first pass, and I'd like to take that off the table.

So then there are a series of constraints to consider. The baseline requirement is the CAR contains all the blocks needed to verify the query. Then we ask:
1. Are duplicate blocks allowed in a valid response?
2. Are extraneous blocks (blocks not in the expected requested CAR allowed) as long as the overall file is included?
3. Are identity CIDs allowed?
4. Is CarV1 / V2 specified?
5. Is there a deterministic ordering to blocks?

Ultimately we're talking about balancing work between server & client. Less constraints makes it easier to write a performant server, and harder to write a performant client. More constraints does the opposite.

For the client, @rvagg 's concern is the most relevant:

> What I don't want to do is have a memory pool of random blocks I was given that may or may not be related to the graph I care about. I could do this with a single block while I run through a traversal looking for where the gap in blocks ends and where to pick up again.

Holding the whole response, either in RAM or disk, in order to verify it feels like a deal breaker for most clients.

There's a secondary concern around time to verification failure -- in other words, how quickly can one determine a response is incorrect (i.e. can I know at each block if the response is correct?). However, this feels less critical if an increase in time to verification failure does not increase memory usage.

From my point of view, deterministic ordering is the most important. If you're willing to sacrifice time to verification failure, you can basically relax all other constraints. Essentially, a client can:

1. safely ignore and drop duplicate blocks
2. safely ignore and drop extraneous blocks
3. safely ignore and drop carv2 padding
4. safely ignore and drop identity blocks

I'd rather avoid a complex signaling mechanism around traversal order on first pass and just say depth first. We can building an ordering parameter later.

mikeal:
I was going to comment but i agree with everything @hannahhoward said so I don’t need to 😁

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: IMO it is out of scope: malicious server will not send proof anyway, just HTTP 404 without body, so a robust client will have to have recovery logic anyway for manually traversing the path segment-by-segment, using multiple gateways, until a proof of non-existence is obtained.

src/http-gateways/path-gateway.md
```diff
@@ -205,6 +205,30 @@ This is a URL-friendly alternative to sending an [`Accept`](#accept-request-head
...
 
+## Query Parameters for CAR Requests
+
+The following query parameters are only available for requests made with either a `format=car` query parameter or an `Accept: application/vnd.ipld.car` request header. These parameters modify shape of the IPLD graph returned within the car file
+
+### `car-scope` (request query parameter)
+
+Optional, `car-scope=(block|file|all)` with default value 'all', describes the shape of the dag fetched the terminus of the specified path whose blocks are included in the returned CAR file after the blocks required to traverse path segments.
+
+`block` - Only the root block at the end of the path is returned After blocks required to verify the specified path segments.
```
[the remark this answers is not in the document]
lidel: Added a note to IPIP about things like "web" scope being out of scope, but something we could add in future IPIP, this falls under the same umbrella.

OPTIONS will be a separate IPIP, just to keep this one easier to review.

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: I've pushed IPIP with some examples of test scenarios.

Gateways support traversing DAG-CBOR  like `/ipfs/some-dag-cbor-thing/path/to/unixfs-dir/file`.

> Does this intentionally use the underlying IPLD set of links rather than logical data?

Yes,  for the meaning of `all`, I've added a note about following all IPLD Links recursively for the sake of simplicity -- lmk if that is good enough.

src/http-gateways/trustless-gateway.md
lidel: Note to self: i will move this to a separate IPIP, and we will continue there

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: renamed: https://github.com/ipfs/specs/pull/402#issuecomment-1544819745

src/http-gateways/trustless-gateway.md
lidel: Yes, this looks like a bad mental shortcut.
I've replaced  "no UnixFS/IPLD decoding server-side" with "no UnixFS/IPLD deserialization"

src/http-gateways/path-gateway.md
lidel: > call out that format=car does not reproduce deterministic results

I've removed mentions of ordering/determinism from this IPIP and added call out (https://github.com/ipfs/specs/pull/402/commits/98191b5fec841d6ed67975efc457744373901b66).
I think we want to move this to a separate IPIP, as noted in https://github.com/ipfs/specs/pull/402#issuecomment-1544819745. 

My current thinking is (notes for future IPIP):
-  this type of thing should be explicit (a flag for order, and one for duplicates), not implicit, 
- clients need a mechanism to explicitly opt-in into specific CAR flavor (`Accept: application/vnd.ipld.car;ord=dfs;dup=t;q=0.9, application/vnd.ipld.car`  header or query params, to avoid issues with HTTP caching, `Etag`, and legacy gateways), 
- gateways / implementers shoudl be free to decide what is the implicit default for this, but then be clear about it in very explicit `Content-Type: application/vnd.ipld.car;version=1;ord=dfs;dup=t` present in responses

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: Renamed to `entity` and removed CLI terms.

As for directories, I agree we want to go with third, safest option.
Added test fixture description to test exactly that;
> The response MUST include the minimal set of blocks required for enumeration of directory contents, and no blocks that belong to child entities.

src/http-gateways/path-gateway.md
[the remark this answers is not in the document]
lidel: renamed:
- car-scope=file → dag-scope=entity
- bytes → entity-bytes

src/http-gateways/path-gateway.md
lidel: (went with `entity`, as Rod sugested – https://github.com/ipfs/specs/pull/402#issuecomment-1544819745)

src/http-gateways/trustless-gateway.md
```diff
@@ -8,28 +8,36 @@ editors:
...
 
 # HTTP API
 
 A subset of "HTTP API" of :cite[path-gateway].
 
-## `GET /ipfs/{cid}[?{params}]`
+## `GET /ipfs/{cid}[/{path}][?{params}]`
 
-Downloads data at specified CID.
+Downloads data at specified **immutable** content path.
```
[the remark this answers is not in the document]
lidel: Yes, we were afraid to expose this, and  https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor states:

> Note: pathing into IPLD Kind other than Link (CID) is not supported at the moment.
> Implementations should return HTTP 501 Not Implemented when fully resolved content path has any remainder left.

hannahhoward:
>   * my understanding  is that we all want to solve ordering for light clients, but it should be explicit behavior signaled via native HTTP mechanism like `Accept` and `Content-Type` headers: legacy gateway exist, and we don't control when they update, client needs to know what they got back

@lidel I'm not sure `Accept` or `Content-Type` make sense unless we're defining a new ordered CAR format.

The only concern I have is on a concrete level Lassie & Caboose need deterministic ordering from the start. As stated before, throwing the whole CAR in RAM (or disk) before verifying a single block just doesn't work for the Rhea case, where lots and lots of retrievals are being verified at once. So while it might not be in the spec, we're gonna just drop out of order blocks via the approach outlined above. Maybe this isn't a spec discussion, but it's relevant in that we're saying "if you serve HTTP for Rhea, we need some ordering".

src/http-gateways/path-gateway.md
aschmahmann: Responding to https://github.com/ipfs/specs/pull/402#issuecomment-1544453059

👍 on not restricting `format=car` to have a defined byte output, I just don't think we're at a place where we can do that yet given the lack of consensus on the "best" way to return a CAR.

> Ultimately we're talking about balancing work between server & client. Less constraints makes it easier to write a performant server, and harder to write a performant client. More constraints does the opposite.

👍

> I'd rather avoid a complex signaling mechanism

Idk how complicated this is going to end up being, my understanding of the existing proposals are to:
1. Just ask the server what they support and then choose what you want explicitly (or yolo and don't specify if you don't care)
2. Maybe allow passing options using standard HTTP negotiation around Accept headers, etc.

IIUC these are both pretty normal HTTP negotiation tactics and negotiation here makes sense since "we're talking about balancing work between server & client" is roughly equal to "the client and server to need to **negotiate** how much work each of them will do".

> around traversal order on first pass and just say depth first. We can building an ordering parameter later.

If we were going to indicate that the blocks MUST be in a traversal order (as opposed to say having the root block at the end of the CAR) AND that we should specify that traversal order then DFS seems reasonable for most cases of UnixFS data. However, literally any traversal order would solve the "I downloaded a large amount of garbage before I realized I didn't want this data" (otherwise known as incremental verifiability) problem. Even not defining the traversal order (as long as it's a valid traversal) would work, albeit with perhaps more memory usage.

From a Slack thread:
> We cannot ... put whole CAR files (likely many at once) in memory or on disk before they verify a single block

IIUC (someone please correct me if I'm wrong) there are currently no people advocating for the data to come in a non-traversal order (e.g. the natural UnixFS encoding order where the root is at the end of the CAR). As long as some traversal order is used this problem goes away.

> Holding the whole response, either in RAM or disk, in order to verify it feels like a deal breaker for most clients. There's a secondary concern around time to verification failure

Does this indicate that you'd be in favor of having duplicate blocks supported in some way? Not having the ability to support duplicate blocks in the current CAR format means that you need to keep the entire response in RAM or disk to verify the response unless you are willing to use caching heuristics and follow-up your CAR requests with block requests to fill in the blanks for any deduplicated blocks that you deleted despite needing them.

IIUC the tradeoff of yes/no on duplicate blocks is that not having duplicate blocks is:
- Good: Fewer bytes over the wire
- Ok: Works reasonably if the data is being written to disk
- Bad: In order to implement streaming responses the client logic has to be more complex and potentially make additional network requests to compensate for not keeping all the data buffered in RAM/disk

lidel: I've tried to synthesize  things, and extracted CAR block ordering spec work to [IPIP-412](https://github.com/ipfs/specs/pull/412), let's continue ordering discussion there.

lidel:
I've  extracted CAR block ordering spec work to [IPIP-412](https://github.com/ipfs/specs/pull/412), let's continue ordering discussion there to keep this IPIP limited to "partial DAG export": pathing, `dag-scope` and `entity-bytes`.

src/ipips/ipip-0402.md
```diff
@@ -0,0 +1,337 @@
...
+
+It is impossible to know if some entity on a sub-path is a file or a directory,
+without sending a probe for the root block, which introduces one round-trip overhead
+per entity.
+
+This problem is not present in the case of `dag-scope=entity`, which shifts the
+decision to the server, and allows for fetching unknown UnixFS entity with a
+single request.
+
+## Test fixtures
```
lidel, line 197: @guanzo @tchardin would appreciate feedback on below test scenarios. (cc @laurentsenta  @galargh  for visibility).

The intention here is to define the bare minimum we need in gateway-conformance to ensure independent implementations of IPFS do the right thing. Anything to add? Remove/simplify?

guanzo: WIP PR for car-scope https://github.com/ipfs/gateway-conformance/pull/49/files

lidel: cc @aschmahmann as he works on `boxo/gateway` implementation (https://github.com/ipfs/boxo/pull/303)

src/http-gateways/path-gateway.md
lidel: - We have something like  the first idea for blocks. 
  - Requesting `?format=raw` with [`Cache-Control: only-if-cached`](https://specs.ipfs.tech/http-gateways/path-gateway/#only-if-cached) in Kubo should abort if it is not in local cache.  
  - If we want something similar for CARs, caching gets tricky, we don't want to cache partial streams, so it sounds like a new parameter `?format=car&dag-scope=cached` is needed -- not against it, but out of the scope of this IPIP., could be future one.
- As for server-side timeouts, they make sense on paper, but in practice client has to implement  own timeout anyway and the server will abort when client cuts the connection (just to be safe when interacting with random gateways on the internet, often running old/custom software/forks). Probably better to spend time elsewhere.

src/http-gateways/trustless-gateway.md
```diff
@@ -66,12 +74,76 @@ Below response types MUST to be supported:
...
+
+- `*` can be substituted for end-of-file
+  - `entity-bytes=0:*` is the entire file (a verifiable version of HTTP request for `Range: 0-`)
+- Negative numbers can be used for referring to bytes from the end of a file
+  - `entity-bytes=-1024:*` is the last 1024 bytes of a file
+    (verifiable version of HTTP request for `Range: -1024`)
+  - It is also permissible (unlike with HTTP Range Requests) to ask for the
+    range of 500 bytes from the beginning of the file to 1000 bytes from the
+    end: `entity-bytes=499:-1000`
+
```
rvagg, line 131:
```suggestion

When the entity is a sharded UnixFS file, and the `from` value is _not_ `0`, the encoded size (`Tsize`) values of the individual shards that make up the file will be implicitly trusted to determine the block that contains that offset. Because this cannot be fully verified for correctness without having all blocks from the start of the file content, consumers of this data must trust that the producer of the data has properly encoded the blocks of the UnixFS sharded file. If a stronger guarantees are required, byte ranges should be avoided, or should be anchored with a `from` of `0`.

```

rvagg: Oh look! Real-world use proving this problem: https://github.com/ipfs/js-ipfs-unixfs/issues/335

aschmahmann: @rvagg dag-pb TSize has nothing to do with `entity-bytes` at all. The UnixFS `filesize` and `blocksizes` (protobuf elements 3 and 4 of the Data field of the dag-pb nodes for UnixFS files) does.

rvagg: ok, different value, same problem though; the value is encoded by the producer of the blocks and has to be implicitly trusted to get the right byte range

aschmahmann: Sure, the issue with malformed graphs still exists along with the trust problems you mentioned. It shows up here, but it also likely belongs as an implementers note on the trusted gateway spec as well.

I was just calling out that:
1. TSize is the wrong field
2. Unlike TSize `filesize` and `blocksizes` have actual definitions which means when they're not obeyed the UnixFS graph is malformed unlike the argument about TSize where people seem more hesitant to give it an actual definition (although it seems to have been defined as cumulative graph size since day 1 https://github.com/ipfs/go-merkledag/commit/63e4477573afa18090eb45e6b836bdccacfa8d89#diff-10837cc3557cec0045183193a03f17af589591f2be0753262027534bb8f64ad9R38-R39)

src/http-gateways/trustless-gateway.md
```diff
@@ -66,12 +74,146 @@ Below response types MUST to be supported:
...
+Value returned in `CarV1Header.version` struct  MUST match the `version`
+parameter returned in `Content-Type` header
+
+#### CAR roots
+
+:::issue
+
+TODO: we need to specify expectations about what should be returned in
+[`CarV1Header.roots`](https://ipld.io/specs/transport/car/carv1/#header).
```
lidel, line 185: Main concern here is that default behavior of many tools (like Kubo) is to  pin every CID from the `CarV1Header.roots` list recursively on `.car` import.

I've added 4 options (https://github.com/ipfs/specs/pull/402/commits/bfb5d3cf1d51225f0748c32e672915169d9eae4f, below) to make it easier for us to decide which one should be part of the spec.

- Options A and C will work with `ipfs dag import` in Kubo, which has `--pin-roots=true` as the implicit default, but we no longer can assume root CIDs are always present.
- Options B and D  will trigger some over-fetch, but are deterministic, always provide meaningful root information:
  - B is probably the least painful and acceptable for real-world use, the worst case it will overfetch the terminating entity, but we won't pin or overfetch parents.
  - D has a nice property of matching `X-Ipfs-Roots`, which we generate anyway, but is the worst case for all past and current  Kubo versions: `ipfs dag import` will fetch entire Wikipedia when a CAR has only one article + blocks for parents, unless we make significant changes to `ipfs dag import` in Kubo.

I am leaning towards B or D (if we are willing to break/cleanup how `ipfs dag import` works)

cc @rvagg @aschmahmann @hacdias

lidel (changes requested):
I've pushed some wip Payload section to clarify the gap was identified in https://github.com/ipfs/gateway-conformance/pull/56#discussion_r1219591806.

src/http-gateways/trustless-gateway.md
lidel: Ok, B is the safest one:
- existing gateways (incl. Kubo  0.20) already do this for `?format=car`
- keeping the behavior avoids breaking change for existing users (overfetching may caus OOM/storage/excessive costs)

This unblocks this IPIP without increasing the scope.
We can always add D in future IPIP, when we have CAR format / content path syntax for expressing partial DAGs with parents of content paths.

lidel: Done: https://github.com/ipfs/specs/pull/402/commits/0639715a6f895c710a682faf00532a21dcb3beff

lidel: @hannahhoward  any concerns around this requirement? (this is what existing ipfs.io does).

> If the response uses version 1 or 2 of the CAR spec, the [`CarV1Header.roots`](https://ipld.io/specs/transport/car/carv1/#header) field MUST contain the CID of the terminating entity.

rvagg: @lidel can you clarify (both here and in the doc) what you mean by "terminating entity"? Are we talking about the entity at the terminus of the path, or at some terminus of the DAG? I think you mean path terminus but would you mind being clearer about that in the doc please?

Two problems to raise here:

1. If it's anything other than not the requested root, then we can't conform to this spec in Lassie without buffering, which I don't think we want to be doing, we need to get the CAR header out ASAP and not wait around until we've figured out `/where/this/path/ends/and/how/many/sharded/directories/we/have/to/navigate/through/and/collect/blocks/from`. I don't think we'd be the only party not keen to be buffering before we are able to produce a spec-compliant "trustless" CAR (similar issue, though not quite as bad, for the case of being able to validate a CAR in a streaming way).
2. Because of `<history>`, go-car is going to throw up errors in some forms of consuming (perhaps most, I'd have to trace these code paths in v2) zero-root CARs. I explicitly left the zero-roots possibility in the spec and in js-car, but we didn't iron this out of go-car, mainly because at the time it was a requirement of online deals in Filecoin where it was originally being used: https://github.com/ipld/go-car/blob/e09deeb6f2fded9370b8f5199123cb4894120996/car.go#L144-L146, https://github.com/ipld/go-car/blob/e09deeb6f2fded9370b8f5199123cb4894120996/v2/internal/carv1/car.go#L155-L157

lidel: > I don't think we'd be the only party not keen to be buffering before we are able to produce a spec-compliant "trustless" CAR [..]

Agree, at the end of the day `roots` value is limited to specific use cases, and in HTTP transport they do not matter, client knows what they've asked. Allowing empty `roots` may end up an important latency optimization for trustless clients like [ipfs in chromium](https://blog.ipfs.tech/2023-05-multigateway-chromium-client/), where miliseconds matter for happy eyeballs.

I've relaxed requirement in https://github.com/ipfs/specs/pull/402/commits/56786ff25ed31c027b07f38b6750df00421cdad7, `roots` is now optional,  and clarified terminus + added explicit note about impact on buffering. 

> I explicitly left the zero-roots possibility in the spec and in js-car, but we didn't iron this out of go-car

Agree, "empty car, no roots" should not be an error.

rvagg: Relevant to the zero-roots question: https://github.com/ipld/go-car/issues/26 - https://github.com/ipld/specs/pull/255/files; current suggestion for compact but safe header is just to use `bafkqaaa`. But I do feel like maybe it's time to just remove that len==1 thing entirely, maybe we could do it in the v2/ code paths at least.

hacdias: @lidel as @rvagg said, `go-car` screams if there aren't any roots. I wouldn't be opposed to changing it to allow reading rootless CARs. I just wonder if that's going to break anything within `go-car`. Perhaps we could go for the empty CID.

Note that conformance tests are blocked on this, since they'll fail if we try opening a rootless CAR.

hannahhoward: @lidel my first response is why would we specify anything about the behavior in roots in this spec.

hannahhoward: this should be a seperate IPIP if it's important.

hannahhoward: And I'm not of the view this is important. If I make a trustless request, and you send me a CAR file back that contains all the blocks needed to trustless verify said request, I do NOT feel that I should reject it cause of what is in the roots field of CARs.

I specifically think the Roots field in the current CAR spec is a foot gun cause its meaning is completely unclear. To the extent people are basing behavior off of it, that's not good.

hannahhoward: my first preference is leave this out of the spec, say we might define it later.

following that, I would say, entity is not good for streaming because you have to traverse the path blocks first to verify it

after that my preference would be EMPTY

after that my preference would be root of the whole dag

my least favorite is to use entity.

lidel: We are in a pickle with this one :) 

Implementers outside PL should have some rules of thumb written down, and should not guess, because they may guess and implement the thing we like the least.  

@rvagg  We could adjust spec and replace `the field MAY be left empty.` with ``the field MAY  include `bafkqaaa` CID to indicate that roots were not known before returning the CAR stream``. This way we always have a CID, will not error on existing/legacy software, and if such CAR is imported to future Kubo it will only pin that identity CID, will not trigger over-fetch of entire Wikipedia if I've only exported a single article + path. Solves multiple problems.

BUT, if we say it is ok to use fake `bafkqaaa` CID here, then why are we not ok with [using empty block or identity CID as tombstone for indicating the end of car stream](https://www.notion.so/pl-strflt/Streaming-CARs-bfa54196ed0740eea3c23dd70efbe07b#97f0454e9c094c98a76aeac6184cf3e7)? iirc Juan was against that, as it mixes data layer with transport layer semantics, and unsure how this is different -- I suspect he will be against this one too, for consistency, but maybe the pragmatism will prevail here?

hannahhoward: The biggest issue here is not one of our making: the roots field of the CAR has unresolved issues, documented in the spec. It's not clear what it's truly supposed to mean in relation to the blocks in the file. (i.e. for example, whether the Root CID even needs to represent an actual block in the collection that follows)

In practice, we have moved farther and farther down the road of treating the CAR as a packed/streaming block store. IMHO, this makes sense cause the CAR format to date doesn't actually contain enough information to treat it as a self-verifying serialization of a DAG traversal. I agree `bafkqaaa` is a weird hack. IMHO, we need to update go-car to handle empty roots. If that requires checking downstream for breaking changes, let's do that. I don't believe the surface area of code that is affected is actually that wide.

hannahhoward: Anyway, I like where you ended @lidel (MAY contain)

hannahhoward:
@lidel @rvagg 

I'd like to gain clarity about where we ended with the CAR Roots value.

As the spec currently stands, it contains 2 MAY blocks:

```
If the response uses version 1 or 2 of the CAR spec, the
[`CarV1Header.roots`](https://ipld.io/specs/transport/car/carv1/#header) field
MAY contain the CID of the terminus of the content path.

If implementation prefers to avoid buffering blocks, and return them as soon as
possible, the field MAY be left empty.
```

This essentially means you have two suggestions but all roots values are technically valid.

That means conformance tests should not be checking roots.

However, I am concerned @lidel intends for this to essentially be an either/or MUST -- either 1 CID with the terminus, or no CIDs at all.

Meanwhile, for streaming support Lassie is definitely not going to implement terminus in root, but at the same time @rvagg has shared with me that proper go-car support for empty roots is a non-trivial change he does not want to make at this time. It would seem unfair to force this change that might affect an install base of go-car users when as I understand it the desire for using terminus is to protect and install base of kubo users.

Also, both Lassie and .storage implementations are sending the actual root CID.

I think we need to come to a clear resolution ASAP. I'm fine with the current text as long as we all understand both lines are suggestions, and conformance tests do no checking. Otherwise, I think we need to discuss synchronously to come to a clear aggreement.

cc: @willscott @alanshaw (for .storage representation)

also, as always my personal position is we should not be specifying this in 402 at all, but I am fine with anything others can agree on.

lidel:
My understanding of the situation is that:
- for CAR roots
  - we've ended up with too many MAYs which make specification hard to test, or make tests meaningless
  - empty CAR `roots` is the wrong thing to recommend, as  Filecoin-specific validation in go-car will produce error
    - we have no bandwidth to clean this up / decouple GO implementation from Filecoin specific validation across the ecosystem at this time
- for CAR determinism in general
  - we have no byte-for-byte or block-for-block determinism guarantees anywhere in the CAR spec anyway
  - can't block this IPIP and implementations until CAR specs or library papercuts are finalized
     - we know the existing CAR is not enough, and instead we should solve these problems in future version
  - we've descoped determinism and moved some aspects into [IPIP-412](https://github.com/ipfs/specs/pull/412)

Due to this, I agree with @hannahhoward that the only path forward is descope  the roots header from this IPIP.

I pushed  61d108887d9f1564d9d651e44bb990c176eb0c30 which clearly states there is no such thing as deterministic CAR at the moment,  and removes any recommendations about CAR `roots` other than "Clients MAY ignore it."
We will remove `roots` validation from gateway-conformance later this week too.

This unblocks this IPIP, and enables future-proof clients.

src/http-gateways/trustless-gateway.md
lidel: Ended up removing this from the scope of this IPIP: https://github.com/ipfs/specs/pull/402#issuecomment-1599428849

hannahhoward:
Thanks @lidel. @rvagg I think we need to move forward with supporting no roots at all in go-car as soon as we can. Checks for roots should move up to the software that needs it. We should chat about how we can get this work prioritized as part of your stewarding of IPLD.

rvagg:
Hey @lidel & @hannahhoward, I've been poking around in CARs all day and realised that Kubo is currently using the go-car/v2/blockstore interface which happens to bypass the roots length check. So it's possible to import a rootless CAR in the currently released version of Kubo. So that does open up the door quite a bit to allow rootless CARs in this spec.

hannahhoward:
@lidel do we have anything left on this PR? What is the process required to merge? Seems like everything I'm aware of is resolved at least for the scope of this IPIP

lidel:
Yes, the scope is finalized, I don't expect functional changes here.

Next steps are a formality:

- [ ] apply final editorial changes 
  - [ ] resolving TODOs, such as filling fixture CIDs with ones from gateway-conformance – this is wip, last remaining CID will land in https://github.com/ipfs/gateway-conformance/pull/85, then I will update this PR
  - [ ] go over all remaining threads and resolve them, for things out of scope create issue (e.g. I will fill one for signaling HTTP OPTIONS)
- [ ] flag this as ready for final review at the next [IPFS Implementers Sync](https://pl-strflt.notion.site/IPFS-Implementers-Sync-f102a74802b34529a759ffbc3ea20303) (2023-06-29), give the wider IPFS community 2 weeks to provide final feedback
- [ ] ratify and merge

src/http-gateways/trustless-gateway.md
```diff
@@ -66,12 +74,137 @@ Below response types MUST to be supported:
...
+interpreted as a continuous array of bytes (such as a UnixFS file), returns
+only the minimal set of blocks required to verify the specified byte range of
+said entity.
+
+Allowed values for `from` and `to` are positive integers where `to` >= `from`, which
+limit the return blocks to needed to satisfy the range `[from,to]`:
+
+- `from` value gives the byte-offset of the first byte in a range.
+- `to` value gives the byte-offset of the last byte in the range; that is,
+the byte positions specified are inclusive.  Byte offsets start at zero.
```
rvagg, line 115: What's the behaviour for a `from` or a `to` that takes us outside of the bounds of the file (positive `from` or `to` that is beyond EOF and negative `to` that resolves to <0?

rvagg: What is the behaviour if `from` and `to resolve to a zero byte range? does that mean you get no blocks, or just the block that the byte that they both point to?

src/http-gateways/trustless-gateway.md
```diff
@@ -66,12 +74,137 @@ Below response types MUST to be supported:
...
+
+Optional, `entity-bytes=from:to` with the default value `0:*`, only available for CAR requests.
+Serves as a trustless form of an HTTP Range Request.
+
+When the terminating entity at the end of the specified content path can be
+interpreted as a continuous array of bytes (such as a UnixFS file), returns
+only the minimal set of blocks required to verify the specified byte range of
+said entity.
+
+Allowed values for `from` and `to` are positive integers where `to` >= `from`, which
```
rvagg, line 110: > where `to` >= `from`, which

Similar (same?) question as the one below about `from` and `to` resolving to a zero byte range. Because `=` is allowed here, what does that mean in the `=` case?

Separate question: what does it mean for them not to be "allowed"? We could reject obviously bogus queries with an error at the top end of the request (e.g. `entity-bytes=100:10` - error, bad range). But what about where you don't know the range falls within this "allowed" ruleset until you have the file details—this uncertainty is created by negative byte ranges which could mean `to` < `from` once you know how long the file is. Is this an error? Or do we massage values for the user?

src/http-gateways/trustless-gateway.md
```diff
@@ -66,12 +74,137 @@ Below response types MUST to be supported:
...
+
+If the entity at the end of the path cannot be interpreted as a continuous
+array of bytes (such as a DAG-CBOR/JSON map, or UnixFS directory), this
+parameter has no effect.
+
+The following additional values are supported:
+
+- `*` can be substituted for end-of-file
+  - `entity-bytes=0:*` is the entire file (a verifiable version of HTTP request for `Range: 0-`)
+- Negative numbers can be used for referring to bytes from the end of a file
```
rvagg, line 125: How critical are negative byte ranges to this spec? Do we have clear and obvious use cases for them? They add considerable complexity both to the spec (as per questions here) and implementation, is it worth the cost?

BigLep:
2023-06-29 Kubo maintainer conversation: there are extra case that @lidel will add about:
"error handling ro entity-bytes"
"CAR responses for sub-resource that does not exist"

src/http-gateways/trustless-gateway.md
lidel: We need this for gateway to be able to support real world HTTP Range Requests. 

We can ignore support for multiple ranges in a single request (not really used, more performant to send separate requests) and only support one range, but we do need to support negative ranges as there are valid use cases around

- append-only data structures, such as logs

src/http-gateways/trustless-gateway.md
lidel: Thank you for pointing these gaps around error handling.

The rule of thumb we try to follow on trustless gateways is empowering the client to be able to self-verify the response, and that includes error scenarios. If a request was done for an invalid range, the response should include enough blocks for the client to see that.

@rvagg  I've added clarification in  608e7b7  – lmk if it is good enough  or how we could make it better.

(Not feeling strongly about zero-range requests, it is an edge case with no impact on real world requests, I made it collapse to be the same as `dag-scope=block` for the sake of simplicity, but lmk if other approach will be easier.)

src/http-gateways/trustless-gateway.md
lidel: There is `=` because HTTP Ranges are inclusive (`0-0` means one byte, see [RFC9110#14.1.2](https://httpwg.org/specs/rfc9110.html#rfc.section.14.1.2)). The `entity-bytes` offsets follow the same notation to avoid translation and one-off errors.  I've added explicit reference to this RFC so its more clear how offsets work.

To create zero-range you would have to use negative values from the end, and match file length perfectly. Doable, but also something that requires knowledge of the DAG, thats why I think we should return `entity-dag

For `to < from` cases, 608e7b7 added:

> A Gateway SHOULD return an HTTP 400 Bad Request error when the requested range cannot be parsed as valid offset positions.

I think if you know upfront the range is bogus, and the client has all information to come up to the same conclusion themselves (to < from), it is ok to return error.  

But in more nuanced cases where client may not have enough knowledge of the DAG,  we need to return  at least the root block.

src/http-gateways/trustless-gateway.md
lidel: @rvagg the second half of this section that allows for positive `from` AND  negative `to` is optional tho: 

>    - It is also permissible (unlike with HTTP Range Requests) to ask for the
>      range of 500 bytes from the beginning of the file to 1000 bytes from the
>     end: `entity-bytes=499:-1000`

This will never be used in the context of HTTP Range Requests, so we could potentially remove it without impacting `ipfs.io` and Brave.
Would removing this simplify anything on your end?

rvagg: Not really, it's once we have negatives at all that we enter tricky territory. Not the end of the world, I'm iterating toward solving it but it does open up some tricky paths. `499:-1000` is not really much harder than `-2000:-1000` I think.

src/http-gateways/trustless-gateway.md
rvagg: OK, so what I'm working on right now gets to the question of bad ranges.

* Selector spec update: https://github.com/ipld/ipld/pull/289
* Selector implementation for negatives: https://github.com/ipld/go-ipld-prime/pull/530

What I'm going with is this: if we get to the range match, and then perform whatever calculations need to be made to adjust `To` and `From` and we find that it specifies a range entirely outside of the bytes we have available, then it's not going to match anything. This means that you'll get the root block and nothing else. But you also won't get an error condition (no dirty stream termination)—although you could figure that out yourself with the data you get back.

Some edge conditions:

* `To` beyond the end of the file is fine (and I need to use this fact to convert `[From,To]` in this spec to `[From,To)` in the impl.)
* `From` before the start of the file is not fine and won't match anything (although perhaps it could be? Maybe it doesn't matter if you end up matching bytes anyway. I might change this but input welcome!)
* `To`<`From` is not fine and won't match anything. If you provide positive values and `To`<`From` at the beginning then we can `400 Bad Request` on that. But we can't do that if you provide negatives, just no blocks beyond the root.

src/ipips/ipip-0402.md
```diff
@@ -235,9 +230,14 @@ single request.
 
 ## Test fixtures
```
lidel, line 231: :information_source:  A detailed list of compliance checks for `dag-scope` and `entity-bytes` can be found in  [`gateway-conformance v0.2.0 → trustless_gateway_car_test.go`](https://github.com/ipfs/gateway-conformance/blob/v0.2.0/tests/trustless_gateway_car_test.go). I've made another pass over them, and pushed https://github.com/ipfs/specs/pull/402/commits/4ebbb9642439abf2b97ebc91e53c79f9d979c7b4  which adds CIDs, CARs, and short summary of each fixture to this IPIP.

src/http-gateways/trustless-gateway.md
rvagg: Update: I'm going to be more permissive with the `From` value underflowing. So the rules end up being like this:

* Incoming request `From>0 && To>0 && From>To` - `400 Bad Request`
* After resolving negative `From` and/or `To` values once we have the file data:
  - Where `From<0`, `From` is adjusted to `0`
  - Where `To>=Length`, `To` is adjusted to `Length-1`
  - Where `From>To`, nothing is returned.
  - Where `To<0`, nothing is returned.
  - Where `From>=Length`, nothing is returned.
  - Where `From==To`, the block at that byte is returned.
  - Normal rules apply for other ranges.

(Although these rules are mediated through a `[From,To]` -> `[From,To)` translation layer which is slightly awkward)

hannahhoward: @rvagg it's worth clarifying to those reading the above that the rule set described is applied AFTER any negative indexes are converted to their equivalent index from the start of the bytes.

src/http-gateways/trustless-gateway.md
hannahhoward: @lidel yea this came up already as something we'd like to have when were adding support for range requests to Lassie -- we need to be able to figure out if upstream providers support them (web3.storage currently does NOT).

hannahhoward: Not sure OPTIONS will help with the immediate problem above, but just saying this sort of "what features do you support?" is pretty useful and probably will remain so.

src/http-gateways/trustless-gateway.md
lidel: @rvagg @hannahhoward  by "nothing is returned"  you mean truly empty car (useless to the client -- unable to tell what went wrong)  or equivalent of `dag-scope=block` (path+root, allowing client to understand what went wrong with their request)?

The spec states the latter – any concerns around that?

> In more nuanced error scenarios, a Gateway MUST return a valid CAR response that includes enough blocks for the client to understand why the requested  `entity-bytes` was incorrect or why only a part of the requested byte range was
returned:
> [..]
> If the requested `entity-bytes` resolves to a zero-length range or falls   fully outside of the entity's bytes, the response is equivalent to   `dag-scope=block`.

BigLep:
2023-07-06 Kubo maintainer conversation: we need to ensure a conformance test involving inline CIDs.

src/http-gateways/trustless-gateway.md
lidel: @hannahhoward thank  you for flagging the need at hand. 
Extracted HTTP OPTIONS discussion to a separate IPIP https://github.com/ipfs/specs/pull/425 – see https://github.com/ipfs/specs/pull/425#discussion_r1254881047 for range request example.

src/http-gateways/trustless-gateway.md
rvagg: by "nothing is returned" I mean none of the child blocks for a sharded file - you get everything up to the parent block of the sharded file and nothing beyond that; so still a valid CAR, and technically you could validate for yourself why it failed by rerunning the byte range calculation locally, so this seems fine I think?

src/http-gateways/trustless-gateway.md
```diff
@@ -67,12 +75,162 @@ Below response types SHOULD to be supported:
...
+
+A Gateway MUST augment the returned `Etag` based on the passed `entity-bytes`.
+
+A Gateway SHOULD return an HTTP 400 Bad Request error when the requested range
+cannot be parsed as valid offset positions.
+
+In more nuanced error scenarios, a Gateway MUST return a valid CAR response
+that includes enough blocks for the client to understand why the requested
+`entity-bytes` was incorrect or why only a part of the requested byte range was
+returned:
```
Jorropo, line 146: This does not specify what error code should be returned, for example if I try to load path `/c` in a directory that only contains `a` and `b` it must be acceptable that 200 is a valid code because a heavily streaming implementation do not know if the file resolves or not while sending the first block, it is possible to do the resolution while sending the blocks for the resolution in the meanwhile.

lidel (approved):
We have conformance tests, reference implementation is in `boxo/gateway` and shipped with Kubo 0.21 and the same API been used in Rhea/Saturn for a while.

Thank you to everyone involved, especially asking questions and pointing gaps – it all made the specification better, and also produced [IPIP-412: ordered cars](https://github.com/ipfs/specs/pull/412) and [IPIP-425: signaling features](https://github.com/ipfs/specs/pull/425) where we continue some of the work that started here.

I believe this is ready for being rattified.
Will flag this during [IPFS-Implementers-Sync-2023-07-20](https://www.notion.so/pl-strflt/IPFS-Implementers-Sync-2023-07-20-d4a7745296cb45028d73e77474df007e?pvs=4).

src/http-gateways/trustless-gateway.md
```diff
@@ -67,12 +75,174 @@ Below response types SHOULD to be supported:
...
+
+When the terminating entity at the end of the specified content path:
+
+- can be interpreted as a continuous array of bytes (such as a UnixFS file), a
+  Gateway MUST return only the minimal set of blocks necessary to verify the
+  specified byte range of that entity.
+
+  - When dealing with a sharded UnixFS file (`dag-pb`, `0x70`) and a non-zero
+  `from` value, the UnixFS values `filesize` and `blocksizes` determine the
+  corresponding starting block for a given `from` offset.
```
Jorropo, line 117: `filesize` indicates nothing about ranging and must absolutely not be used.

lidel: @Jorropo what would be the correct phrasing here? just `blocksizes`?
(this is until we have UnixFS specs landed and can point to that):
```suggestion
  - When dealing with a sharded UnixFS file (`dag-pb`, `0x70`) and a non-zero
  `from` value, the UnixFS  `blocksizes`  determine the
  corresponding starting block for a given `from` offset.
```

Jorropo: it's `blocksizes` + `data`, but I feel this is too pedentic, your latest suggestion is ok while we unixfs spec is still not merged.

lidel:
There were no ratification concerns during IPFS Implementers sync, nor during the week after it. 
Merging to unblock IPIP-412 which is built on top of this one.

For future reference, implementation of IPIP-402 lives in [`boxo/gateway`](https://github.com/ipfs/boxo/tree/main/gateway) library and shipped for the first time with [Kubo 0.21](https://github.com/ipfs/kubo/blob/v0.21.0/docs/changelogs/v0.21.md#gateway-support-for-partial-car-export-parameters-ipip-402).

PR #407, opened 2023-04-25
fix(ipns): use PascalCase for IpnsEntry.data (CBOR) fields

John-LittleBearLabs:
For preceeding discussion:
https://github.com/ipfs/specs/issues/406#issuecomment-1520901854
(and before that, Slack).


lidel (approved):
Thank you!

PR #408, opened 2023-05-08
feat: use frontmatter title and description, automatic lists

hacdias:
Required before #385.

For this I had to make a small change in spec-generator, which I released: https://github.com/ipfs/spec-generator/pull/13.

It uses `title` and `description` in frontmatter for rendering. Also description included in meta tags.

Automatic listings thanks to the `tags` feature of Eleventy. Ordering within each tag is based on the date field. Couldn't find much information on how to sort by a different field, such as title. For that, we probably have to write our own code, and potentially hard code more things.

Also updated the template because it seems we did forget before.


lidel (changes requested):
I think automatic lists make sense for things like IPIPs, but for other specs we have we need a mechanism for adjusting the ordering.

Some specs are more meaningful than others. For example, we want:
- IPNS Record specs before we introduuce random routers like PubSub. 
- Path Gateway and Trustless Gateway before web ones, additions like _redirects at the very end.


| before (meaningful order) | now (not the best) | 
| ---- | ---- |
|   |  |


@hacdias  thoughts on potential fixes? my initial suggestion would be to do (B) below:
- (A) use automated list for IPIPs only, other things remain manual (but we leverage title and description from front matter)
- (B) add `hierarchy` value to the front matter, and then create filter to `sortByHierarchy`  similar to [this](https://stevenwoodson.com/blog/a-step-by-step-guide-to-sorting-eleventy-global-data-files-by-date/#manual-date-sorting-by-data-property) that adjust order based on the number if `hierarchy` field is present  
   - (0 is the highest, 999 is the lowest, 999 is the implicit default if `hierarchy` is not present in frontmatter)
   - this will allow us to distinguish between top level specs like IPNS Record and Path Gateway, and always put the at the top

hacdias:
B seems the best solution.

lidel:
@hacdias  maybe call it `order` ?  the nice side-effect will be we could put IPIP number there too, and we get deterministic ordering  from latest to oldest  (` | sortByOrder | reversed`) :)

lidel:
@hacdias mind updating first paragraph of http://localhost:8642/meta/spec-for-specs/#title-sections to say the title should be in front matter?

PR #412, opened 2023-05-15
IPIP: Signaling Block Order in CARs on Gateways

lidel:
> This PR supersedes https://github.com/ipfs/specs/pull/330 and continues order discussions from #402
 
cc @hannahhoward @aschmahmann @Jorropo @willscott @rvagg @ribasushi @John-LittleBearLabs @fabricedesre

## What

This  IPIP is based on prior art from [IPIP-330](https://github.com/ipfs/specs/pull/330), and recent discussions cited in the header front matter, including ones in [IPIP-402](https://github.com/ipfs/specs/pull/402).

## Why

:point_right:  To keep scope in check, I made a conscious decision to only specify a single opt-in ordering that aims to reduce resource usage on light clients like [IPFS in Chromium](https://github.com/little-bear-labs/ipfs-chromium/), [Capyloon](https://github.com/capyloon), [RAPIDE](https://www.youtube.com/watch?v=Cv01ePa0G58), or even  [bifrost-gateway](https://github.com/ipfs/bifrost-gateway). 

These are the north star use cases  for now.

## TLDR How

The proposed solution introduces two new parameters for the content type headers
in HTTP requests and responses: `order` and `dups`.

- The `order` parameter allows the client to indicate its preference for a
specific block order in the CAR response, for now we only specify the [Depth-First Search](https://en.wikipedia.org/wiki/Depth-first_search) order.
- The `dups` parameter specifies
whether duplicate blocks are allowed in the response

The proposed spec creates room for other ones to be added in future IPIPs.

cc @hannahhoward @aschmahmann @Jorropo @willscott @rvagg @ribasushi


src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+`order=dfs` with `dups=y` instead (lower priority indicated via `q` parameter,
+as noted in :cite[rfc9110]).
+
+#### `order` CAR content type parameter
+
+The `order` parameter accepts the following values:
+
+- `dfs`: [Depth-First Search](https://en.wikipedia.org/wiki/Depth-first_search)
+  order, allows for streaming responses with minimal memory usage
+- `rnd`: Unknown (random) order, the implicit default when `order` parameter is missing.
```
Jorropo, line 92:
```suggestion
- `dfs`: [Depth-First Search](https://en.wikipedia.org/wiki/Depth-first_search)
  order, allows for streaming responses with minimal memory usage
- `weak`: weak maintains a seen set, the seen set is initialized with the CID passed in, the gateway is only allowed to send blocks that are in the seen set. When a block is sent it's childrens that are relevent to the content of the request are added to the seen set, if `dups` is `n` the block that was sent is removed from the seen set.
  In other words this provide vertical ordering but not horizontal ordering.
  This ensures the gateway never send a block the client can't know about yet and ensure the stream is incrementally verifiable while providing a weaker ordering requirement which allows the gateway implementation more creative potentially faster strategies.
- `rnd`: Unknown (random) order, the implicit default when `order` parameter is missing.
```

John-LittleBearLabs: `rnd` doesn't mean random, correct, but just arbitrary/unspecified?
I assume we're not requiring a gateway to shuffle things by default.

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+### Signaling in Response
+
+The Trustless Gateway MUST always respond with a `Content-Type` header that includes
+information about all supported/known parameters, even if the client did not
+specify them in the request.
+
+The `Content-Type` header format is as follows:
+
+```
+Content-Type: application/vnd.ipld.car;version=1;order=dfs;dups=y
```
lidel, line 120: @hannahhoward  to unblock work in service worker, biforst-gateway/boxo and gateway-conformance before this IPIP is ratified, is this acceptable default to implement in Lassie / Saturn L1?  

Your implementation is free to be more strict and only support one set of parameters.
Namely, the ask here would be for Lassie/Saturn L1 to return blocks in Depth-First Search order & sending duplicates.

src/ipips/ipip-0412.md
aschmahmann: I think something like `unknown` is better than `rnd` as a name here. I'm also curious about the interest in support for `unknown/rnd` given that it's not incrementally verifiable and therefore necessitates both added trust and added (potentially unbounded) buffering.

@Jorropo IIUC what you are describing with `weak` is more succinctly "any valid graph traversal order", since the types of orders you're trying to exclude are ones like putting all the leaves first and the root at the end. Is that right?

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+
+#### `dups` CAR content type parameter
+
+The `dups` parameter specifies whether duplicate blocks (the same block
+occuring multiple times in the requested DAG) will be present in the CAR
+response.
+
+It accepts two values:
+- `y`: duplicate blocks are allowed
+- `n`: duplicates are not allowed
```
fabricedesre, line 102: What is the default value when this parameter is not specified?

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+The `Content-Type` header format is as follows:
+
+```
+Content-Type: application/vnd.ipld.car;version=1;order=dfs;dups=y
+```
+
+
+Gateway implementations are free to decide on the implicit default ordering or
+other parameters, and use it in responses when client did not explicitly
+specify, or requested unsupported or unknown query parameter.
```
fabricedesre, line 126: I would rather have all default values defined to make interop easier. Also, return a 4xx response when receiving an unknown value or unknown parameter.

src/ipips/ipip-0412.md
rvagg: -1 to adding additional items to this list in the first pass, the more overengineered this spec is in the first instance the less likely it is to be properly, or fully implemented in >1, or even >0 places. This stuff is so far from nontrivial and the cross-language barriers are huge and we haven't even crossed that barrier for simple depth-first yet.

hannahhoward: and also let's keep it simple like @rvagg says in first pass. We can add more later.

src/ipips/ipip-0412.md
hannahhoward: @lidel you guys need duplicates?

rvagg:
Does identity CID handling belong in here or #402? It feels very closely related to duplicate blocks so might be in scope to be described here.

src/ipips/ipip-0412.md
Jorropo: @rvagg @hannahhoward we already have implementations that makes download faster if using `weak` instead of `dfs` in `ipfs pin add` and RAPIDE.
I don't have the unixfs decoder since we havn't deployed thoses on gateway yet, this is about an afternoon of work (have a map that keeps tracks of all open ranges in the file, when a new block is downloaded progress open spaces and use `.WriteAt` to write to the correct offset).

`dfs` is already a valid implementation of `weak` so this is not represent any real cost for implementations since we already want to support `dfs`. Breath first search is also a valid implementation, anything that sends parents before it sends children is a valid implementation.

hannahhoward: Nothing prevents us from adding `weak` later. But any "arbitrarily valid traversal" order feels ill defined for me to be going in a formal IPFS spec, at this point, in order to support prototype clients.

I'd like to request additional discussion and a seperate spec, similar to how this spec evolved from 402.

src/ipips/ipip-0412.md
hannahhoward: we specifically DO NOT send duplicates for the time being, it's non-trivial to throw them back in.

hannahhoward: cc: @aschmahmann 

Basically, current Lassie is:
Content-Type: application/vnd.ipld.car;version=1;order=dfs;dups=n

For "dups=y" and maybe this is not clear from the spec, are dups simply allowed, or required? (that is worth putting in the spec in fact)

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+- `rnd`: Unknown (random) order, the implicit default when `order` parameter is missing.
+
+#### `dups` CAR content type parameter
+
+The `dups` parameter specifies whether duplicate blocks (the same block
+occuring multiple times in the requested DAG) will be present in the CAR
+response.
+
+It accepts two values:
+- `y`: duplicate blocks are allowed
```
hannahhoward, line 101: this should specify if duplicate blocks are allowed or required. IOW, is a server response that conforms to dups=`n` still valid for dups=`y` cause a client can still verify based on previous blocks?

src/ipips/ipip-0412.md
aschmahmann: @hannahhoward are you only objecting to `weak` or also `unknown/rnd`?

src/ipips/ipip-0412.md
aschmahmann: IMO the point of duplicates is to enable people to write clients that stream results and do not need to keep blocks buffered.  If people prefer a different name/nomenclature that's fine, but the end result of `Content-Type: application/vnd.ipld.car;version=1;order=dfs;dups=y` should be that you are doing a preorder DFS traversal where even if the same block is encountered multiple times in the traversal it is still emitted once per time it is needed as part of the traversal.

For example, consider having go-ipld-prime be hooked up to a traversal and an inbound car such that:
1. If you ask it for a block other than the next block it has available it errors
2. It can throw away blocks after the first time they're used and the traversal will still complete

This should be usable to verifiably stream out all of the file bytes of a UnixFS file with only a single network request and a small fixed amount of memory (e.g. a few MiBs).

aschmahmann: > it's non-trivial to throw them back in.

This is somewhat off-topic/implementation specific so if there's another place you'd prefer to engage lmk, but why is this the case?

IIUC this should be fairly straight-forward to implement given that go-ipld-prime traversals do not buffer data in memory and the strategy you take towards populating your CAR files is to do a traversal and on block-load get the block. If the go-car functions are forcing deduplication on you and not giving you an alternative way out that seems like a fixable problem.

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,240 @@
...
+- `n`: duplicates are not allowed
+
+When allowed (`y`), light clients are able to  discard blocks after
+reading them, removing the need for caching in-memory or on-disk.
+
+<!-- TODO: do we need a  parameter for inclusion of identity CIDs?
+It seems to be only relevant in Filecoin due to legacy hiccup:
+https://github.com/ipfs/specs/pull/330#issuecomment-1274106892 -->
```
aschmahmann, line 109: @rvagg seems at least @lidel thinks it's fair game to discuss here https://github.com/ipfs/specs/pull/412#issuecomment-1548773588.

Do you know of anyone that wants/needs this for exporting CARs using this API? I'm not personally aware of anyone and given that IIUC the trustless gateway components are being used mostly as data transports it doesn't seem like anyone should need identity CIDs in here. If no one needs this we should probably clarify that they will not be included.

src/ipips/ipip-0412.md
hannahhoward: just weak. I don't care about parameter names. 

Nothing prevents someone from passing `weak` any more than the range request standard allows you specify other values besides `bytes`. I'm not saying it's a bad idea or a useless ordering. I'm just saying from a standards point of view, we should be putting in values that have deployment in prod and an existing product using it. such that the standard is just bringing things in line with how production is already working, and potentially enabling others to use it. we shouldn't generally design standards around possible use cases. my argument is if @Jorropo wants to keep working on `weak` and develop some deployments, and propose it's a recognized value for others to use, go for it. But right now we're aiming for the minimal standard to get Rhea over the line.

src/ipips/ipip-0412.md
hannahhoward: Yes I agree it should be non-trivial.  For reasons, ATM, it's not.

Do you need `dups=y` on your side for Rhea? I can talk to @rvagg if so.

src/ipips/ipip-0412.md
olizilla: It's true! But we can change if needed.

src/ipips/ipip-0412.md
lidel: Agree, "allowed"  should be replaced with stronger  "required" / "MUST" here + what is default if missing.
Clarified in e216ea91460a5f070302e2ebbf2f8bd419aa739b

src/ipips/ipip-0412.md
lidel: Clarified in e216ea9

src/ipips/ipip-0412.md
rvagg: If we can strictly rule it out, at least in the first pass of this thing, then that'd be awesome. They cause so much pain and only made it into the original Filecoin CommP generation "spec" (really just two sides needing to agree and it being hard to signal a change after it was deployed) because we forgot to consider them, then spent many hours banging heads against walls making them properly handled.

They're unnecessary and wasteful, and a potential abuse vector too. Having it specifically spelled out in the spec would be great to make sure implementers all consider them.

src/ipips/ipip-0412.md
rvagg: https://github.com/filecoin-project/lassie/issues/239#issuecomment-1550736498

Would be good to know if we need to support this for the *->saturn->lassie flow or just in the lassie->provider flow. As I said in the above comment I think it'd be useful option to have for data coming out of lassie for streaming cases (to avoid [this](https://github.com/ipld/go-car/blob/92d28eb23ba496288e309dc9672edf13cf9329c8/cmd/car/extract.go#L378-L444)) but there'd be some work to do to make it happen, hopefully mainly on the testing side.

But, we should factor in the dangers of allowing this, we open up that space where deeply interlinked graphs can explode both in traversal recursion and, with this option, output data. You could probably make a retriever turn a few Mb of source data into Gb. Nice attack amplification potential with that.

src/ipips/ipip-0412.md
lidel: - I've renamed `rnd` to `unk` (unknown) and clarified purpose in 50223b296708b90fc610fb3f55387b0d8117d9d2
  - We don't expect to see `unk`—it is there to document the implicit behavior when `order` is not explicitly stated
- Agree we want to keep the scope of this IPIP surgical and limited to what we already have and use in the ecosystem (`unk` and `dfs`)
- @Jorropo feel free to propose and discuss more orderings in separate IPIPs  -- `weak` is a good candidate, but needs working code first AND fixtures that are different than `dfs`, and we don't want to slow down this IPIP by the details of fleshing it out.

src/ipips/ipip-0412.md
lidel: > For "dups=y" and maybe this is not clear from the spec, are dups simply allowed, or required?

@hannahhoward  updated this section to clarify that value returned in `dups` means required behavior (MUST).
(goal of this IPIP is to introduce opt-in signaling mechanism about how deterministic the block order in the response is, so the client can apply optimizations)

> Would be good to know if we need to support this for the *->saturn->lassie flow or just in the lassie->provider flow. 

@rvagg  imo Saturn needs it on the  client→saturn→lassie flow  because we want to have more and more light clients[^1] to hit  Saturn L1 directly, and `dups=y` will make it possible to process response cheaper/faster by removing the extra storage steps similar to the one you linked. 


[^1]: current example is  Saturn's own [browser clients](https://github.com/filecoin-saturn/browser-client), but in a few months this could include [native IPFS in Chromium based on the Trustless Gateway spec](https://github.com/ipfs/ipfs-blog/pull/560) in places like Brave, Electron etc.


> You could probably make a retriever turn a few Mb of source data into Gb. Nice attack amplification potential with that.

@rvagg  I need some help unpacking this. My understanding is that it is already possible to craft a huge blob from well-available data, and then request it from any gateway to waste resources, does not even need `?format=car`.

The difference with `dups=y` is that a retrieval client like Lassie would not pay the cost of retrieving all the blocks, 
and instead return a single one multiple times. 

The end result is the same (wasted resources, CPU/bandwidth capped) but Lassie did not waste resources of other peers to retrieve the payload.

 Is there a version of this where more resources are wasted than it is already possible?

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,252 @@
...
+
+We want to make it easier to build light-clients for IPFS. We want them to have
+low memory footprints on arbitrary sized files. The main pain point preventing
+this is the fact that CAR ordering isn't specified.
+
+This require to keeping some kind of reference either on disk, or in memory to
+previously seen blocks for two reasons.
+
+1. Blocks can arrive out of order, meaning when a block is consumed (data is
+   red and returned to the consumer) and when it's received might not match.
```
olizilla, line 36:
```suggestion
   read and returned to the consumer) and when it's received might not match.
```

olizilla (commented):
From a clients point of view:
- `order=dfs; dups=y` means _"make it as easy as possible for me to unpack"_
- `order=dfs; dups=n` means _"i will track blocks but verifiably get me to the leaf blocks asap!"_

Are there other combos that make sense for a client to set?

Could we encode this as a single param to avoid having combinations tha don't make sense. Something like `pack=dfs` implying _plz dedupe if you can_ and `pack=dfs-stream` implying _dupes must to be included_?

For the server-side, we may struggle to guarantee full dedupe in freeway. A CF worker gets 128MB of ram shared between concurrent requests from that region. In theory we could track ~2 million CIDs, but that number will drop as it gets busy. Stable ordering will eat some more ram too.

src/ipips/ipip-0412.md
lidel: I'd like  to avoid them as well -- added sentence that clearly states they are never welcome (80c09d050b4776da4791829fb307de0a7b553bf4), lmk if we need more/less.

olizilla: Adding a checks to not include identity CIDs should be painless.

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,260 @@
...
+We reuse exiting HTTP content type negotiation, and the CAR content type, which
+already had the optional `version` parameter.
+
+### User benefit
+
+The proposed specification change brings several benefits to end users:
+
+1. Improved Performance: Gateways can decide on their implicit default ordering
+   and cache CAR responses for popular content. In turn, clients can benefit
+   from strong `Etag` in ordered (deterministic) responses. This reduces the
```
aschmahmann, line 170: Is a strong `Etag` something we can (and/or want to) support here? There are still some areas that are unspecified even for responses where the blocks and their order are consistent (e.g. `order=dfs&dups=y`).

See https://github.com/ipfs/specs/pull/402#discussion_r1174223244.

The main ones remaining are:
1. CAR version (we can make v1 part of the etag, but then it'll need to change if people try to change the CARv1 format to support EOF without changing the name of the format)
2. What goes in the header of the CARv1?
   - Maybe I've just missed this, is it just the root CID (i.e. before any pathing occurs)?

Jorropo:
> Could we encode this as a single param to avoid having combinations tha don't make sense. Something like `pack=dfs` implying plz dedupe if you can and `pack=dfs-stream` implying dupes must to be included?

I was already proposing to join dupness and ordering in the previous IPIP but it was apparently really confusing.

> Are there other combos that make sense for a client to set?

- `order=weak, dups=n`

Allows the gateway to use stream different parallel downloads and the client can still incrementally and stream the data if you are writing it to an `io.ReadWriterAt`.

(this is being moved to an other IPIP)

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,260 @@
...
+
+#### `dups` CAR content type parameter
+
+The `dups` parameter specifies whether duplicate blocks (the same block
+occuring multiple times in the requested DAG) will be present in the CAR
+response. Useful when a deterministic block order is used.
+
+It accepts two values:
+- `y`: Duplicate blocks MUST be sent every time they occur during the DAG walk.
+- `n`: Duplicate blocks MUST be sent only once.
```
aschmahmann, line 106: Using a thread because IMO makes things a little easier to manage since there might be some opinions here 🙃 , but think @olizilla has some good points here https://github.com/ipfs/specs/pull/412#pullrequestreview-1430979342.

`dups=n` with a `MUST` can be pretty expensive to implement with certain scenarios (as can `dups=y`, although for UnixFS data it's not really more expensive than implementing the trusted gateway equivalent behavior).

Are the people who need `dups=n` in need of a MUST here, or is something like returning at most the duplicate blocks from `y` ok? This does make life a little harder for clients who would otherwise just be writing the data to disk as they now have to do a little more work and guessing to determine if getting data from someone who effectively sends them `dups=y` labelled as `dups=n` is worth dealing with in the future.

From my perspective no strong feelings on multiple parameters vs one name for all the things. If people really wanted strong etags and the ability to export CARs in a very particular way (e.g. matching how certain Filecoin imports used to work) then having names for the combination of parameters seems like a good idea. IIUC that's not the case though so 🤷.

src/ipips/ipip-0412.md
alanshaw: If I'm understanding correctly, we have something similar to @Jorropo's "weak" ordering in our [dagula](https://github.com/web3-storage/dagula) library for exporting DAGs:

It just fetches blocks and sends them as soon as it knows about them. The interesting part is that you'll not receive a block you don't already know that you will need, so it's verifiable in that sense, but they might not arrive quite in the order you need them.

It makes the server side very fast and memory efficient because you don't keep any state or buffer and don't need to recurse. The trade off is that the _receipient_ has to re-order and buffer any out of order blocks.

I believe implementing depth first ordering would make that export slower and/or more memory intensive...so it might be nice to have something inbetween `dfs` and `rdn` that is verifiable _and_ fast.

In our case we have memory constrained workers running in the cloud that are serving blocks to clients that probably have significantly more memory/ability to buffer than the worker does. They also probably want to receive data very quickly (who doesn't right?).

Another factor to consider is that our workers fetch blocks from CAR files stored in a bucket, with RTTs that are significantly greater than an SSD. Fetching blocks 1 by 1 to order as DFS (without blowing out memory) is going to be really slow.

So I guess this comment is to +1 on the "weak" ordering, and to say there _is_ a prod implementation of this (we use Dagula in our gateway [Freeway](https://github.com/web3-storage/freeway)).

src/ipips/ipip-0412.md
lidel: The meaning behind the presence of `dups=y|n` is MUST: you signal need or that you returned a deterministic thing.
that is a feature, because content-type parameters are optional,  so if you  dont know, dont care, or don't want to specify, you don't include it.

Order and duplicates are different properties, but clumping various things like thaty together under `pack` concept proposed by @olizilla (`pack=dfs` and `pack=dfs-stream`) makes it easier to test and reason about.

So we would have:
- `order=dfs` (unknown `dupes`) →  `pack=dfs`
- `order=dfs; dupes=y` → `pack=dfs-stream`

and we would have to only care about testing these specific bundles + easier to add future `pack` strategies without balooning parameter space.

cc @hannahhoward @fabricedesre thoughts?

rvagg: I've been thinking a bit about this the last few days as I'm working on some implementation things and I think what I'd like is something like this - being able to _request_ that a particular order be used, but then being _told_ what order is being used in the response. I'd like to be able to default to the more network-efficient `dups=n` but choose to switch to `dups=y` where a user doesn't want to burn temporary and/or wants to cleanly stream; but knowing that there are cases where `dups=n` isn't going to be practical (eipfs notably), I'd like to be told that by the server in order to make decisions on the client end - I can proceed and switch modes internally, or maybe bail because it doesn't fit within client constraints.

The same could be said for `order`. Lassie will always (currently) request `order=dfs` but if we get back an `order=something else` then we could either abort early, or choose to handle it in some other way (maybe the client trusts the provider explicitly and doesn't need the implicit guarantees of trustless dfs traversal so random is fine).

Could we add these to the `Content-Type` response header? `application/vnd.ipld.car; version=1; dups=y; order=dfs`.

rvagg: Also worth noting here that this is possible in CARv2, we could choose to go that route, it's slightly more annoying though. `application/vnd.ipld.car; version=2`, then use an indexless carv2 with characteristic bits set for order and dups.

lidel: @rvagg  yes, this is ok. Including `version` in response's `Content-Type`  is already happening in kubo/boxo/gateway/production sinc last year.

This IPIP follows that convention, and states we want gateway to return `dups` and `order` too (if these are known, even when not explicitly asked for):

> ### Signaling in Response
> 
> The Trustless Gateway MUST always respond with a `Content-Type` header that includes
> information about all supported/known parameters, even if the client did not
> specify them in the request.
> 
> The `Content-Type` header format is as follows:
> 
> ```
> Content-Type: application/vnd.ipld.car;version=1;order=dfs;dups=y
> ```

lidel: @rvagg so is the remaining ask here to 

- [ ] add an explicit spec note that "Gateway MAY choose to return different order/dups than the requested one instead of just erroring"

? 

It makes sense imo, because that is what will effectively be already happening in production with legacy gateways that do not support IPIP-412.

src/ipips/ipip-0412.md
lidel: Let's keep the scope of this IPIP limited to opt-in `order=dfs` and `dups=y|n`.

After this IPIP lands, we can fill a new one to add specification for `weak` or any other ordering.

For now, if you can't support `order=dfs`, or `dups`, just don't set these parameters on CAR responses.

src/ipips/ipip-0412.md
lidel: HTTP OPTIONS is tracked in https://github.com/ipfs/specs/pull/425

src/ipips/ipip-0412.md
```diff
@@ -0,0 +1,208 @@
...
+<!-- TODO after conformance release is tagged
+A detailed list of compliance checks for `order` and `dupes` can be found in
+[`v0.TODO.0/trustless_gateway_car_test.go`](https://github.com/ipfs/gateway-conformance/blob/v0.TODO.0/tests/trustless_gateway_car_test.go) or later.
+-->
+
+Below are CIDs, CARs, and short summary of each fixture.
+
+TODO:
+1. a CAR with blocks for a small file in DFS order
+2. a CAR with blocks for a small file with one block appearing twice
```
lidel, line 198: - [ ] This needs to be updated after https://github.com/ipfs/gateway-conformance/issues/77 is merged.

Not doing this yet to make sure IPIP includes the same CIDs as conformance test suite.

lidel (commented):
We have conformance tests (https://github.com/ipfs/gateway-conformance/pull/87), reference `order=dfs` implementation is in `boxo/gateway` (https://github.com/ipfs/boxo/pull/370) and plan to ship this with Kubo 0.22 (https://github.com/ipfs/kubo/pull/9989).

Thank you to everyone involved, especially asking questions and pointing gaps – it all made the specification better.

I believe this is ready for being ratified and merged after final fixture CIDs are known (comment below)

Will flag this during [IPFS-Implementers-Sync-2023-07-20](https://www.notion.so/pl-strflt/IPFS-Implementers-Sync-2023-07-20-d4a7745296cb45028d73e77474df007e?pvs=4).

lidel (commented):
There were no ratification concerns during IPFS Implementers sync, nor during the two weeks after it. 
Merging! :tada:

For future reference:
- CAR fixtures listed here are part of [ipfs/gateway-conformance](https://github.com/ipfs/gateway-conformance), consider using  that for compliance testing before rolling your own tests.
-  implementation of IPIP-412 lives in [`boxo/gateway`](https://github.com/ipfs/boxo/tree/main/gateway) library and shipped for the first time with [Kubo 0.22.0-rc1](https://github.com/ipfs/kubo/blob/v0.22.0-rc1/docs/changelogs/v0.22.md#gateway-support-for-order-and-dups-parameters-ipip-412).


Thank you to everyone involved in this one :heart:

Issue #363, opened 2023-01-11
"Other IPFS Overviews" link in README is 404ing 


ElPaisano:
On https://github.com/ipfs/specs#index, under the "IPFS Protocol" bullet point, https://github.com/ipfs/specs/blob/main/overviews is 404ing.

The markdown shows that the link is `[Other IPFS Overviews](/overviews)`. I don't see `/overviews` in the repo, so not sure where this supposed to be linking to.

Mr0grog:
Looks like it was removed in early 2020 with the goal of pushing people to the docs site instead: #243. Someone should either remove the bullet or tweak the text and link to https://docs.ipfs.tech/.

PR #413, opened 2023-05-15
fix: remove broken link and update README

Tikaryan:
- Fixes: [Issue-363](https://github.com/ipfs/specs/issues/363)
- Removed the `IPFS Overviews` broken link.
- Added `IPFS Guide` which redirects users to [here](https://docs.ipfs.tech/).


lidel (approved):
Thank you!

PR #415, opened 2023-05-18
Specify "piece gateways"

willscott:
As we move towards specifying what it means to be a "storage provider with retrieval" we have found ourselves increasingly pointing to a set of specification documents - namely [IPIP-402](https://github.com/ipfs/specs/pull/402) / `Trustless HTTP gateway` spec, and the IPNI [ingestion](https://github.com/ipni/specs/blob/main/IPNI.md#ingestion) spec.

The remaining functionality expected is providing full "piece" files for HTTP download. These are content addressed archives, but are not incrementally verifiable in the form described here.

### DIFF
--- a/src/http-gateways/piece-gateway.md
+++ b/src/http-gateways/piece-gateway.md
@@ -1,0 +1,190 @@
+---
+title: Piece Gateway Specification
+description: >
+  The Piece Gateway describes an HTTP server namespace for integrating
+  Filecoin Piece resources within the existing HTTP stack.
+date: 2023-05-18
+maturity: WIP
+editors:
+  - name: Will Scott
+    github: willscott
+    url: https://wills.co.tt/
+  - name: Dirk McCormick
+    github: dirkmc
+tags: ['httpGateways', 'pieceGateway']
+order: 5
+---
+
+The most basic retrieval mechanism in filecoin is piece retrieval, an
+HTTP endpoint for directly fetching stored pieces.
+
+These pieces are mounted at the  `/piece/` namespace under the HTTP server root.
+
+**Note:** the Piece Gateway is one aspect of filecoin retrieval. The full retrieval
+semantics also include CID based retrieval within pieces using an IPFS-compatible gateway,
+and availability of IPNI-compatible indexes of the CIDs within each piece for indexing.
+
+# HTTP API
+
+Piece Gateway is an HTTP interface for requesting data at a specified content path.
+
+## `GET /piece/{piece cid}`
+
+Downloads data at the specified **immutable** path.
+
+- `piece cid` – a valid piece identifier ([CID](https://docs.ipfs.io/concepts/glossary#cid))
+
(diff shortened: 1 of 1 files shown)


lidel (changes requested):
Thank you for filling this, let's find it the right home.

Copying and continuing discussion from [IPFS-Implementers-Sync-2023-05-18](https://pl-strflt.notion.site/IPFS-Implementers-Sync-2023-05-18-438883f3714d4e029b3319855128392c):

>> **What is “piece”?**
> 
> Slightly bad merkletree of verifiable data (a CAR with extra extra steps)
> There is a spec for it 
    
I've looked at [the spec](https://spec.filecoin.io/systems/filecoin_files/piece/), and:
- The best to my knowledge, Filecoin Piece is an internal format proprietary to Filecoin network: a CAR with extra steps used for replicating blockchain deals. 
- Some of extra steps are not defined anywhere, one probably needs to read Go code to know what is the order of blocks in the CAR inside of Piece etc. 
- I was not able to find any uses outside Filecoin for this format, nor what is the IPFS user benefit here.

Lmk if it missed some aspect of Filecoin Piece that makes it generally useful to IPFS users.

>> **Why is this here?**
> 
> FIPs are for consensus changes
> There is a less developed governance process for Filecoin off-chain (FRC)

The way I see it, [FRC prior art exists](https://github.com/filecoin-project/FIPs/tree/master/FRCs), at least for things related to Filecoin Piece introduced in this PR:
- [FRC-0038](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0038.md) defines what Indexer is, how "piece" is mapped to location, and how Advertisement IPLD schema looks like. This is already way more than a basic HTTP API proposed in this PR. 
- [FRC-0040](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0040.md) describes what Boost is. It provides [HTTP retrieval services for Filecoin SPs](https://boost.filecoin.io/http-retrieval)** and implements `/piece` API proposed in this PR.

This suggests this spec belongs next to them, in [filecoin-project/FIPs/FRCs](https://github.com/filecoin-project/FIPs/tree/master/FRCs).  

My suggestion is to:
- if you believe this is trully useful outside Filecoin, create IPIP using [this template](https://github.com/ipfs/specs/raw/main/ipip-template.md) and fill  "Motivation", "User benefit" and "Alternatives" sections, so we have some fodder for discussion
- if not, move this spec to [filecoin-project/FIPs/FRCs](https://github.com/filecoin-project/FIPs/tree/master/FRCs) and work with Filecoin Project to figure out FRC governance. IPFS Project should not meddle with internal Filecoin standards.
  - reference implementation-wise, just like `boxo/gateway` provides one for `/ipfs`, something else should provide `/piece` (sounds like Boost already does this?)

willscott:
I think i'm okay pushing this to an FRC - that said I do think there is interrelation / a continuum here and it may benefit us to have some way to align things

PR #417, opened 2023-05-29
IPIP-417: Delegated Peer Routing HTTP API

hacdias:
This IPIP introduces the `/routing/v1/peers` endpoint to the Routing HTTP V1 API for peer routing support.

### DIFF
--- a/src/ipips/ipip-0417.md
+++ b/src/ipips/ipip-0417.md
@@ -1,0 +1,60 @@
+---
+title: "IPIP-0417: Delegated Peer Routing HTTP API"
+date: 2023-05-29
+ipip: proposal
+editors:
+  - name: Henrique Dias
+    github: hacdias
+    url: https://hacdias.com/
+    affiliation:
+      name: Protocol Labs
+      url: https://protocol.ai/
+relatedIssues:
+  - https://github.com/ipfs/specs/pull/410
+  - https://github.com/ipfs/kubo/pull/9877
+order: 417
+tags: ['ipips']
+---
+
+## Summary
+
+This IPIP specifies `/routing/v1/peers/{peer-id}` HTTP API to offload peer routing onto another server.
+
+## Motivation
+
+The motivation of this IPIP extends the one of :cite[ipip-0337] and :cite[ipip-0379],
+which introduced delegated content routing and delegated naming, respectively. Now,
+we expand upon those basis to introduce peer routing, reducing the barrier for interaction
+across different systems.
+
+## Detailed design
+
+Add `/routing/v1/peers/{peer-id}` to the existing :cite[http-routing-v1] specification,
+as well as the new Peer schema.
+
+## Design rationale
+
+In line with the remaining Routing V1 API, this IPIP introduces a new HTTP GET
+endpoint that is used to retrieve peer records for a certain peer.
+
+### User benefit
+
+The user benefit brought by this PR is similar to the one in :cite[ipip-0379],
+but instead of offloading the naming process, we offload the peer discovery and
+routing.
+
+### Compatibility
+
+See the "Compatibility" section of :cite[ipip-0337].
+
+### Security
+
+TODO
+
+### Alternatives
+
+TODO
+
+### Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).
--- a/src/ipips/ipip-0417.md
+++ b/src/ipips/ipip-0417.md
@@ -30,7 +30,7 @@ across different systems.
 ## Detailed design
 
 Add `/routing/v1/peers/{peer-id}` to the existing :cite[http-routing-v1] specification,
-as well as the new Peer schema.
+as well as the new Peer schema, that replaces the existing "known transfer protocols".
 
 ## Design rationale
 
@@ -45,7 +45,12 @@ routing.
 
 ### Compatibility
 
-See the "Compatibility" section of :cite[ipip-0337].
+The section "Known Transfer Protocols" has been removed and replaced by a "Known Schemas"
+section. Before, we used to have protocol specific schemas, such as `bitswap` and `graphsync-filecoinv1`.
+The usage of the removed schemas is no longer encouraged. Instead, clients and servers
+SHOULD be updated to use the new, more generic, `peer` schema.
+
+See more in the "Compatibility" section of :cite[ipip-0337].
 
 ### Security
 
--- a/src/ipips/ipip-0417.md
+++ b/src/ipips/ipip-0417.md
@@ -54,11 +54,11 @@ See more in the "Compatibility" section of :cite[ipip-0337].
 
 ### Security
 
-TODO
+See the "Security" section of :cite[ipip-0337].
 
 ### Alternatives
 
-TODO
+See the "Alternatives" section of :cite[ipip-0337].
 
 ### Copyright
 
--- a/src/ipips/ipip-0417.md
+++ b/src/ipips/ipip-0417.md
@@ -47,8 +47,11 @@ routing.
 
 The section "Known Transfer Protocols" has been removed and replaced by a "Known Schemas"
 section. Before, we used to have protocol specific schemas, such as `bitswap` and `graphsync-filecoinv1`.
-The usage of the removed schemas is no longer encouraged. Instead, clients and servers
-SHOULD be updated to use the new, more generic, `peer` schema.
+
+The usage of these schemas is no longer encouraged. Instead, clients and
+servers SHOULD be updated to use the new, more generic, `peer` schema, which
+avoids returning the same peer multiple times, making results more efficient
+when a peer supports more than one protocol.
 
 See more in the "Compatibility" section of :cite[ipip-0337].
 
--- a/src/routing/http-routing-v1.md
+++ b/src/routing/http-routing-v1.md
@@ -9,11 +9,26 @@ maturity: reliable
 editors:
   - name: Gus Eggert
     github: guseggert
+    affiliation:
+      name: Protocol Labs
+      url: https://protocol.ai/
   - name: Masih H. Derkani
     github: masih
+    affiliation:
+      name: Protocol Labs
+      url: https://protocol.ai/
   - name: Henrique Dias
     url: https://hacdias.com/
     github: hacdias
+    affiliation:
+      name: Protocol Labs
+      url: https://protocol.ai/
+  - name: Marcin Rataj
+    github: lidel
+    url: https://lidel.org/
+    affiliation:
+      name: Protocol Labs
+      url: https://protocol.ai/
 xref:
   - ipns-record
 order: 0
@@ -32,7 +47,7 @@ As such, human-readable encodings of types are preferred. This specification may
 
 - CIDs are always string-encoded using a [multibase]-encoded [CIDv1].
 - Multiaddrs are string-encoded according to the [human-readable multiaddr specification][multiaddr].
-- Peer IDs are string-encoded according [PeerID string representation specification][peer-id-representation].
+- Peer IDs are string-encoded according [PeerID string representation specification][peer-id-representation]: either a Multihash in Base58btc, or a CIDv1 with libp2p-key (`0x72`) codec.
 - Multibase bytes are string-encoded according to [the Multibase spec][multibase], and SHOULD use base64.
 - Timestamps are Unix millisecond epoch timestamps.
 
@@ -62,19 +77,21 @@ This API uses a standard version prefix in the path, such as `/v1/...`. If a bac
 {
   "Providers": [
     {
-      "Schema": "peer",
-      "Protocols": ["<protocol-a>", "<protocol-b>", ...],
-      "ID": "12D3K...",
+      "Schema": "<schema>",
+      "ID": "bafz...",
       "Addrs": ["/ip4/..."],
+      ...
     },
     ...
   ]
 }
 ```
 
-Responses SHOULD be limited to 100 providers. This endpoint MAY support [streaming](#streaming).
+The `application/json` responses SHOULD be limited to 100 providers.
 
-Each object in the `Providers` list is a record conforming to the [Peer Schema](#peer).
+The client SHOULD be able to make a request with `Accept: application/x-ndjson` and get a [stream](#streaming) with more results.
+
+Each object in the `Providers` list is a record conforming to a schema, usually the [Peer Schema](#peer-schema).
 
 ## Peer Routing API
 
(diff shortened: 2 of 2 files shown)


src/routing/http-routing-v1.md
```diff
@@ -103,6 +106,22 @@ Response limit: 100 providers
 
 Each object in the `Providers` list is a *read provider record*.
 
+### `GET /routing/v1/peers/{peer-id}`
+
+#### Path Parameters
+
+- `peer-id` is the [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) to fetch peer records for.
```
lidel, line 111: There are 3 string types that are "peer ids", two of them are legacy, see the end of:
https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#string-representation

@hacdias  Thoughts on dropping legacy and requiring CIDv1 with libp2p-key codec here? Makes implementation easy -- plug in CID library, extract Multihash, done. Don't worry about legacy.

If we don't do it, we introduce complexity that will be additional work for implementers who care, and bugs and bad UX  if they don't

src/routing/http-routing-v1.md
```diff
@@ -199,15 +217,38 @@ Access-Control-Allow-Origin: *
...
+  "Schema": "peer",
+  "ID": "12D3K...",
+  "Addrs": ["/ip4/..."],
+  "Protocols": ["transport-bitswap", ...]
+}
+```
+
+- `ID`: the [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
+- `Addrs`: a list of known [multiaddrs][multiaddr] for this peer. This list MAY be incomplete.
+- `Protocols`: an OPTIONAL field with the transports known to be used by this peer. If empty, it means the server does not know and the client can use the `identify` protocol to learn more.
```
lidel, line 240: @hacdias  Unsure if this field should be allowed to be `undefined`. There is no precedence for this in this API, all other fields are always present. Would it be safer to always return an empty list here?

```suggestion
- `Protocols`: a list of protocols known to be supported by this peer. 
  - If empty, it means the server is missing protocol information, and the client should use `ID` and `Addrs` to connect to the peer and use the [libp2p identify protocol](https://github.com/libp2p/specs/tree/master/identify) to learn about supported ones.
```

src/ipips/ipip-0417.md
```diff
@@ -0,0 +1,54 @@
+---
+title: "IPIP-0417: TODO"
```
lidel, line 2:
```suggestion
title: "IPIP-0417: Delegated Peer Routing HTTP API"
```

lidel (commented):
@hacdias I had to rebase this one to be based on the latest versions of the other two, apologies for the noise.

src/routing/http-routing-v1.md
hacdias: I'm happy with only allowing the CIDv1 version.

src/routing/http-routing-v1.md
hacdias: That's also okay.

src/routing/http-routing-v1.md
```diff
@@ -199,15 +218,37 @@ Access-Control-Allow-Origin: *
...
+The `peer` schema represents an arbitrary peer.
+
+```json
+{
+  "Schema": "peer",
+  "ID": "12D3K...",
+  "Addrs": ["/ip4/..."],
+  "Protocols": ["transport-bitswap", ...]
+}
+```
```
lidel, line 236: @masih just a quick check-in around using this schema more for DHT proxying:

This schema is added because of `/routing/v1/peers` but we also want to use it on `/routing/v1/providers` in cases where:
- a single peer speaks multiple protocols and we know that upfront, and protocols have no special metadata like the filecoin one, or
-  we don't know the protocol, dont want to assume bitswap, but we know peerid and/or multiaddrs and want to return them,

To illustrate, the current response when a peer speaks more than one protocol:
```json
{
      "Protocol": "transport-bitswap",
      "Schema": "bitswap",
      "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6ZmC",
      "Addrs": [
        "/dns4/elastic.dag.house/tcp/443/wss"
      ],
      "Metadata": "gBI="
    },
    {
      "Protocol": "unknown",
      "Schema": "unknown",
      "ID": "QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
      "Addrs": [
        "/dns4/dag.w3s.link/tcp/443/https"
      ]
    },
```

Our plan for DHT proxy in Kubo ([wip PR](https://github.com/ipfs/kubo/pull/9877/)) is to not return `unknown` schema, nor assume `bitswap`.  

Instead, reuse  `peer` schema from this IPIP:

```json
{
      "Schema": "peer",
      "ID": "QmUA9D3H7HeCYsirB3KmPSvZh3dNXMZas6Lwgr4fv1HTTp",
      "Addrs": [ 
        "/dns4/dag.w3s.link/tcp/443/https",
        "/dns4/elastic.dag.house/tcp/443/wss"
      ],
      "Protocols": [...] // optional entries, empty means unknown
    },
```

Any concerns with this approach?

guillaumemichel:
IMO it would be great to have delegated peer routing use signed peer records. It seems essential because without signatures you have to trust a middleman.

A simple "Signature" field could be added to the proposed schema. The signature can be the array of bytes resulting of the signature using the private key associated with the peer ID (and corresponding algorithm) of (all) the other (non-empty) fields of the schema.

src/routing/http-routing-v1.md
masih: `Protocols` is an array of string labels correct? 

If so, this is different from the existing unknown (in [boxo](https://github.com/ipfs/boxo/blob/main/routing/http/types/provider_unknown.go#L46)) where one could have additional fields in a protocol to convey information o whoever understands that protocol/schema.  I suspect it would be limiting to only allow string lables in `Protocols` for `/providers`.

hacdias:
@guillaumemichel I like the idea, but that'd require deeper changes to the way we implement things. Currently, the idea was to mostly "relay" the results of `dht findpeer` in Kubo for example. That doesn't give us signed records. Besides, this specification also indicates that the addresses and protocols may be incorrect.

All in all, we have to find the balance. For IPNS (#379), it is easy to send signed records, because that's literally what is passed around. Here it seems it'd be more complicated.

src/routing/http-routing-v1.md
lidel: Good catch @masih!

- yes, `Protocols` would be opaque strings
-  `Bytes` is undocumented, but sounds like we should add it here: good idea to allow opaque passtrough, it would also allow us to potentially pass signed records [mentioned](https://github.com/ipfs/specs/pull/417#issuecomment-1580795385) by @guillaumemichel 

My thinking is to add optional `ProtocolData` map, where keys would be the same as values in `Protocols`:

```json
{
      "Schema": "peer",
      "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6Zmb",
      "Addrs": [ 
        "/dns4/example.com/tcp/443/foobar",
        "/dns4/example.com/tcp/443"
      ],
      "Protocols": [ "transport-bitswap", "transport-foobar", "signed-libp2p-peer-record" ],
      "ProtocolData": {
           "transport-foobar": "[base64-blob]",
           "signed-libp2p-peer-record": "[base64-blob]",
      },
    },
```

This way, client can easily check if data of some protocol is present, or ignore it otherwise.

masih: The rationale behind `Bytes` in the implementation is to, i believe, keep the raw content of the message for signature verification etc.  The way it currently works is that:
* all messages include `Protocol` and `Schema`. 
* For known protocols, the parser expects other well known fields.
* for unknown protocols, the parser simply dups the message in a field **at implementation level** for user to the parse as they see fit. 

This way, at spec level we don't need a dedicated `Bytes` field. We just expect JSON objects that have at least the fields they must have. The rest is left to the user. This IMHO is a much better UX, because the complexity can be encapsulated in implementation level without adding clause at spec level to say "If protocol is unknown check the `Bytes` field, if not etc. etc."

src/routing/http-routing-v1.md
```diff
@@ -102,6 +102,25 @@ Response limit: 100 providers
...
+
+#### Response Status Codes
+
+- `200` (OK): the response body contains the peer record.
+- `404` (Not Found): must be returned if no matching records are found.
+- `422` (Unprocessable Entity): request does not conform to schema or semantic constraints.
+
+#### Response Body
+
+A [`peer` schema record](#peer).
```
hacdias, line 122: **Open Question**: should the response body for `/routing/v1/peers` be a single entity? What we want to obtain is the addresses of the peer. But addresses can end up being discovered on the fly. Supporting streaming is reasonably easy and could be part of the spec.

hacdias:
I'm marking this PR as open so we can discuss the feedback and open questions. /cc @lidel @masih

src/routing/http-routing-v1.md
lidel: @masih note  this IPIP is about Peer routing, not Content routing. 
Is your point that `Bytes` in `peer` Schema should not be a map, but a single item? 


I worry that a single value will not be enough: the IPFS ecosystem needs to have a surface for experimentation, we should allow alternative routing systems to have different types of signed provider records, which in my mind means this API should (at least in theory) allow passing the original bytes for more than one blob.

In practice, I don't expect us to use `Bytes` map much any time soon.

Or did I misread you and your point is that we should not have `peer` Schema at all, and instead return duplicated multiaddrs for every known transport protocol, and use `unknown` schema for unknown one?

src/routing/http-routing-v1.md
```diff
@@ -199,6 +218,28 @@ Access-Control-Allow-Origin: *
...
+  "Protocols": ["transport-bitswap", ...]
+}
+```
+
+- `ID`: the [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
+- `Addrs`: a list of known [multiaddrs][multiaddr] for this peer. This list MAY be incomplete.
+- `Protocols`: a list of protocols known to be supported by this peer. 
+  - If empty, it means the server is missing protocol information, and the client should use `ID` and `Addrs` to connect to the peer and use the [libp2p identify protocol](https://github.com/libp2p/specs/tree/master/identify) to learn about supported ones.
+
 ## Known Transfer Protocols
```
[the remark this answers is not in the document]
lidel: Imo yes, if we add `peer` schema, then schemas are no longer specific to transport protocols, and we want to rename to "Known Schemas"

src/routing/http-routing-v1.md
masih: > Is your point that Bytes in peer Schema should not be a map, but a single item?

Hi @lidel, that's not my point. Let me rephrase my suggestion:

 * Flatten the `Bytes` field in the outer message, where the spec mechanism tolerates unknown fields.

For example, your earlier example would instead look something like this:

```json
{
  "Schema": "peer",
  "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6Zmb",
  "Addrs": [
    "/dns4/example.com/tcp/443/foobar",
    "/dns4/example.com/tcp/443"
  ],
  "Protocols": [
    "transport-bitswap",
    "foobar",
    "signed-libp2p-peer-record"
  ],
  "foobar": "[base64-blob]",
  "signed-libp2p-peer-record": "[base64-blob]"
}
```

Here is another example that shows extensibility of the values where they don't have to be blobs:

```json
{
  "Schema": "peer",
  "ID": "QmQzqxhK82kAmKvARFZSkUVS6fo9sySaiogAnx5EnZ6Zmb",
  "Addrs": [
    "/dns4/example.com/tcp/443/foobar",
    "/dns4/example.com/tcp/443"
  ],
  "Protocols": [
    "transport-bitswap",
    "foobar",
    "signed-libp2p-peer-record",
    "fish"
  ],
  "foobar": "[base64-blob]",
  "signed-libp2p-peer-record": "[base64-blob]",
  "fish": {
    "fresh-water": false,
    "name": "barreleye"
  }
}
```


A parser that does not know of "foobar" or "fish" protocol, would simply ignore the unknown fields.

Does that make sense? WDYT?

src/routing/http-routing-v1.md
masih: cid.contact offers both: `/providers` returns list of all providers and `/providers/<peer-id>` returns a specific one.

src/routing/http-routing-v1.md
lidel: Ack, then we are aligned! I don't feel strongly about `Bytes`, just want to document a way for doing this in the spec to remove guesswork. Putting additional fields in the outer message sgtm. 

Editorial tasks here would be (note to self)
- [ ] explicitly state that parser MUST allow for unknown fields and clients MUST ignore unknown ones
- [ ] add paragraph/example with `foobar` protocol and `[base64-blob]`

hacdias: @lidel I added that information. Feel free to edit/change.

src/routing/http-routing-v1.md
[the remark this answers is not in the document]
hacdias: Okay, I did remove the "Known Transfer Protocols" section. I also removed the "Provider Record" section and updated the wording on the content providers API to link to the Peer schema.

src/routing/http-routing-v1.md
hacdias: I updated the Peers API to indicate it can return more than one. It will conform with the streaming API we already have too.

src/routing/http-routing-v1.md
```diff
@@ -199,58 +208,46 @@ Access-Control-Allow-Origin: *
...
-  "Protocol": "transport-graphsync-filecoinv1",
-  "Schema": "graphsync-filecoinv1",
+  "Schema": "peer",
   "ID": "12D3K...",
   "Addrs": ["/ip4/..."],
-  "PieceCID": "<cid>",
-  "VerifiedDeal": true,
-  "FastRetrieval": true
+  "Protocols": ["transport-bitswap", "foobar", ...],
+  "foobar": "[base64-blob]",
```
masih, line 246:
```suggestion
  "foobar": "[base64-blob]"
```

masih (approved):

lidel:
I believe this is ready for final reviews and ratification talk during next IPFS Implementers call.

FYSA we've got implementation in 
- client and server in `boxo/routing/http` library ([boxo/pull/422](https://github.com/ipfs/boxo/pull/422))
  - which will ship in Kubo 0.23 
    - server opt-in via `Gateway.ExposeRoutingAPI=true` config: [ipfs/kubo/pull/9877](https://github.com/ipfs/kubo/pull/9877)
    - client mode can be enabled via config (example in [test/cli/delegated_routing_v1_http_proxy_test.go#L28-L44](https://github.com/ipfs/kubo/blob/d2caa678ad30f70de32ce6b0a5e9bc8e45952dc0/test/cli/delegated_routing_v1_http_proxy_test.go#L28-L44))

lidel (approved):
Applied final editorials, and included legacy schemas with guidance to parse them the same way as `peer` one (to leverage presence of `ID` and `Addrs` fields). 

This is ready for final reviews and ratification.

BigLep:
@lidel : are we good to merge now?

lidel (approved):
This has been waiting for ratification feedback for a few weeks now, and no concerns were raised.
Thank you @hacdias  for landing this.

For anyone reading this in the future:
-  [Self-hosting /routing/v1 endpoint for delegated routing needs](https://github.com/ipfs/kubo/blob/master/docs/changelogs/v0.23.md#self-hosting-routingv1-endpoint-for-delegated-routing-needs) will ship in Kubo 0.23

PR #421, opened 2023-06-07
docs: IPIP for Delegated Routing Privacy Upgrade

ischasny:


src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,79 @@
...
+- a malicious actor spying on the user by observing the user to Content Router traffic and then downloading the same data;
+- the new API is a first step towards fully private HTTP Delegated Routing protocol that will eliminate IPNI as centralised observers.
+
+There are no other functional improvements.
+
+### Compatibility
+
+#### Backwards Compatibility
+
+The new API will be implemented in [go-delegated-routing](https://github.com/ipfs/boxo/tree/main/routing/http) and will not introduce any breaking changes.
```
lidel, line 69:
```suggestion
The `/routing/v1/encrypted/` API will be implemented in [`boxo/routing/http`](https://github.com/ipfs/boxo/tree/main/routing/http) and will not introduce any breaking changes to existing clear text endpoints.
```

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+
+## API Specification
+
+### Magic Values
+
+ All salts below are 64-bytes long, and represent a string padded with `\x00`.
+
+ - `SALT_DOUBLEHASH = bytes("CR_DOUBLEHASH\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
+ - `SALT_ENCRYPTIONKEY = bytes("CR_ENCRYPTIONKEY\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
+ - `SALT_NONCE = bytes("CR_NONCE\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
```
lidel, line 32: Missing definition what `CR_DOUBLEHASH`, `CR_ENCRYPTIONKEY` and `CR_NONCE` are, who comes up with values, and why we need them.

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+Hence the IPNI server needs to be able to compare encrypted values without having to decrypt them as that would require a key that it is unaware of.
+That means that the nonce has to be deterministically chosen so that `enc(passphrase, nonce, payload)` produces the same output for the same 
+`passpharase` + `payload` pair. Nonce must be calculated as `hash(SALT_NONCE || passphrase || len(payload) || payload)[:12]`, where `len(payload)` is 
+an 8-byte length of the `payload` encoded in Little Endian format. Choice of nonce is not enforced by the IPNI specification. The described approach will 
+be used while IPNI encrypts Advertisements on behaf of Publishers. However once Writer Privacy is implemented, the choice of nonce will be left up to the Publisher. 
+- **`CID`** is the [Content IDentifier](https://github.com/multiformats/cid).
+- **`MH`** is the [Multihash](https://github.com/multiformats/multihash) contained in a `CID`. It corresponds to the 
+digest of a hash function over some content. `MH` is represented as a 32-byte array.
```
lidel, line 48: Does this mean CIDs with longer hash functions are truncated at 32-byte mark?

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+- **`HASH2`** is a second hash over the multihash. Second Hashes must be of `Multihash` format with `DBL_SHA_256` codec. 
+The digest must be calculated as `hash(SALT_DOUBLEHASH || MH)`.
+- **`ProviderRecord`** is a Provider Record as described in the [HTTP Delegated Routing Specification](http-routing-v1.md).
```
lidel, line 51:
```suggestion
- **`ProviderRecord`** is a JSON with Provider Record as described in the [HTTP Delegated Routing Specification](http-routing-v1.md).
```

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+- **`ProviderRecordKey`** is a concatentation of `peerID || contextID`. There is no need for explicitly encoding lengths as they are
+already encoded as a part of the multihash format. 
+- **`EncProviderRecordKey`** is `Nonce || enc(deriveKey(multihash), Nonce, ProviderRecordKey)`.
+- **`HashProviderRecordKey`**  is a hash over `ProviderRecordKey` that must be calculated as `hash(SALT_DOUBLEHASH || ProviderRecordKey)`.
+- **`Metadata`** is free form bytes that can represent such information such as IPNI metadata.
```
lidel, line 56: Is there a max response size clients should expect, and reject anything bigger?
(for entire provider record, not just this field)

lidel (changes requested):
Thank you for submitting this @ischasny.
I'm OoO for the rest of the week, but dropping some quick questions after very rushed first pass (fine to link to prior discussions, if these were already asked before).

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,79 @@
...
+
+### User benefit
+
+With the new APIs users can protect themselves from:
+- a malicious actor spying on the user by observing the user to Content Router traffic and then downloading the same data;
+- the new API is a first step towards fully private HTTP Delegated Routing protocol that will eliminate IPNI as centralised observers.
+
+There are no other functional improvements.
+
+### Compatibility
```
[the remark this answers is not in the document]
ischasny: I think that Kubo's HTTP router should explicitly specify whether to use encrypted or not encrypted API. If a Content Routing Server claims that they support Privacy Preserving API then the user should expect that it provides a similar QoS than the non-Privacy Preserving one. 

That's a very good question - let me add a clarification to the IPIP.

src/routing/http-routing-reader-privacy-v1.md
ischasny: Thanks, added explanation, also removed `SALT_NONCE` as it's not relevant for this API.

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+
+```json
+{
+    "EncMetadata": "EBxdYDhd....."
+}
+```
+
+Where:
+
+- `EncMetadatas` is base58 encoded `EncMetadata`;
```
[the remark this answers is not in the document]
ischasny: Good point - changed. I specified it as base58 as we use that encoding in IPNI.

src/routing/http-routing-reader-privacy-v1.md
ischasny: Removed the 32-byte length as indeed it can be longer.

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+- **`EncProviderRecordKey`** is `Nonce || enc(deriveKey(multihash), Nonce, ProviderRecordKey)`.
+- **`HashProviderRecordKey`**  is a hash over `ProviderRecordKey` that must be calculated as `hash(SALT_DOUBLEHASH || ProviderRecordKey)`.
+- **`Metadata`** is free form bytes that can represent such information such as IPNI metadata.
+- **`EncMetadata`** is `Nonce || enc(deriveKey(ProviderRecordKey), Nonce, Metadata)`.
+
+### API
+
+Assembling a full `ProviderRecord` from the encrypted data will require multiple roundtrips to the server. The first one to fetch a list of `EncProviderRecordKey`s and then one per  
+`EncProviderRecordKey` to fetch `EncMetadata`. In order to reduce the number of roundtrips to one the client implementation should use the local libp2p peerstore for multiaddress discovery
+and [libp2p multistream select](https://github.com/multiformats/multistream-select) for protocol negotiation.
```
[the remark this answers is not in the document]
ischasny: Moved

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### `GET /routing/v1/encrypted/providers/{HASH2}`
+
+##### Response codes
+
+- `200` (OK): the response body contains 0 or more records
+- `404` (Not Found): must be returned if no matching records are found
```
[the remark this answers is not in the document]
ischasny: Good spot - fixed

src/routing/http-routing-reader-privacy-v1.md
ischasny: Added maximum lengths. Also added a comment that those might change without having to increment the API version (for example if IPNI starts allowing larger ContextIDs / Metadata in the future).

guillaumemichel:
LGTM!

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,94 @@
...
+Users should verify themselves whether a custom router of their choice supports Reader Privacy or not when configuring it.
+
+The `/routing/v1/encrypted/` API will be implemented in [`boxo/routing/http`](https://github.com/ipfs/boxo/tree/main/routing/http) and will not introduce any breaking changes to existing clear text endpoints.
+The API will be released in a new minor version. 
+
+#### Forwards Compatibility
+
+Reader Privacy relies on usage of specific hashing and encryption functions. Function rotation will require a network-wide migration. Content Routers might not be able to migrate "under the hood" as they
+don't possess the original values. Function rotation should be a very infrequent event and will require network-wide efforts. When function rotation is needed - a version of the API will be incremented.  
+
```
lidel, line 86: Some sections from https://raw.githubusercontent.com/ipfs/specs/main/ipip-template.md are  missing. If you have no content for them,  Just enter `n/a` for now.

```suggestion

### User benefit

TODO: How will end users benefit from this work?

### Security

TODO:  Explain the security implications/considerations relevant to the proposed change.

### Alternatives

TODO: Describe alternate designs that were considered and related work.

- TODO:  Oblivious HTTP ([IETF](https://www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html), [Cloudflare](https://blog.cloudflare.com/stronger-than-a-promise-proving-oblivious-http-privacy-properties/))

## Test fixtures

TODO:  List relevant CIDs (if any). Describe how implementations can use them to determine
specification compliance. This section can be skipped if IPIP does not deal
with the way IPFS handles content-addressed data, or the modified specification
file already includes this information.

```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,115 @@
...
+### User benefit
+
+TODO: How will end users benefit from this work?
+
+### Security
+
+TODO:  Explain the security implications/considerations relevant to the proposed change.
```
lidel, line 93:
```suggestion
See "Threat Modelling" section of TODO <!--ref[http-routing-reader-privacy-v1]-->
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,115 @@
...
+
+#### Forwards Compatibility
+
+Reader Privacy relies on usage of specific hashing and encryption functions. Function rotation will require a network-wide migration. Content Routers might not be able to migrate "under the hood" as they
+don't possess the original values. Function rotation should be a very infrequent event and will require network-wide efforts. When function rotation is needed - a version of the API will be incremented.  
+
+### User benefit
+
+TODO: How will end users benefit from this work?
+
```
lidel, line 90: (already present)
```suggestion
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,115 @@
...
+Users on older nodes can continue using the old API and turn on reader Privacy at a alter point. 
+
+Content Routers should provide the same QoS for both Privacy Preserving and regular APIs. This is because both can be served over the same encrypted data. If presented with a regular CID, a Content Router 
+can perform decryption operations on behalf of the user (i.e. mimic the client logic) and return results in clear text. If presented with a second hash the Content Router can return encrypted results and let the
+user to do decryption themselves.
+
+It's possible that not all Content Routers will adopt Reader Privacy. The default HTTP Delegated Router like `cid.contact` should have Reader Privacy enabled by default in the newer versions of Kubo / Helia. 
+Users should verify themselves whether a custom router of their choice supports Reader Privacy or not when configuring it.
+
+The `/routing/v1/encrypted/` API will be implemented in [`boxo/routing/http`](https://github.com/ipfs/boxo/tree/main/routing/http) and will not introduce any breaking changes to existing clear text endpoints.
```
lidel, line 79:
```suggestion
The `/routing/v1/encrypted/` API will be implemented in existing libraries like  [`boxo/routing/http`](https://github.com/ipfs/boxo/tree/main/routing/http) and will not introduce any breaking changes to existing clear text endpoints.
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,115 @@
...
+
+### Security
+
+TODO:  Explain the security implications/considerations relevant to the proposed change.
+
+### Alternatives
+
+TODO: Describe alternate designs that were considered and related work.
+
+- TODO:  Oblivious HTTP ([IETF](https://www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html), [Cloudflare](https://blog.cloudflare.com/stronger-than-a-promise-proving-oblivious-http-privacy-properties/))
```
lidel, line 99: I am trying to wrap my head around threat models for two hypothetical worlds (those are not exclusive):
- (A) IPNI `/routing/v1` supporting Double Hashed records (this IPIP)
- (B) IPNI `/routing/v1` behind OHTTP ([IETF](https://www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html), [Cloudflare](https://blog.cloudflare.com/stronger-than-a-promise-proving-oblivious-http-privacy-properties/))

My understanding is that in (A)  user needs to trust IPNI operator anyway to some degree, because IPNI's HTTP endpoint still can track a bunch of metadata about users query usage patterns, hours they are online, their IPs, user-agent, etc.

I think it would be beneficial to the reader privacy story if "Alternatives" section here mentioned OHTTP or Onion Services as a complimentary solutions that can be implemented and deployed in addition to this IPIP close the described gap.

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+### Magic Values
+
+ All salts below are 64-bytes long, and represent a string padded with `\x00`.
+
+ - `SALT_DOUBLEHASH = bytes("CR_DOUBLEHASH\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
+ - `SALT_DOUBLEHASH = bytes("CR_ENCRYPTIONKEY\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
+
+ Magic values are needed to calculate different digests from the same value for different purposes. For example a hash of a Multihash that is used for lookups should be different from the one that is used for
+ key derivation, even though both are calculated from the same original value. In order to do that the Multihash is concatenated with different magic values before applying the hash funciton - `SALT_DOUBLEHASH`
+ for lookups and `SALT_DOUBLEHASH` for key derivation as described in the `Glossary`.
```
lidel, line 35: We have `SALT_DOUBLEHASH` defined twice now  – probably a typo:
```suggestion
 - `SALT_DOUBLEHASH = bytes("CR_DOUBLEHASH\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
 - `SALT_ENCRYPTIONKEY = bytes("CR_ENCRYPTIONKEY\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`

 Magic values are needed to calculate different digests from the same value for different purposes. For example a hash of a Multihash that is used for lookups should be different from the one that is used for
 key derivation, even though both are calculated from the same original value. In order to do that the Multihash is concatenated with different magic values before applying the hash funciton - `SALT_DOUBLEHASH`
 for lookups and `SALT_ENCRYPTIONKEY` for key derivation as described in the `Glossary`.
```

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+This specification describes a new HTTP API for Privacy Preserving Delegated Content Routing provider lookups. It's an extension to HTTP Delegated Routing API and inherits all of its formats and design rationale. 
+
+## API Specification
+
+### Magic Values
+
+ All salts below are 64-bytes long, and represent a string padded with `\x00`.
+
+ - `SALT_DOUBLEHASH = bytes("CR_DOUBLEHASH\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
+ - `SALT_ENCRYPTIONKEY = bytes("CR_ENCRYPTIONKEY\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00")`
```
lidel, line 31: @ivilatamind adding definition of `CR_DOUBLEHASH` and `CR_ENCRYPTIONKEY` below, just for completeness?

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,107 @@
...
+- **`ProviderRecord`** is a JSON with Provider Record as described in the [HTTP Delegated Routing Specification](http-routing-v1.md).
+- **`ProviderRecordKey`** is a concatentation of `peerID || contextID`. There is no need for explicitly encoding lengths as they are
+already encoded as a part of the multihash format. Max `contextID` length is 64 bytes.
+- **`EncProviderRecordKey`** is `Nonce || enc(deriveKey(multihash), Nonce, ProviderRecordKey)`. Max `EncProviderRecordKey` is 200 bytes. 
+- **`HashProviderRecordKey`**  is a hash over `ProviderRecordKey` that must be calculated as `hash(SALT_DOUBLEHASH || ProviderRecordKey)`.
+- **`Metadata`** is free form bytes that can represent such information such as IPNI metadata. Max `Metadata` length is 1024 bytes. 
+- **`EncMetadata`** is `Nonce || enc(deriveKey(ProviderRecordKey), Nonce, Metadata)`. Max `EncMetadata` length is 2000 bytes.
+
+Note: maximum allowed lengths might change without incrementing the API version. Such fields as `contextID` or `Metadata` are free-form bytes and
+their maximum lengths can be changed in the underlying protocols.
```
lidel, line 57:
```suggestion
:::note

Maximum allowed lengths might change without incrementing the API version. Such fields as `contextID` or `Metadata` are free-form bytes and
their maximum lengths can be changed in the underlying protocols. 

:::
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,111 @@
...
+editors:
+  - name: Andrew Gillis
+    github: gammazero
+  - name: Ivan Schasny
+    github: ischasny 
+  - name: Masih Derkani
+    github: masih
+  - name: Will Scott
+    github: willscott
+order: XXX
```
lidel, line 14:
```suggestion
order: 421
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,111 @@
+---
+title: "IPIP-0421: HTTP Delegated Routing Reader Privacy Upgrade"
+date: 2023-05-31
+ipip: ratified
```
lidel, line 4:
```suggestion
ipip: proposal
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,111 @@
...
+
+See the Delegated Routing Reader Privacy Upgrade spec (:cite[http-routing-reader-privacy-v1]) included with this IPIP.
+
+## Design rationale
+
+This API proposal makes the following changes:
+- Adds new methods for looking up encrypted Provider Records and encrypted Metadata;
+- Defines Hashing and Encryption functions and response payloads structure.
+
+There are no ideomatic changes to the API - all data formats, design rationale and principles outlined in the original [HTTP Delegated Routing IPIP](./ipip-0337.md) apply here.
```
lidel, line 55:
```suggestion
There are no ideomatic changes to the API - all data formats, design rationale and principles outlined in the original :cite[ipip-0337] apply here. 
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,111 @@
...
+The API will be released in a new minor version. 
+
+#### Forwards Compatibility
+
+Reader Privacy relies on usage of specific hashing and encryption functions. Function rotation will require a network-wide migration. Content Routers might not be able to migrate "under the hood" as they
+don't possess the original values. Function rotation should be a very infrequent event and will require network-wide efforts. When function rotation is needed - a version of the API will be incremented.  
+
+### Security
+
+See "Threat Modelling" section of TODO <!--ref[http-routing-reader-privacy-v1]-->
```
lidel, line 89:
```suggestion
See "Threat Modelling" section of :cite[http-routing-reader-privacy-v1]
```

src/ipips/ipip-0421.md
```diff
@@ -0,0 +1,111 @@
...
+
+### Alternatives
+
+TODO: Describe alternate designs that were considered and related work.
+
+- TODO:  Oblivious HTTP ([IETF](https://www.ietf.org/archive/id/draft-thomson-http-oblivious-01.html), [Cloudflare](https://blog.cloudflare.com/stronger-than-a-promise-proving-oblivious-http-privacy-properties/))
+
+## Test fixtures
+
+TODO:  List relevant CIDs (if any). Describe how implementations can use them to determine
```
lidel, line 99:
```suggestion
TODO:  List relevant CIDs or JSON payloads. Describe how implementations can use them to determine
```

src/ipips/ipip-0421.md
masih: > Onion Services as a complimentary solutions


Agreed. I will take this.

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,111 @@
...
+
+:::note
+
+Maximum allowed lengths might change without incrementing the API version. Such fields as `contextID` or `Metadata` are free-form bytes and
+their maximum lengths can be changed in the underlying protocols. 
+
+:::
+
+### API
+#### `GET /routing/v1/encrypted/providers/{HASH2}`
```
masih, line 64: Thinking out loud: @lidel I am curious what your thoughts are in terms of making this work with existing `/routing/v1/providers/{cid}`, where depending on the multihash code of `cid` in URL the router returns results in encrypted form. This is how the IPNI lookup API works and I think it offers a better UX, where there is a single endpoint for users to look things up at. Given a multihash the users would no longer have to swtich over different multihash codes to pick the endpoint.

Thoughts?

aschmahmann: @masih you're saying that if you see a CID where the multihash code is https://github.com/multiformats/multicodec/blob/b98f2f38fc63c992545263815c2afdd692d2dc8e/table.csv#L42 that you automatically assume it's a double-hashed record lookup? If so that's IMO quite a bad idea as it leads to domain mixing/confusion.

For example: `bafkvmibdoraoaissb2hgfuowe6kfpb3lrgxgfusza74f5i6zob6wl5p5uy` is a valid CID with data in it that can render on gateways (CAR attached for posterity [dblsha256-example.car.gz](https://github.com/ipfs/specs/files/11818033/dblsha256-example.car.gz)). However, with this approach it becomes impossible to discover this data via cid.contact.

---

Given the misunderstanding here I'm also concerned that IPNI has also misused the code as it's been defined since 2016 which is `SHA256(SHA256(data))` https://github.com/multiformats/multicodec/pull/23#discussion_r85876712 not say `SHA256(SHA256_MH(data)`, `SHA256("My Salt" | SHA256(data))`, etc. Am I missing something here?

masih: > ~you~ IPNI automatically assume(s) it's a double-hashed record lookup?

Some what. Backends that only store encrypted records refuse to store or lookup records with any other multihash. 
On loadbalancer, queries are scatter gathered across nodes. So if there is an unencrypted records associated to a double hashed multihash then it does get discovered. 

Having said that there are implicit assumptions in ipni to treat double hashed multicodec as encrypted record lookup.

> Am I missing something

Nope. Thank you for digging up those comments I was unaware that this code is from 2016. 
Looks like we cannot make that implicit assumption. Its reuse for reader privacy is maybe unfortunate as having a dedicated code would have help a bunch of optimisations in lookup path.

src/routing/http-routing-reader-privacy-v1.md
masih: These are literal string values. I'll add a paragraph to clarify this

gammazero (approved):

src/routing/http-routing-reader-privacy-v1.md
```diff
@@ -0,0 +1,128 @@
...
+}
+```
+
+Where:
+
+- `EncMetadata` is a base64 encoded `EncMetadata`;
+
+### Notes
+
+Assembling a full `ProviderRecord` from the encrypted data requires multiple server roundtrips. The first fetches a list of `EncProviderRecordKey`s, followed by one for each `EncProviderRecordKey` to retrieve `EncMetadata`. To minimize the number of roundtrips to one, the client implementation should use the local libp2p peerstore for multiaddress discovery and [libp2p multistream select](https://github.com/multiformats/multistream-select) for protocol negotiation.
```
aschmahmann, line 128: @masih we've talked about this before and it's more exploratory, but I'm noting here to make it public and see if folks have any thoughts about how this might impact the interface/API here in the future.

Note: This is not a "please rewrite everything" request.

---

How well this works is a function of how important the metadata is to performing a useful retrieval, and how important the metadata is depends on the distribution of information between the "ProviderRecord" and the "(ProviderRecord)Metadata".

IIUC the reason it's implemented this way is to keep data storage in routing backends like IPNI from needing to store the same data but encrypted many many times (i.e. once per multihash advertised).

At the extremes we have:
-  A lot of space could be saved by making `ProviderRecord` just contain a small pointer and `Metadata` contain all the actual provider record information (e.g. peerID, multiaddrs, protocols, ...). This means two round-trips unless the client already has the pointer information locally (e.g. if the pointer was a peerID, then having the multiaddrs, etc. locally, and if the pointer was some system-specific unique-ID then having that cached from a prior lookup). Also, unless there's some aggregation service/proxy it allows correlation between many different requests that use the same metadata (not necessarily a big deal here).
- The second round-trip could disappear if we store all the information in the `ProviderRecord` portion. However, this means encrypting all the data for every advertised multihash

I could situationally see reasons to shuffle data between these two, depending on things like:
1) how  reusable the metadata is
2) how frequently the metadata information is to be cached
3) cost models for routing systems
...

Two areas where I could see the extremes being in use:
1. Save storage by just returning the target's "identifier": e.g. libp2p peerID or an unauthenticated [HTTP+libp2p](https://github.com/libp2p/specs/pull/508) URL (that has .well_known for protocol negotiation) since peer routing makes sense to be separable in libp2p, and protocol negotiation to be separate for unauthenticated HTTP + libp2p.
2. Save round-trips by returning all the information: e.g. a [webseeds-like](https://www.bittorrent.org/beps/bep_0019.html) advertisement that points to an [outboard blake3 HTTP URL](https://github.com/oconnor663/bao/blob/805a042d56bd671ff805c18b2b226689d378593b/docs/spec.md#outboard-encoding-format) and an HTTP URL for the data (data could be a separate advertisement)

This makes me wonder if there's a better way to do this. For example:
- `encrypted/providers` returns a (JSON) blob mimicking the routing-v1 results
- It contains some information indicating if there's metadata and/or what might be in the metadata
- `encrypted/metadata` provides the metadata

This could allow systems like IPNI to optimize data layouts on ingestion and have some flexibility without breaking downstream clients.

PR #425, opened 2023-07-06
IPIP-425: Signaling Features on HTTP Gateways

lidel:
- [x] merge [IPIP-402](https://github.com/ipfs/specs/pull/402)  and [IPIP-412](https://github.com/ipfs/specs/pull/412) first (in order), this PR is based on top of them
- [ ] gateway conformance test

### DIFF
--- a/src/http-gateways/path-gateway.md
+++ b/src/http-gateways/path-gateway.md
@@ -542,6 +542,46 @@ Optional, present in certain response types:
   non-executable binary response types are not used in `<script>` and `<style>`
   HTML tags.
 
+### `Ipfs-Gateway-Features` (response header)
+
+Optional, this header SHOULD be only returned in response to HTTP `OPTIONS` request.
+
+The value is a list of key-value pairs, as specified by the 5.6.1. Lists section of :cite[rfc9110].
+
+Each feature is indicated by a key and optional value. When more than one value is supported, `|` is used as a separator.
+For example:
+
+```
+Ipfs-Gateway-Features: foo, bar=1, buzz=a|b|c
+```
+
+A Gateway SHOULD use this header to communicate support for specific Gateway feeatures, enabling clients to make better decisions on how to retrieve data.
+
+A Gateway MAY define and return their own features.
+
+A Client MUST send HTTP OPTIONS request to inspect this header before performaning any more expensive feature-detection.
+
+#### Canonical Ipfs-Features values
+
+- `trustless-gateway` for :cite[trustless-gateway]
+  - `path-proof` indicates support for returning parent blocks up to the terminus element
+  - `car-version=1|2` indicates CAR support
+  - `dag-scope=block|entity|all` from :cite[ipip-0402]
+  - `entity-bytes` from :cite[ipip-0402], implies support for `dag-scope=entity` as well
+  - `car-block-order=dfs` from :cite[ipip-0412]
+  - `car-block-dupes=y|n`  from :cite[ipip-0412]
+- `path-gateway` for deserialized responses defined by :cite[path-gateway]
+- `subdomain-gateway=example.com` for :cite[subdomain-gateway] support based on `Host` header
+- `dnslink-gateway` for :cite[dnslink-gateway] support based on `Host` header
+- `ipns` indicating :cite[ipns-record] support on `/ipns/` content paths
+- `dnslink` indicating [DNSLink](https://dnslink.dev) support on `/ipns/` content paths
+
+<!-- TODO do we want these too?
+- `multibase` list of prefixes, indicates which multibase encoding are supported in CIDs
+- `multihash` indicates which hash functions are supported in CIDs
+- `multicodec` indicates which codecs are supported in CIDs
+-->
+
 ### `Server-Timing` (response header)
 
 Optional. Implementations MAY use this header to communicate one or more
--- a/src/ipips/ipip-0425.md
+++ b/src/ipips/ipip-0425.md
@@ -1,0 +1,92 @@
+---
+title: "IPIP-0425: Signaling Features on HTTP Gateways"
+date: 2023-07-06
+ipip: proposal
+editors:
+  - name: Marcin Rataj
+    github: lidel
+    url: https://lidel.org/
+relatedIssues:
+  - https://github.com/ipfs/specs/pull/402#pullrequestreview-1396116569
+  - https://github.com/ipfs/specs/pull/412#pullrequestreview-1427137365
+order: 425
+tags: ['ipips']
+---
+
+## Summary
+
+Add ability to query HTTP Gateway for an explicit list of supported features.
+
+## Motivation
+
+A Gateway always ships with an opinionated set of supported hash functions and
+IPLD codecs, and the differences between implementations will grow over time.
+
+For example, some legacy gateways may not support newly added features like
+`dag-scope` and `entity-bytes` from :cite[ipip-0402] or the ability to get some
+block ordering guarantees introduced in :cite[ipip-0412]. Future IPIPs may add
+more features and response formats.
+
+We need a light mechanism for clients to detect which gateway supports partial CARs
+
+## Detailed design
+
+This IPIP introduces a set of HTTP headers returned in response to `OPTIONS` request:
+
+The `Ipfs-Gateway-Features` header is used for signalling support for specific Gateway features to the client.
+
+The lack of the header, or missing key-value pair within the header means support status is unknown.
+
+Initial list of key-value pairs is documented in `Ipfs-Gateway-Features` section of :cite[path-gateway]
+
+## Design rationale
+
+There is a good prior art for this in web browsers where HTTP `OPTIONS` method
+is used in [CORS Preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request)
+that checks if the CORS protocol is understood and a server is aware using
+specific methods and headers.
+
+The `OPTIONS` request if often sent by web browser anyway, so we would not be
+introducing much overhead.
+
+### User benefit
+
+Reduced cost for both client and gateway. Sending `OPTIONS` request will be
+very inexpensive, especially when compared with current status quo where a
+client has to send at least one request to probe a specific feature like
+support forp Blake3 hash function or `dag-scope` or `entity-bytes`.
+
+This translates to decreased latency and ability to choose the best retrieval
+strategy faster.
+
+### Compatibility
+
+This IPIP is fully backward-compatible with browsers and existing IPFS
+ecosystem. Gateways already return CORS headers with `OPTIONS` responses, we
+will simply return additional headers with the same responses.
+
+:::issue
+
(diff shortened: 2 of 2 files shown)


src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+
+A Client MUST send HTTP OPTIONS request to inspect this header before performaning any more expensive feature-detection.
+
+#### Canonical Ipfs-Features values
+
+- `trustless-gateway` for :cite[trustless-gateway]
+  - `path-proof` indicates support for returning parent blocks up to the terminus element
+  - `car-version=1|2` indicates CAR support
+  - `dag-scope=block|entity|all` from :cite[ipip-0402]
+  - `entity-bytes` from :cite[ipip-0402], implies support for `dag-scope=entity` as well
```
lidel, line 570: @hannahhoward @alanshaw @olizilla  I know you want a way for signaling support for partial cars and range requests between Lassie and .storage to avoid expensive feature-sniffing.

With this IPIP, signaling support for partial cars WITHOUT range requests woudl look like this:

```http
Ipfs-Gateway-Features: dag-scope=block|entity|all
```

and if you add support for `entity-bytes` at some point:

```http
Ipfs-Gateway-Features: dag-scope=block|entity|all, entity-bytes
```

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+- `subdomain-gateway=example.com` for :cite[subdomain-gateway] support based on `Host` header
+- `dnslink-gateway` for :cite[dnslink-gateway] support based on `Host` header
+- `ipns` indicating :cite[ipns-record] support on `/ipns/` content paths
+- `dnslink` indicating [DNSLink](https://dnslink.dev) support on `/ipns/` content paths
+
+<!-- TODO do we want these too?
+- `multibase` list of prefixes, indicates which multibase encoding are supported in CIDs
+- `multihash` indicates which hash functions are supported in CIDs
+- `multicodec` indicates which codecs are supported in CIDs
+-->
```
lidel, line 583: Thoughts :question:  

The need for list of supported IPLD codecs was raised in https://github.com/ipfs/specs/pull/402#discussion_r1180101957, having the other two would not hurt too.

src/ipips/ipip-0425.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Security
+
+This IPIP does not introduce any new security concerns. Probing gateway for
+supported features and hash functions is already possible via regular `GET`
+requests.
+
+### Alternatives
+
+- Exposing the list of suported features via `GET
```
MarcoPolo, line 83: I agree that using `OPTIONS` here makes more sense for this use case than using `.well-known`. My argument is that this the HTTP Gateway API is an application-level protocol, and these are features for this application. [`.well-known`](https://www.rfc-editor.org/rfc/rfc8615) on the other hand is for metadata about an origin, not a specific application.

This is also why https://github.com/libp2p/specs/pull/508 uses `.well-known` for metadata on where application protocols are mounted. This is metadata about the origin.

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+  - `dag-scope=block|entity|all` from :cite[ipip-0402]
+  - `entity-bytes` from :cite[ipip-0402], implies support for `dag-scope=entity` as well
+  - `car-block-order=dfs` from :cite[ipip-0412]
+  - `car-block-dupes=y|n`  from :cite[ipip-0412]
+- `path-gateway` for deserialized responses defined by :cite[path-gateway]
+- `subdomain-gateway=example.com` for :cite[subdomain-gateway] support based on `Host` header
+- `dnslink-gateway` for :cite[dnslink-gateway] support based on `Host` header
+- `ipns` indicating :cite[ipns-record] support on `/ipns/` content paths
+- `dnslink` indicating [DNSLink](https://dnslink.dev) support on `/ipns/` content paths
+
```
lidel, line 578: Another thing that could be useful is max block size:
```suggestion
- `max-block-size` to indicate what is the biggest block that can be retrieved by this gateway (default: 1MiB? 2MiB? -- whatever we currently have in boxo/gateway)
```

BigLep (commented):
Thanks for writing this up.  Makes sense and seems like a great step forward for letting nodes announce the functionality they support.  

Rather than embedding all of this in the header, is there any precedent for putting it into a body with JSON?  (Just was wondering in case easier to work with and avoids any max header length situations?)

src/http-gateways/path-gateway.md
lidel, line 578: Another thing, signal  hereif gateway is recursive (cc @aschmahmann  @Jorropo)

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
   non-executable binary response types are not used in `<script>` and `<style>`
   HTML tags.
 
+### `Ipfs-Gateway-Features` (response header)
+
+Optional, this header SHOULD be only returned in response to HTTP `OPTIONS` request.
```
rvagg, line 547: (just tuning into this PR now that I care about signalling, sorry!)


Whatever this header is, I think it should also be returned when a server doesn't implement a feature requested, either via `Accept` header or other means and in that case the server should return a `415`. That way I can parse the list of `Accept` content types, and be strict about the parameters, and if I don't find one that I fully support, return a `415` with this signal that says what I do support.

That way, a client can optimistically make a request, but be prepared to downgrade if it doesn't get what it wants. Rather than needing to `OPTIONS` first, thereby saving a round-trip.

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+
+A Gateway MAY define and return their own features.
+
+A Client MUST send HTTP OPTIONS request to inspect this header before performaning any more expensive feature-detection.
+
+#### Canonical Ipfs-Features values
+
+- `trustless-gateway` for :cite[trustless-gateway]
+  - `path-proof` indicates support for returning parent blocks up to the terminus element
+  - `car-version=1|2` indicates CAR support
```
rvagg, line 568:
```suggestion
  - `car-version=1` indicates CAR support
```

we should strictly do v1 for now, v2 doesn't make sense for this (yet), but it might be worth signalling this so we can also add `3` etc. when we get to it

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+  - `dag-scope=block|entity|all` from :cite[ipip-0402]
+  - `entity-bytes` from :cite[ipip-0402], implies support for `dag-scope=entity` as well
+  - `car-block-order=dfs` from :cite[ipip-0412]
+  - `car-block-dupes=y|n`  from :cite[ipip-0412]
```
rvagg, line 572:
```suggestion
  - `dups=y|n`  from :cite[ipip-0412]
```

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+A Client MUST send HTTP OPTIONS request to inspect this header before performaning any more expensive feature-detection.
+
+#### Canonical Ipfs-Features values
+
+- `trustless-gateway` for :cite[trustless-gateway]
+  - `path-proof` indicates support for returning parent blocks up to the terminus element
+  - `car-version=1|2` indicates CAR support
+  - `dag-scope=block|entity|all` from :cite[ipip-0402]
+  - `entity-bytes` from :cite[ipip-0402], implies support for `dag-scope=entity` as well
+  - `car-block-order=dfs` from :cite[ipip-0412]
```
rvagg, line 571:
```suggestion
  - `order=dfs` from :cite[ipip-0412]
```

src/http-gateways/path-gateway.md
rvagg: good idea but hold off until we have a real use-case for it reckon - unless you can conjure one now?

src/http-gateways/path-gateway.md
rvagg: too much information, let's start simple for now

rvagg:
I ~like this, but would rather its feature set be somewhat limited within the bounds of 412 plus anything else we can come up with a legitimate use-case for signalling. I don't really buy the multicodec thing yet—I know why it's a problem, but how does signalling it help at this stage? The tighter we scope it the easier it is to agree and ship, and I'd really like to turn this "ignore" into an error so evolution is easier: https://github.com/ipld/go-trustless-utils/blob/27d979b92be45fa189c9bef4a351e5cb02d95285/http/parse.go#L187-L189

rvagg:
My original thought for this was to just `X-Supported-Content-Types` and enumerate all the permutations of `Accept` that would be acceptable. That gets kind of combinatorially explosive as we add more items; although we could parameterize or regex it:

```
X-Supported-Content-Types: application/vnd.ipld.(raw|car);dups=(y|n);order=dfs;version=1
```

But I also suppose there's things in here that are not content-type related.

src/http-gateways/path-gateway.md
```diff
@@ -542,6 +542,46 @@ Optional, present in certain response types:
...
+A Gateway SHOULD use this header to communicate support for specific Gateway feeatures, enabling clients to make better decisions on how to retrieve data.
+
+A Gateway MAY define and return their own features.
+
+A Client MUST send HTTP OPTIONS request to inspect this header before performaning any more expensive feature-detection.
+
+#### Canonical Ipfs-Features values
+
+- `trustless-gateway` for :cite[trustless-gateway]
+  - `path-proof` indicates support for returning parent blocks up to the terminus element
```
rvagg, line 567:
```suggestion
- `trustless-gateway` for :cite[trustless-gateway]
  - `formats=car|raw` indicates whether CAR and/or raw block responses are supported
  - `path-proof` indicates support for returning parent blocks up to the terminus element
```

lidel:
Explicit signaling is difficult to rollout in decentralized ecosystem.

I no longer believe this IPIP is the realistic way of signalic capabilities to clients.
It has been 2 years and nobody pinged me for this, but we had to solve this problem by different means.

What works well, is active probing. Clients have to use standard HTTP GET and HEAD together with query params and `Accept` header  to probe new/niche capabilities before fully depending on them.

Identity CID is a good and backward-compatible way of performing a probe without incurring retrieval cost. We've documented this technique in https://specs.ipfs.tech/http-gateways/trustless-gateway/#dedicated-probe-paths

libp2p+HTTP interop is also documented in https://specs.ipfs.tech/http-gateways/libp2p-gateway/  and uses `.well-known` path for signaling support for additional libp2p+HTTP protocols.


Closing this for now. If someone find this IPIP useful, feel free to reuse whatever feels useful, but open a new PR with updated rationale.

PR #431, opened 2023-08-08
IPIP-431: Opt-in Extensible CAR Metadata on Trustless Gateway

bajtos:
Define an optional enhancement of the CARv1 stream that allows a Gateway server to provide additional metadata about the CARv1 response. Introduce a new content type that allows the client and the server to signal or negotiate the inclusion of extra metadata.

~~The PR discussing a new multi-codec `car-metadata`: https://github.com/multiformats/multicodec/pull/334~~


bajtos:
@lidel @willscott Here is my proposal for allowing gateway clients to request the response to include a metadata block at the end.

This is my first IPIP. Please let me know what and how to improve, where to add more details, etc. Feel free to edit the text directly if you like (edits by maintainers are allowed).

src/ipips/ipip-0431.md
lidel, line 1: @bajtos  fyi i've applied some editorial changes so you now get HTML preview on the PR.

due to this make sure to fetch my changes before  applying new ones :pray: 

>

src/ipips/ipip-0431.md
```diff
@@ -0,0 +1,125 @@
...
+- does not require registering new codec, or sniffing  the last DAG-CBOR block
+
+Downsides of this solution:
+
+- maintenance cost, requires duplicating of all CAR-related tests and features
+- ecosystem opportunity cost, in creating new content type, we increase
+  cognitive overhead for everyone working with IPFS over HTTP
+- no backward-compatible interop with existing tools and gateways that only
+  speak `application/vnd.ipld.car`
+- distracts us away from working on things like large blocks and CARv3
```
lidel, line 115: :information_source:  added a new content type as  "Plan B" to "Alternatives" here, with some initial pros/cons.

lidel (changes requested):
Thank you for submitting this @bajtos.

I like the framing of this as an extensible opt-in CAR manifest. Quick thought:
- Best case, it provides a way of solving multiple problems using existing tools and formats (CARv1, DAG-CBOR, CAR content type parameters). 
- Worst case, creates a standard for the ecosystem on how random metadata can be passed between paid services, which also allows CAR-aware clients to identify it and strip it out before storing in caches etc.

Made some editorial tweaks + quick first-pass feedback in comments inline.

willscott:
the keys / schema presented here, i think, should be considered an example. I would hope that it would be treated as an arbitrary key-value map of metadata objects, and that these key-values could be used to signal an error, could be used to signal an 'eof' signal, and/or could be used to provide additional check-sum attestation as described in the current text

rvagg:
I'm prototyping a form of option 1 above with Frisbii and Lassie, will let you know how it goes.

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,32 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+
+This parameter SHOULD only be used with CAR `version=1`.
+Values other than `eof` SHOULD be ignored.
+
+When the parameter is not set, the server must not add any extra CAR blocks to the response.
+
+The metadata block is a regular CAR block with the following properties:
+
+- CID specifies multicodec `car-metadata` (`0x04ff`), see
+  [multicodec#334](https://github.com/multiformats/multicodec/pull/334).
```
lidel, line 318: :information_source:  flagging that this custom codec is not a hard requirement for this IPIP.

As noted in "Option 1" from https://github.com/ipfs/specs/pull/431#issuecomment-1670958141  
`meta=eof` could use regular `dag-cbor` in CID it will still be ok, because `eof` behavior  is opt-in via `Accept` request header and there is explicit signaling in `Content-Type` response header returned before CAR stream.

One could argue this is more friendly to existing implementations, as `dag-cbor` code will not create errors in alternative implementations, while a custom code might.

```suggestion
The metadata block is a regular CAR block with the following properties:

- CID specifies multicodec `dag-cbor` (`0x71`)
```

rvagg:
https://github.com/filecoin-project/lassie/pull/378 and https://github.com/ipld/frisbii/pull/15 demonstrate an approximation of the option 1 I presented above.

* A `meta=eof` item in the `Accept` header is interpreted by the server as needing the trailing metadata to be sent, otherwise proceed as normal.
* Server writes the CAR, then a NUL byte, then a dag-json (opted for dag-json because it's human readable in case someone inspects the output in confusion) metadata chunk that includes some information about the CAR stream. Response is sent with a `Content-Type` that includes `meta=eof`.
* Client sees `meta=eof` in the `Content-Type` and reads the CAR using `ZeroLengthSectionAsEOF`; after that it slurps up the remaining bytes as dag-json into the metadata schema. It can then do `<stuff>` with it.

`<stuff>` that I've chosen to do at the moment is validate the car body according to the stats in the metadata and print out a "checksum" multihash that was provided by the server—this could be a signed checksum, or whatever it needs to be. There's also affordance for an `error` property in the metadata to signal that an error occurred and a message about it, so we get to see messages such as "block not found" now.

rvagg:
I'll acknowledge that it may be better to just go with the plain map approach, without a schema, as Will's suggested. That would even let us do novel things for specific situations like having Lassie tell Saturn about retrieval clients and their timings (currently can only use Server-Timings header which ends when the data starts). But there's a bit of a can of worms that we open up that I wondered if we could avoid by having a strong schema, at least for the first version. All of the things that http headers have to deal with - like what to do with duplicate keys, what limits we need to put on the sizes of things to avoid abuse, etc. Constraining within the bounds of dag-json, which itself is a bit strict, and having a schema, let's us be very clear about rules and avoid abuse.
But, maybe it's better to basically replicate the Trailers section but in the payload. 🤷

lidel:
@rvagg prefixing metadata with  `NUL` changes the scope of this IPIP, effectively moves metadata outside CARv1 serialization format that we have specs and compatible implementations, which makes it harder to argue reuse of the same content type as CAR. 

Existing CARv1 implementations will error without explicit support for `ZeroLengthSectionAsEOF`, right?

This thing starts looking like a new content type, changing the scope of this IPIP to something similar to `application/vnd.ipld.car-stream` from "Alternatives" section.  

Not saying it is bad, maybe a separate content type for streaming CARs  is the right call here. It mitigates risks around mixing regular CAR responses with ones that include metadata trailer and causing issues on clients that don't support `ZeroLengthSectionAsEOF`, and poisoning HTTP caches in scenarios when a rogue Saturn L1 sends a fishy CAR response. 

But i'm worried about duplicated effort across teams and project in light of CARv3, which (iiuc) also needs to happen some time in the next ~12-24 months and might have overlapping scope, solving similar problems.

---

@willscott @bajtos  is this IPIP something we intend to expose on all gateways and support forever in the IPFS ecosystem, even when we have CARv3? Would this be intended for wrapping CARv2 and v3 too? Or is this just a stop-gap for Rhea/Boost internally until we have CARv3 with built-in metadata/eof support?

rvagg:
Yeah you're right @lidel, this content type is "CARv1 but terminated with NUL byte", you need to tell the parser to handle it in that mode to see the content properly.

Arguably it could also be seen as "CARv1 within another container that has additional content at the end", which is also a different content type.

I think there is a strong desire here not to block metadata sending on a CARv3 because of how long that's likely to take to happen, whereas the work that needs this metadata has some time pressure associated with it.

willscott:
> is this IPIP something we intend to expose on all gateways and support forever in the IPFS ecosystem, even when we have CARv3?

it would probably need to support the retroactive validation of commitments.

rvagg:
Here's another option to throw into the mix: https://github.com/ipld/ipld/pull/295

* CARv2 wrapping the CARv1 (but no indexes)
* New characteristics bits to provide some interesting information but also signal new features
* `zero-terminated-payload` characteristic bit says that the length of the CARv1 payload isn't defined up-front, read it until it terminates at an `0x0` section.
* `trailer-message` characteristic bit says that there's a happy little message for you at the end of the payload that you can do what you like with.

The big catch with this is that it's not compatible with existing CAR implementations; they'll read a CAR created with this and bork at the `data-size=0` and think there's nothing in there. But we can easily upgrade implementations to understand these.

I think I consider this less breaky than my CARv1 zero-length-termination version because with that one there's not really an upgrade path for CAR implementations other than to do things like provide `--zero-length-termination` flags for the user to opt in to when consuming those CARs. At least with this one the answer is "oh, you're using an older <X> CAR implementation, upgrade and it'll load just fine".

Trustless Gateway implementations can still provide a CARv1 when we don't need extra stuff, we evolve it with a new `Accept` format that lets them switch to this format for additional signalling and optional metadata.

bajtos:
Hi folks, thank you for a constructive discussion and lots of new ideas for improving my proposal. :pray:


IIUC, we all agree with the following aspects:
- The ability to add some metadata to Gateways responses is very useful and can help solve multiple problems beyond what SPARK needs.
- The way to go is to add a new content-type parameter (or a new content type) to allow clients to opt into the new feature.

Then there are two areas where haven't reached a consensus yet:
1. How to design the new container format "CAR + metadata".
2. What is the schema of the metadata block, what is required vs. optional, and so on?

Let me also explain more about our context:
- As far as SPARK is concerned, we are looking for a stop-gap solution that will require the least effort to implement in Lassie & Frisbii & booster-http while not disrupting the rest of the IPFS ecosystem. 
  - We don't need all gateways to support this IPIP.
  - Since this IPIP is opt-in, I think we don't need support from all IPFS clients either.
  - We don't care about CARv2 as long as trustless gateways support CARv1 responses
- Once there is a "proper" solution for streaming CAR blocks and extra metadata, SPARK can migrate from the stop-gap solution to the new "proper" one. Such a migration will require effort and time, though. We cannot turn off the old feature until most SPs are running booster-http with the new feature. 

---

Let's dive into the container format now.

Considering the strong push-back in https://github.com/multiformats/multicodec/pull/334, I think we should look for options that don't require a new multicodec.

I quite like Rod's proposal to use dag-cbor and `car-metadata/v1` as the top-level key. It makes it easy to detect the metadata block by inspecting the first 17 bytes of the block payload and comparing them to the magic string `\xA1\x6Fcar-metadata/v1` (`A1` signals `map(1)`, `6F` signals `text(15)`). 

I am reluctant to introduce a zero-length EOF block:
- It opens the door for insertion attacks. People could interleave these EOF blocks into systems that don’t understand them, and when that data makes it into a system that does, these blocks end up closing the streams prematurely.
- As mentioned by @lidel, CARv1 spec does not allow zero-length blocks, so we no longer produce valid CARv1 and need to adjust CAR parsers (enable ZeroLengthSectionAsEOF).
- It changes the intent of this IPIP. Instead of adding an extra metadata block that does not change the current "end of transport stream" semantics, we are introducing a signalling mechanism indicating the end of the CARv1 file.

What do you think about the following proposal?

- The metadata block (tombstone) is sent after all CARv1 blocks are sent.
- There is no delimiter between CARv1 blocks and the tombstone.
- A block is a tombstone iff:
  - It's the last block in the CARv1 response received from the server.
  - Its CID indicates dag-cbor encoding.
  - The first 17 bytes of payload are `\xA1\x6Fcar-metadata/v1`.
    _I am open to tweaking this rule. E.g. we can say that the payload is a map with a top-level key "car-metadata/v1". That way, we can later support metadata blocks with two top-level keys, "car-metadata/v1" and "car-metadata/v2"._
- The entire response is still a valid CARv1 file. In the worst case, clients end up with a CARv1 file containing a CAR block that does not belong to the Merkle tree. (This should not happen in practice, because clients have to explicitly ask the gateway to include the metadata block.)

rvagg:
I'm not opposed to just tacking on a trailing block; it does make validating/consuming the data slightly harder than having a clean termination of the payload and the metadata sitting outside of it, but not terrible since we should know what the "end" is if we're validating while streaming. My main discomfort, and reason for suggesting other paths, is that you're mixing metadata in with the data that the metadata is describing. `<a thing>` `<something about the thing>` should be separate items.

You also get a CAR that's not quite what you asked for. I doubt there are handlers out there that are going to have problems with this (although Kubo `dag import` will find 2 separate DAGs in there to pin). As has been said already, content-type is the way to differentiate that and it'd have to be caveat emptor.

It is the path of least resistance if that's what we're going for with this.

With any of these approaches, if we're able to build in generally useful metadata (such as error messages or other information), then what we'd do in Lassie is attempt to opt-in all of our HTTP requests for this, and make allowance for both getting and not getting the metadata in whatever form it is, and not directly passing it on to the user, but separating it out appropriately (on the CLI it might just be extra printed information if that makes sense, like what https://github.com/filecoin-project/lassie/pull/378 does, in our API maybe we provide a separate result packet containing the information).

willscott:
* i think the tacking on extra block makes sense - both as a 'path of least resistance' / low work as you mention, and as a direct way to accomplish the immediate need.
* you note "I doubt there are handlers out there that are going to have problems with this" - if the request needs to explicitly opt-in signalling to ask for this extra block, then I'm not too worried about this.
* agree we should have Lassie attempt to opt-in all http requests and handle appropriately.

bajtos:
> I'm not opposed to just tacking on a trailing block; it does make validating/consuming the data slightly harder than having a clean termination of the payload and the metadata sitting outside of it, but not terrible since we should know what the "end" is if we're validating while streaming. My main discomfort, and reason for suggesting other paths, is that you're mixing metadata in with the data that the metadata is describing. `<a thing>` `<something about the thing>` should be separate items.

I see your point, @rvagg. As was thinking about how to update your Lassie PR to my new proposal, I realised that the condition "the last block" is indeed more difficult to implement.

I quite like how the zero-byte proposal allows us to think about the payload as a container containing two compartments: CARv1 and metadata.

```
<CARv1 blocks> <0x00 byte> <metadata>
```

We can prevent insertion attacks by requiring the party producing this container (Frisbee, Boost) to remove any zero-length blocks from the CARv1 payload. I think this is a reasonable requirement, considering that `0x00` is not allowed in CARv1 anyway.

Because we are inventing a new container-based format here, I am arguing that we don't necessarily have to adjust CAR parsers. Instead, we can add a simple middleware-like filter between the source of CARv1 blocks (the upstream HTTP service) and the CARv1 parser, and let this middleware stop sending bytes to the parser when it reaches the first zero-length block. In practice, enabling `ZeroLengthSectionAsEOF` is the faster option to a working version in our Go stack. I think it's important that `ZeroLengthSectionAsEOF` is one of many options for how to implement this, a kind of implementation detail.

When we look at the proposal as a new container format, we are not changing the current "end of transport stream" semantics. The stream contains the entire container, and the zero-length block is a part of that container. 

The biggest remaining item is the content type for this new container format.
 - The container is not a valid CARv1 because of the `0x00` block.
 - I am reluctant to call it CARv3, as there are high expectations and a lot of requirements for what CARv3 should be.
 - We should support all existing parameters of `application/vnd.ipld.car` content type, most notably `dag-scope`, `order`, `dfs`.
 
In https://github.com/ipfs/specs/pull/431#issuecomment-1683041576, @lidel proposed a new content type `application/vnd.ipld.car-stream`.
- Any objections against that particular name? Suggestions for a more accurate name?
- What are the ramifications of introducing a new content type? Can you foresee any problems this may cause to the ecosystem?
  - Keep in mind that HTTP clients must explicitly ask for this new content type.
  - To allow new clients to work with old gateways, should we require them to send both "car" and "car-stream" content types in the `Accept` header? Can existing gateways handle such `Accept` headers?

Thoughts?

> With any of these approaches, if we're able to build in generally useful metadata (such as error messages or other information), then what we'd do in Lassie is attempt to opt-in all of our HTTP requests for this, and make allowance for both getting and not getting the metadata in whatever form it is, and not directly passing it on to the user, but separating it out appropriately (on the CLI it might just be extra printed information if that makes sense, like what https://github.com/filecoin-project/lassie/pull/378 does, in our API maybe we provide a separate result packet containing the information).

+1 to separate out the metadata for the CLI and Go API users.

I think Lassie's HTTP server should produce its own metadata block if the client asks for it via content-type negotiation. It allows Lassie to add extra metadata to the response, which is one of the goals of this IPIP.

Where it becomes tricky: In SPARK, we need to obtain the original metadata block that Lassie received from the storage provider. I was initially thinking that Lassie could simply forward the metadata block received from the storage provider.

Now that I think about this again, I can see how that can be problematic in the case when Lassie modifies the CAR stream received from the SP, e.g. to re-order blocks in DFS order and include duplicate blocks. In such cases, SP-provided metadata fields like "car length" and "Blake3 hash of CARv1 payload" have no meaning for Lassie clients, as they will receive a different CAR file.

Is there a way to ask Lassie's HTTP server to respond with the exact CAR file it received from the upstream (the storage provider)? If there isn't an easy solution for this (existing or easy to implement), then I guess SPARK will have to retrieve content directly from SPs, crafting the request to be indistinguishable from Lassie's requests.

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,32 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
 the raw data is already present in the parent block that links to the identity
 CID.
 
+## CAR `meta` (content type parameter)
+
+The `meta=eof` parameter allows clients to request the server to include additional metadata about the
+CAR to be included at the end of the response body.
```
lidel, line 308: :thought_balloon: 

If this IPIP ends up doing `<CARv1> <0x00 byte> <dag-cbor-metadata>` would it be less mysterious and more self-explanatory if we did `eof=cbor`?

lidel (commented):
> We should support all existing parameters of application/vnd.ipld.car content type, most notably dag-scope, order, dfs.
> What are the ramifications of introducing a new content type?
> Can you foresee any problems this may cause to the ecosystem?

### On the scope here and keeping `application/vnd.ipld.car`

When we've discussed creation of this IPIP during the initial call last month, my understanding was that this is just a low-impact stop-gap until CARv3 exists. The goal was to find a way of passing arbitrary metadata in a way that is interoperable with existing ecosystem  to minimize work and impact until we can switch to CARv3.

Now, this IPIP seems to grow with scope and is about  `<CARv1> <0x00 byte> <dag-cbor-metadata>` which starts looking like a new wire format and a separate content type, BUT at the same time it aims to support the same features from IPIP-402 as existing CAR content type. 

It is the same CAR bytes, same features, and the only difference is a DAG-CBOR trailer prefixed with `0x00`.

At this point, I am honestly not sure if a separate content type is the lesser evil.

My rationale is:

- **introducing new content type like `application/vnd.ipld.car-stream`** 
  - comes with huge externalized cost: increased complexity in specs, tests, spotted support across ecosystem, issues with content negotiation
  - we will most likely end up leaking this new content type across ecosystem, and be stuck with it. external actors will cut corners. sooner or later there will be gateways that only implement this new content type, which shifts implementation complexity cost to all clients who want to be able to leverage all gateways
  - This will create pretty annoying cognitive overhead that makes adoption harder.  Explaining why we need two CAR _flavours_ when we already have _version_ field makes IPFS stack look a bit silly.
- **keeping CAR content type (`application/vnd.ipld.car`) and having opt-in via optional content type parameter feels like net positive:
  - implementers still leverage gateway-conformance test suite for correctness. 
    - regular CAR is mandatory, and manifest at the end is optional and explicit opt-in
      - this makes it more difficult for a gateway to become harmful to the ecosystem and only support non-CAR content type, as the cost of ignoring conformance is higher (one would have to write and maintain own tests)
      - it is easier fr ecosystem actors to maintain radical interop (following [the robustness principle](https://specs.ipfs.tech/architecture/principles/#robustness)) and make CAR libraries, proxies and parsers ignore`<0x00 byte> <metadata>` by default for better DX and interop (in case of cache bugs etc). 
        - arguably, easier to pull of in a decentralized system, than to force everyone to support two content types.
  - reduces spec/testing complexity, legacy clients simply ignore it and can still use the same gateway.
  - implementatons/specifications/compliance no longer duplicate everything, only need to test opt-in scenario

@bajtos  @rvagg  @willscott  with that in mind, given this is effectively a surgical stop-gap until CARv3, even if you want to go the `<CARv1> <0x00 byte> <metadata>` route, my suggestion would be to reuse the CAR content type and introduce this via optional param

Keeping the same content type feels like the lesser evil for the ecosystem (smaller waves, easier to deal with fallout).

------

PS.  re: content type negotiation on legacy gateways:

>  Keep in mind that HTTP clients must explicitly ask for this new content type.
> To allow new clients to work with old gateways, should we require them to send both "car" and "car-stream" content types in the Accept header? Can existing gateways handle such `Accept` headers?

On the spec level, regular HTTP content negotiation rules apply, the intention is for Gateways do not invent anything special, just use `Accept` header for format opt-in.

In practice, gateways that use  `boxo/gateway` implement very naive content type negotiation that only accounts for one custom format. The `Accept` header can be a comma separated list, the gateway ignores unknown ones, and will pick the first content type that starts with `application/vnd.ipld` ([boxo//gateway/handler.go#L639-L656](https://github.com/ipfs/boxo/blob/574df963653c940ea19b1b8a0d37cd30360c22e9/gateway/handler.go#L639-L656)).

This means if you ask for `application/vnd.ipld.foo` first, then boxo gateway will error with `unsupported format "application/vnd.ipld.foo"` ([handler.go#L324-L326](https://github.com/ipfs/boxo/blob/574df963653c940ea19b1b8a0d37cd30360c22e9/gateway/handler.go#L324-L326)).

Ways to avoid this error:
- implement proper content negotiation in `boxo/gateway` (with support for `q` weights), 
- or workaround for now by sending  `Accept` with `application/vnd.ipld.car` before the custom content type 
- or have content type that does not start with `application/vnd.ipld`
- or use `application/vnd.ipld.car` with optional parameter, like this IPIP initially suggested – imo preferable, as noted above.

rvagg:
At this point I'm easy; let me know how I can help iterate forward here.

* We now have https://github.com/ipld/go-trustless-utils and I think it might be a good place to start toying with schema options and some basic implementation around the metadata chunk so we can get it right and start experimenting.
* I've already expressed a mild preference for dag-json rather than cbor for this trailing chunk--it's human readable which might help in the case that someone downloads this as a CAR that doesn't parse like a normal CAR so they go inspecting it.
* I defer to @lidel re content-type, seems reasonable that we don't introduce a new one that we're shouldered with forever that only does this one thing and not all the other things we want to do for a CARv3 (but still, let's get CARv3 done this year perhaps?).
* Lassie isn't really designed like a `curl` that'll pass on the stream, mainly because it's a multi-protocol client and has dreams of being much more complex in its ability to slice and dice sub-retrievals for a whole unit. But, there are reusable parts and it may just be better for SPARK to reuse the [http retriever](https://github.com/filecoin-project/lassie/blob/main/pkg/retriever/httpretriever.go) which mostly does what you want since you (a) only care about HTTP and (b) know who you're fetching from so you don't need the lookup piece (I believe?). We can work on making that more reusable as a separate component--let's discuss requirements here.
* I'm going to go work on content-type / accept handling for Frisbii / Lassie -- currently we're only expecting and asserting a single one and the comma-separated option list has been a TODO, for various reasons we're reaching the end of that as a viable approach and the need arising from this issue makes it more important.

src/http-gateways/trustless-gateway.md
lidel, line 308: :thought_balloon: since the metadata is now outside and after the CAR stream, and @rvagg noted preference for JSON over CBOR, would it be good time to rename this to more self-explanatory `eof=json`? This way we could also support `eof=nil` (removing the need for sending  `{}` when metadata is not needed)

src/ipips/ipip-0431.md
```diff
@@ -0,0 +1,145 @@
...
+## Motivation
+
+SPARK is a Filecoin Station module that measures the reputation of Storage Providers by periodically
+retrieving a random CID. Since both SPs and SPARK nodes are permissionless, and Proof of Retrieval
+is an unsolved problem, we need a way to verify that a SPARK node retrieved the given CID from the
+given SP. To enable that, we need the Trustless Gateway serving the retrieval request to include a
+retrieval attestation after the entire response was sent to the client.
+
+Aside from this specific use case, the IPFS Ecosystem at large has no reliable
+mechanism to signal that a CAR file transmission over HTTP completed successfully.
```
alanshaw, line 32: Can you elaborate?

We already have a multicodec for CAR. Couldn't you retrieve a CAR file from a gateway by CID e.g. https://w3s.link/ipfs/bagbaierabxhdw7wglmlehzgobjuoq3v3bdv64iagjdhu74ysjvdecrezxldq - you don't need to signal successful transmission if the content hashes to the same CID.

If you're using the graph API (`?format=car` or `Accept: application/vnd.ipld.car`) you're being specific about what you want in the request and the client is verifying the blocks...and so they know if the transmission over HTTP completed successfully or not...right?

willscott: I think there are a couple rough edges that motivated the desire to have additional signaling / metadata here:

1. In a Filecoin deal or car uploaded that has links to blocks that don't exist. How do you differentiate if a requested car has all the blocks that it should have, or if a missing link in the traversal was skipped / incorrectly transmitted?
2. If the client doesn't have all the codecs of all blocks, so can't parse links in blocks / "follow" the traversal or structure of the car it has asked for, how does it know it got the right blocks?

src/http-gateways/trustless-gateway.md
bajtos: How about `meta=json`?

bajtos: > `eof=nil` at some point (removing the need for sending  `{}` when metadata is not needed)

FWIW, if we don't need metadata, then we can omit the entire `meta` (or `eof`) parameter.

patrickwoodhead: @bajtos what is your reasoning for preferring `meta` over `eof`?

bajtos: I interpret `meta=json` as "CAR blocks with additional metadata in JSON format". It's open to extension, e.g. allowing metadata blocks in more places than just EOF.

In one of the discussions (sadly not captured in writing), somebody was objecting against using `eof` in `meta=eof` and suggesting to use just `meta` key with no value. I don't remember exact reasons.

On the second thought, considering that we are looking for a short-term fix with least resistance, maybe the key name (`meta` vs `eof`) doesn't matter that much.

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,31 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+- The payload contains metadata encoded as DAG-CBOR.
+
+The metadata MUST include the following fields:
+
+- `len` - byte length of the CAR data (excluding the metadata block)
+- `b3h` - Blake3 hash (checksum) of the CAR data (excluding the metadata block).
+- `b3h_sig` - A signature over `<len><b3h><request>` using server's Ed2559 identity.
+  - `len` is encoded as `varint`,
+  - `b3h` is encoded as 32 bytes,
+  - The effective query as executed by the gateway. This query is the request url - path and query string arguments.
```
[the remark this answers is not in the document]
patrickwoodhead: @lidel what is the reasoning behind splitting up `dag_params` and `car_params`? Could we instead go for `content_path` and `query_params` to keep it simple and generic (allowing for other query params)?

src/http-gateways/trustless-gateway.md
patrickwoodhead: I like having `json` or `cbor` or other being the value since it makes the metadata options extensible. This means we can start with dag-json as per @rvagg 's suggestion.  For the key, I think that both `eof` and `meta` are not quite right to describe this. Metadata could be placed elsewhere and eof doesnt clear signal that this is metadata. What about `eofmeta=json`?

src/http-gateways/trustless-gateway.md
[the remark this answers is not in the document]
patrickwoodhead: I agree that the Spark use case belongs in Userland section. However, the individual keys of the metadata section, and what the client must do to implement them, feels like something that should be in the trustless gateway spec. 

Keys like `car_bytes`, `data_bytes`, `block_count` will be used by Spark but also may be used by others, and the definition of a key must be the same for each use case. E.g. if one client sets `data_bytes` to be the total byte length of blocks and another sets it to be the total byte length of the CAR stream then trustless gateway implementers will need to implement different logic for each use case.

What's more, for the Spark use case, we do not want gateway operators to know that they are serving a Spark request and not some other request. Since the Spark ones will be incentivised and other request may not be, servers can simply provide a good retrieval service to Spark clients and a poor service to other clients.

`car_bytes`, `data_bytes`, `block_count` seem generic enough. The troublesome one is then the Blake3Hash and signature.

src/http-gateways/trustless-gateway.md
bajtos: > What about `eofmeta=json`?

Sounds good to me.

Maybe instead of combining `eof` and `meta`, we can use the name `trailer`? E.g. `trailer=json`

src/ipips/ipip-0431.md
lidel: > We already have a multicodec for CAR. Couldn't you retrieve a CAR file from a gateway by CID

Where do we get the "CAR CID" from? 
It does not exist as a concept anywhere in the specs related to retrieval or routing.

AFAIK in the majority of real world use cases:
- client does not know the "CAR CID" because CARv1 is not deterministic ([example](https://specs.ipfs.tech/http-gateways/trustless-gateway/#car-roots), [example](https://specs.ipfs.tech/http-gateways/trustless-gateway/#car-format-parameters-and-determinism)), it only knows he CID of actual data
- server does not know "CAR CID" because it is generating CAR stream in streaming fashion

> the client is verifying the blocks...and so they know if the transmission over HTTP completed successfully or not...right?

This works only if there is no HTTP midleware in-between client and server, which is never the case. There is always some HTTP middleware or CDN in production.

Once you are limited to HTTP semantics, you will cache truncated responses, and the client has to be smart enough to (1) detect that (2) be able to retry in a way that does not hit the same cache.

This is why it does not work in places like Rhea/Saturn, where HTTP responses are (last time i checked) cached blindly based only on HTTP semantics without understanding internal Block/DAG structure.

lidel:
Mind, you don't save much with `dag-cbor` if your keys and values are strings. 
Readability of `dag-json` is more valuable.

src/http-gateways/trustless-gateway.md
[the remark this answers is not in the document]
patrickwoodhead: Perhaps what we need to do is leave this IPIP to concern the metadata block being appended without any constraints on what can be included in it. Then in a separate place, we define a canonical way to include a key value   object in the metadata block and how the server should implement certain useful keys such as `car_bytes` et al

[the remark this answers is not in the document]
lidel: I think it would be ok to suggest a key name convention for generic things like  `car_bytes` `data_bytes` and `block_count` in the section that described `meta` parameter, as long it is
- scoped to JSON (perhaps list under explicit `meta=eof[+json]`?)
- change requirement from MUST to SHOULD (convention, not a hard requirement)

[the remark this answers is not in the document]
lidel: I believe that was a conscious design choice, to avoid mixing data selector with details of the transport format (not everything is in URL query params):

- `dag_params` are about what user data was selected, not tied to any specific transport (could be applied to something other than CAR)
   - these are things that land in URL query
- `car_params` are specific to CAR container format, they do not change the user data that was selected, only the way it is represented when sent as CAR  
   -  these are things that land in Accept/Content-Type headers

@patrickwoodhead  that being said, if you want to simplify, this IPIP could go with a single dag-json map named `response_params` (to avoid confusion with URL query params, and account for the fact that server may ignore some of request params when producing a response)

lidel:
@bajtos wiil you have time to update this IPIP to reflect the latest changes based on discussions so far?
iiuc they are:

- [ ] dropping custom codec requirement and moving metadata to the trailer after `0x00` (`<CARv1 stream> <0x00 byte> <metadata>`)
- [ ] switching metadata encoding from dag-cbor to dag-json
- [ ] documenting JSON keys that may be useful outside of SPARK (`car_bytes` `data_bytes`, `block_count`, `car_cid`?)
- [ ] decision if you want to rename from `meta=eof` to `meta=eof+json`, or leave as-is

patrickwoodhead:
> @bajtos wiil you have time to update this IPIP to reflect the latest changes based on discussions so far? iiuc they are:
> 
> * [ ]  dropping custom codec requirement and moving metadata to the trailer after `0x00` (`<CARv1 stream> <0x00 byte> <metadata>`)
> * [ ]  switching metadata encoding from dag-cbor to dag-json
> * [ ]  decision? documenting JSON keys that may be useful outside of SPARK (`car_bytes` `data_bytes`, `block_count`, `car_cid`?)
> * [ ]  decision? if you want to rename from `meta=eof` to `meta=eof+json`, or leave as-is

@lidel I will be taking on updating this based on discussion so far. Agreed on first two bullets. I think `meta=eof+json` works well. the last remaining point of ambiguity is how we document the useful JSON keys

rvagg:
At this stage I'm not sure I see much of real use other than a field for an error message. I mocked up a schema in my prototype version of this in https://github.com/filecoin-project/lassie/pull/378/files#diff-c0f5012f54c2bfa71efce83a8f158229509132b9a8548158e105df4847fa9f18 which I then went on to specify that the `"properties"` and `"error"` fields were both optional with the intention that it would be one or the other (perhaps I should have used some type of union to specify this).

I did make a `"request"` field mandatory though—echo back what the original request was, so you end up with a block that says "here's what I was asked for, and here's what I was able to provide for it"—in perfect circumstances, a provider given the same request and having access to the full DAG should produce identical information for both of these fields. The uniqueness would come from SPARK stuff.

src/http-gateways/trustless-gateway.md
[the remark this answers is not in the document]
bajtos: How about using `retrieval_params` instead of `response_params`? In my mind, response parameters don't describe "what user data was selected".

What is our motivation in SPARK: 
We want the gateway to describe what exactly the client is retrieving (CID, subpath, dag params, car params) and provide a signature over that.

If two SPARK checker clients submit a metadata block with the same retrieval parameters (CID, subpath, dag params, car params) then we want:
- To be able to verify that the clients retrieved the DAG (sub)tree they were expected to retrieve.
- Confidence that both clients made the same request to the gateway (semantically?) and were supposed to receive exactly the same response from the gateway.

> `content_path` - requested content path

When the client requests `GET /ipfs/bafy1234/cat.jpg`, what is `content_path`?
- `/ipfs/bafy1234/cat.jpg`
- `bafy1234/cat.jpg`
- `/cat.jpg`
- something else?

We need the metadata block to describe both the CID requested (`bafy1234`) and the resource subpath if specified (`/cat.jpg`).

src/http-gateways/trustless-gateway.md
bajtos: We are going forward with `meta=eof+json`

src/http-gateways/trustless-gateway.md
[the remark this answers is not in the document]
bajtos: > I think it would be also ok to have a documented convention for passing a hash of the CAR stream (aka CAR CID) – maybe name it `car_cid` and use CIDv1 with 0x0202 codec – this convention is already used by .storage folks, no need to invent anything new.

+1 to document `car_cid`.

For SPARK, we specifically want a Blake3 hash so that we can use inclusion proofs. That's why we want to use a dedicated field `b3checksum` instead of a more generic `car_cid`.

bajtos:
Hello folks; thank you for your patience! Together with @patrickwoodhead, we incorporated your feedback and updated both the proposal and the spec.

We are ready for the next round of reviews. 🙏🏻

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+      "type": "string"
+    },
+    "b3checksum": {
+      "description": "A Blake3 hash (checksum) of the CAR stream (excluding the 0x00 byte and the metadata block). The value should be serialized as a multihash with multibase prefix, preferably using Base58 encoding.",
+      "type": "string"
+    },
+    "content_path": {
+      "description": "The url path in the request as executed by the gateway, e.g. `/ipfs/bafy1234/cat.jpg`. The query string MUST BE stripped from the path.",
+      "type": "string"
+    },
```
bajtos, line 375: Discussion point leading back to https://github.com/ipfs/specs/pull/431#discussion_r1287474957:

How do we represent the information about what content was requested?
- The CID
- An optional path to a file inside UnixFS

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+When `meta=eof+json`, the JSON object SHOULD conform to the following [JSON schema](https://json-schema.org/).
+
+```json
+{
+  "type": "object",
+  "properties": {
+    "data": {
+      "type": "object",
+      "description": "Properties of the response"
+    },
```
bajtos, line 332: Discussion point:

In the current proposal, the top-level "data" object combines fields about "what was requested" (e.g. CAR & DAG params) with "what was returned" (e.g. CARv1 length in bytes).

I'd like to discuss an alternative: split `data` into two fields `req` and `res`. The first will describe what the client requested, the second will describe what the server returned.

Such division would allow us to shorten field names, e.g. `data.car_params.dup` can become `req.dups`.

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+<Response body as CARv1 stream> <0x00 byte> <Metadata>
+```
+
+The only supported value for the data type parameter is `json`. This signifies that the metadata MUST be a JSON object.
+
+This parameter MUST only be used with CAR `version=1`.
+
+When the parameter is not set or does not equal `eof+json`, the server SHOULD not add any extra blocks to the response, neither the 0x00 byte nor any metadata.
+
+When `meta=eof+json`, the JSON object SHOULD conform to the following [JSON schema](https://json-schema.org/).
```
bajtos, line 323: Discussion points:

- In the current spec & IPIP, we are formatting metadata as JSON. Should we say DAG-JSON instead?
- Do we want to serialise the metadata as a CAR block, prefixing the JSON data with `varint | CID` header?

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+
+The `meta` parameter allows clients to request the server to include additional metadata about the CAR along with the response body.
+
+The value of this parameter includes both the location where the metadata is given (e.g. `eof`) as well as the type of data received (e.g. `json`) separated by a `+`, to give a value such as `meta=eof+json`
+
+When the location parameter is set to `eof`, which is currently the only supported value, the server SHOULD respond with the following response body:
+
+```
+<Response body as CARv1 stream> <0x00 byte> <Metadata>
+```
```
bajtos, line 315: It looks like the template used by the website cannot handle `<>` characters correctly :( Is it something worth fixing or should you change this PR to use different characters?

E.g. follow the IPLD style:

```
Response body as CARv1 stream | 0x00 byt | Metadata
```

bajtos: Screenshot 2023-10-18 at 16 16 38

src/ipips/ipip-0431.md
```diff
@@ -0,0 +1,197 @@
...
+
+#### Using CBOR instead of JSON for the metadata block
+
+We could use CBOR instead of JSON for the metadata block. However it was [decided](https://github.com/ipfs/specs/pull/431#issuecomment-1719634928) to opt for user readibility over number of bytes since CBOR doesn't greatly reduce the number of bytes in a key value map compared with JSON.
+
+## Test fixtures
+
+TBD
+
+Using one CID, request the CAR data using various combinations of content type parameters.
```
bajtos, line 193: Flagging this TODO to show in the PR discussion.

src/ipips/ipip-0431.md
```diff
@@ -0,0 +1,197 @@
...
+#### Create CARv3
+
+We could admit we've clearly hit limitation of what we can do with HTTP and CARv1 and CARv2 and stop abusing existing CARv1 by mixing data plane with control plane.
+
+Spend energy on creating CARv3 that solves the problems from "Motivation" section and more:
+- optional index or key-value metadata before or after data
+- native truncation detection and standardized error handling and passing during streaming
+- support for things like [Large Blocks](https://discuss.ipfs.tech/t/supporting-large-ipld-blocks/15093/)
+
+TODO: link to some public artifact about CARv3
```
bajtos, line 179: Flagging this TODO to show in the PR discussion.

Any suggestions for the artefacts I can link to?

bajtos:
Previews of the current version:

- [IPIP-0431: Opt-in Extensible CAR Metadata on Trustless Gateway](https://bafybeidhqto2hrnf6ijpnrvewbdjhl25ppwrnz2y7tldccsqg2dezdegke.on.fleek.co/ipips/ipip-0431/)
- [Trustless Gateway Specification](https://bafybeidhqto2hrnf6ijpnrvewbdjhl25ppwrnz2y7tldccsqg2dezdegke.on.fleek.co/http-gateways/trustless-gateway/)

bajtos:
@lidel @rvagg @willscott Ping 👋🏻 What's the best way to move this proposal forward?

src/http-gateways/trustless-gateway.md
lidel: This is a bug in website generator, not blocking this PR. 
Fix tracked in https://github.com/ipfs/spec-generator/issues/19

src/ipips/ipip-0431.md
```diff
@@ -0,0 +1,197 @@
...
+bytes, which provides a backward-compatible solution for the [CARv1 streaming problem](https://github.com/ipfs/specs/pull/332) until new CAR version is introduced.
+
+- As another example, the metadata object includes the `error` field, allowing the server to pass back additional information about why the response is an error, such as why the CAR stream was incomplete.
+
+- In the SPARK use case, retrieval clients would like to prove they have retrieved an entire file from a specific retrieval provider that has implemented the trustless gateway spec. The additional metadata block allows checksums and signatures to be passed along with the data, allowing the retrieval client to create a proof of correct retrieval.
+
+- The metadata `sig` field SHOULD also be populated, returning a signature, using the server's Ed2559 identity, over the metadata properties object. This allows gateway clients to submit the metadata block as an attestation of retrieval that 3rd parties can verify.
+
+### Compatibility
```
lidel, line 97: @bajtos  Let's go extra mile here and elaborate what happens when CAR response with  `0x00`-prefixed suffix  is parsed by existing CAR software.

My suggestion is to add some clear statement about expected interop, like "libraries and implementations SHOULD ignore the suffix after 0x00", otherwise we will create a bad UX/DX, where developer tries to debug things with existign tooling and the tooling errors.

Personally, I imagine we don't want things to fail due to `0x00` suffix, namely:

- \>80% of Amino DHT IPFS network (including IPFS Desktop and Brave) is  Kubo
  - `ipfs dag import` should ignore suffix
- reference CAR libraries 
  - [js-car](https://github.com/ipld/js-car) (JS library used by things like custom Service Workers, Helia)
  - [go-car](https://github.com/ipld/go-car)  v1 and v2  (GO libraries)
- CLI tools we recommend to developers, they will try to use these for debugging CAR responses with the suffix:
  - [ipfs-car](https://www.npmjs.com/package/ipfs-car) (JS CLI) 
  - [car](https://github.com/ipld/go-car/tree/master/cmd/car#readme)  (CLI) 
  - [go-fixtureplate](https://github.com/ipld/go-fixtureplate) (CLI)

src/ipips/ipip-0431.md
lidel: @aschmahmann do we have anything on GH?

src/http-gateways/trustless-gateway.md
lidel: For now just use 
1```suggestion
```
[Response body as CARv1 stream] [0x00 byte] [Metadata]
```
````

File: src/http-gateways/trustless-gateway.md [hunk 13]
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+      "type": "integer"
+    },
+    "car_cid": {
+      "description": "A hash of the CAR stream giving a CIDv1 with 0x0202 codec",
+      "type": "string"
+    },
+    "b3checksum": {
+      "description": "A Blake3 hash (checksum) of the CAR stream (excluding the 0x00 byte and the metadata block). The value should be serialized as a multihash with multibase prefix, preferably using Base58 encoding.",
+      "type": "string"
+    },
```
lidel, line 371: nit: Hardcoding Blake3 in field name and description makes no sense if you use Multihash.
Rename to `car_checksum` ?

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+  "type": "object",
+  "properties": {
+    "car_bytes": {
+      "description": "The total byte length of the CAR stream (excluding the 0x00 byte and the metadata block)",
+      "type": "integer"
+    },
+    "data_bytes": {
+      "description": "Total byte length of the flat file before it was encoded into a CAR file",
+      "type": "integer"
+    },
```
lidel, line 359: @bajtos what happens when returned CAR is for:
- HAMT-sharded UnixFS directory?
- a single file under some sub-path of HAMT-sharded UnixFS directory?

Is the semantic meaning here to be "raw bytes of all files, ignoring UnixFS directory metadata", or something else?

src/http-gateways/trustless-gateway.md
lidel: Splitting into `req` and `res` sgtm, improves clarity

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +302,112 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+      "type": "object",
+      "description": "Properties of the response"
+    },
+    "error": {
+      "type": "string",
+      "description": "Error message"
+    },
+    "sig": {
+      "type": "string",
+      "description": "A signature, using the server's Ed2559 identity, over the `data` object serialized as JSON."
```
lidel, line 339: @bajtos 

- HTTP Gateways have no concept of "sever ED25519" introduced here. How one verifies the signature without knowing the pubkey?
  - One way to avoid being prescriptive about key type or its location, is to have `sig_key` with CID-encoded public  `libp2p-key` that can be used for signature verification.
    - The nice thing about this is that Gateway/client implementation will already have relevant code/library as we use these in IPNS and libp2p.
- If you sign JSON, you want it to be deterministic variant like DAG-JSON, otherwise someone will run into bugs when they use less strict JSON library in different languages.

```suggestion
   "sig_pubkey": {
      "type": "string",
      "description": "A libp2p-key used for signing"
    },
    "sig": {
      "type": "string",
      "description": "A signature, using the `sig_pubkey`, over the `data` object serialized as DAG-JSON."
```

src/http-gateways/trustless-gateway.md
lidel: @rvagg  thoughts? Value added in DAG-JSON prefixed with own CID is that it allows client to detect truncation beyond `0x00` byte.

bajtos: I believe clients can already easily detect truncation of the metadata block.
- The block is a DAG-JSON object, it must start with `{` and end with a matching `}`.
- If the block is truncated, it will not end with the matching `}` and the JSON parser will throw an error.

src/http-gateways/trustless-gateway.md
bajtos: I see that https://github.com/ipfs/spec-generator/issues/19 was fixed, I will stick to `<>`.

src/ipips/ipip-0431.md
bajtos: > Let's go extra mile here and elaborate what happens when CAR response with 0x00-prefixed suffix is parsed by existing CAR software.

It's a great idea to think about compatibility with existing & future tooling and clearly describe our thinking. 👍🏻 

The most important aspect is avoiding the "0x00 insertion attack" vector. You can find more details in the section [Zero-length-block insertion attacks](https://github.com/ipfs/specs/blob/b6069bf0d0d9ff57ebdba36cf57f0d2f079bef1f/src/ipips/ipip-0431.md#zero-length-block-insertion-attacks) (including the Filecoin-specific logic). I am cross-posting the mitigation I proposed:

>  Our proposal avoids this attack vector:
>  - It does not change the current semantics of CARv1. Zero-length blocks remain invalid.
>  - Instead, we treat the response body as a new container format combining the CARv1 file with additional data.
>  - Clients must explicitly request this new container format. Existing clients not aware of the new metadata will not receive responses in the new format.

When developers use existing tooling, they will never receive a CAR file with the `0x00` suffix.

There are two major ways how a CAR with a `0x00` suffix can emerge:

1. Somebody makes an HTTP request to a Trustless Gateway, explicitly asks to receive CAR with `meta=eof+json`, saves the response body to a `.car` file and forgets to extract the CAR payload from the container (remove the `\x00{metadata}` trailer).

2. Somebody uses a tool that is aware of `meta=eof+json`. The tool opts into this new feature when requesting content from a Trustless Gateway, but does not extract the CAR payload from the container in the response body before returning the content back to the user.

I am arguing that (2) is a bug in the tooling, introduced by the change that modified Trustless Gateway requests to opt-into `meta=eof+json`, and therefore, the maintainers of that tool should fix that bug - make the tool adhere to spec.

Regarding (1): do you think this will happen frequently enough to justify the effort required to change all libraries you mentioned to start ignoring the `0x00` byte?

Maybe it's actually a good thing that the tooling reports an error because it tells the user they are using the new `meta=eof+json` feature incorrectly.

As an alternative to silently stripping the `0x00` suffix, the tooling can detect the situation where `0x00` is followed by a valid DAG-JSON object and report a more helpful error message to the user, advising them to either change the "accept" header in the request to the Trustless Gateway or else remove the `0x00` suffix (unpack CARv1 from the container format).

Thoughts?

---

`go-car/cmd/car/inspect.go` seems to always treat `0x00` as EOF, if I am reading the source code correctly:

https://github.com/ipld/go-car/blob/5c5d432d582564f88fd2124f2fce4f2f3e47a654/cmd/car/inspect.go#L26
```go
	rd, err := carv2.NewReader(inStream, carv2.ZeroLengthSectionAsEOF(true))
```

`js-car` seems to always reject zero-length blocks:

https://github.com/ipld/js-car/blob/562c39266edda8422e471b7f83eadc8b7362ea0c/src/decoder.js#L94-L97
```js
  let length = decodeVarint(await reader.upTo(8), reader)
  if (length === 0) {
    throw new Error('Invalid CAR section (zero length)')
  }
```

I guess I can test how existing tooling handles zero-length blocks and document this behaviour in the IPIP, so that we better understand the current landscape.

src/http-gateways/trustless-gateway.md
bajtos: Here is your description of `car_cid`, see https://github.com/ipfs/specs/pull/431#discussion_r1327713175:

> I think it would be also ok to have a documented convention for passing a hash of the CAR stream (aka CAR CID) – maybe name it `car_cid` and use CIDv1 with 0x0202 codec – this convention is already used by .storage folks, no need to invent anything new.

Regarding `b3checksum`:

For SPARK, we specifically need the Blake3 hash of the CAR stream, and we need gateways to always return this hash. In particular, clients cannot ask the server to use Blake3 for the CAR checksum because the server could use this information to detect SPARK clients vs. other clients and provide different quality of service.

I agree it's confusing to have both `car_cid` and `b3checksum`, but I don't see a better solution. Do you?

src/http-gateways/trustless-gateway.md
bajtos: Great questions! TBH, I don't know the answers. We don't need `data_bytes` for SPARK. I think this field was added based on the discussion in this proposal, but I could not find the specific comment requesting it.

I am proposing to remove `data_field` from the spec. We can introduce it later if there is a clear need. We will better understand the desired semantics at that point.

src/http-gateways/trustless-gateway.md
bajtos: Here is our use case:
1. An untrusted/permissionless client makes a retrieval request to the Storage Provider's booster-http address advertised in IPNI.
2. The client submits the measurement to the SPARK orchestration layer.
3. Later, SPARK's evaluation service wants to verify that the client contacted the SP.

To do so, we must not accept signatures from any identity, only the signature from the identity advertised by SP to IPNI.

I am arguing this is true for everybody else who wants to use the signature to verify that a metadata block submitted by an untrusted party was indeed produced by the expected Trustless Gateway.

Consider a simple attack vector: the attacker takes the metadata block produced by the origin gateway and replaces the signature with one created using the attacker's identity. Clients verifying the signature against the `sig_pubkey` field in the metadata will not notice the attack.

Now I can see how including `sig_pubkey` can simplify troubleshooting:
- If `sig_pubkey` does not match the pubkey we expected, then we know the metadata block was signed by somebody else
- If `sig_pubkey` matches but the signature does not, then we know the metadata block was modified from the original.
Compare that with my proposal:
- If the signature is not valid, then either the metadata block was tampered with or it was signed by a different identity.

IMO, this improvement is not worth the cost of increasing metadata block size and, thus, egress traffic for Trustless Gateways.

Do you have any other use case for the signature in your mind?

IMO, the clients making retrieval requests don't need this signature for validating the metadata block, as they can rely on guarantees provided by the underlying transport - HTTPS.

> * HTTP Gateways have no concept of "server ED25519" introduced here. 

Good point. We don't require all Gateways to sign the metadata block, SPARK needs the signature only from Storage Providers' servers handling retrieval (`booster-http`).

Let's update the spec to explicitly mention the signature is an optional field.

> How one verifies the signature without knowing the pubkey?
>   * One way to avoid being prescriptive about key type or its location, is to have `sig_key` with CID-encoded public  `libp2p-key` that can be used for signature verification.

As I wrote above, if you don't know the expected server identity, then the signature is not useful for you.

Having said that, I like the idea of adding more details about the identity/public key to the spec. 

The proposed format _CID-encoded public `libp2p-key`_ seems like a good candidate, although AFAICT, that's not the format advertised to IPNI. In IPNI, I see identities in the format that can be used in multiaddr's `/p2p/{id}` part:

```
12D3KooWAWHEbCQy22d45mKbKSewoB1xksDDhR7o5S4mDrSNKXNk
12D3KooWAy5kaLtHf5uS7PZVLjSYd8sGqJ6fn7bxMjqLLZ1uULp9
12D3KooWEiPRcfjXJVehty8okJGJpBZP8zM5UBoCK5yw2MXfx98x
12D3KooWFpv7LP1MUmjfQ8sAUXgJXG5FRMJLnqnJyR32fVboqspB
12D3KooWHKeaNCnYByQUMS2n5PAZ1KZ9xKXqsb4bhpxVJ6bBJg5V
12D3KooWNHwmwNRkMEP6VqDCpjSZkqripoJgN7eWruvXXqC2kG9f
12D3KooWSfsqUahHLCmiENT8oN4FkVtz5pSCxKtNEb7wrR1rrRjk
```

> * If you sign JSON, you want it to be deterministic variant like DAG-JSON, otherwise someone will run into bugs when they use less strict JSON library in different languages.

Makes sense; I'll update the spec to require the metadata to be a DAG-JSON.

PR #434, opened 2023-08-29
feat(gateway): add specification for running the gateway over libp2p

### LINKED ISSUE #433: Gateway: define a protocol.ID for the gateway protocol.
As [libp2p+HTTP](https://github.com/libp2p/specs/pull/508) work advances, it would be good to define what the `protocol.ID` should be for the IPFS gateway protocol. This is just a string that serves as the canonical name for this protocol. It's useful for signaling support for IPFS gateway to peers. In libp2p+HTTP, this will be included in the `.well-known/libp2p` (the HTTP equivalent of identify), along with a path prefix metadata (this path prefix could be "/"). In the future I would expect something similar to maybe be included in peer records and IPNI metadata (there's already something similar in IPNI).

This can be as simple as `/ipfs-gateway`.

As an example, a libp2p+http node that supports `/ipfs-gateway` mounted at `/` would return this under `/.well-known/libp2p`:
```
$ curl -v localhost:51654/.well-known/libp2p 
*   Trying 127.0.0.1:51654...
* Connected to localhost (127.0.0.1) port 51654 (#0)
> GET /.well-known/libp2p HTTP/1.1
> Host: localhost:51654
> User-Agent: curl/8.0.1
> Accept: */*
> 
< HTTP/1.1 200 OK
< Content-Length: 30
< Content-Type: application/json
< Date: Wed, 16 Aug 2023 20:43:30 GMT
< 
{"/ipfs-gateway":{"path":"/"}}
```

This tells other peers that this node supports the IPFS gateway protocol and it's mounted on the root path (`"/"`). This system would also work if we wanted to mount it somewhere else as well.

This works exactly the same when run on top of a libp2p stream, except the HTTP/1.1 message is sent over a new libp2p stream (using protocol.ID `/http/1.1` as defined in the HTTP spec).

Just for completeness, not directly relevant to this, A peer advertises HTTP support by either advertising an HTTP transport in its multiaddr and/or advertising support for `/http/1.1` in libp2p's Identify protocol.

---

Once we agree on a canonical protocol.ID for the gateway protocol, I don't mind opening a PR.

aschmahmann:
closes #433 

Recommendations on wording and if more examples/specifics from the current draft of libp2p+http spec PR (https://github.com/libp2p/specs/pull/508) are desired are welcome.

Overall, this PR is just specifying that the protocol ID to use for the Gateway API is `/ipfs/gateway`

### DIFF
--- a/src/http-gateways/libp2p-gateway.md
+++ b/src/http-gateways/libp2p-gateway.md
@@ -2,8 +2,8 @@
 title: libp2p+HTTP Transport Gateway Specification
 description: >
   The libp2p Transport Gateway specification describes how HTTP Gateway semantics can be used over libp2p transports.
-date: 2023-08-29
-maturity: reliable
+date: 2023-10-04
+maturity: draft
 editors:
   - name: Adin Schmahmann
     github: aschmahmann
--- a/src/http-gateways/libp2p-gateway.md
+++ b/src/http-gateways/libp2p-gateway.md
@@ -1,24 +1,82 @@
 ---
-title: libp2p Transport Gateway Specification
+title: libp2p+HTTP Transport Gateway Specification
 description: >
-  The libp2p Transport Gateway specification describes how Gateway APIs can be used over libp2p transports.
+  The libp2p Transport Gateway specification describes how HTTP Gateway semantics can be used over libp2p transports.
 date: 2023-08-29
 maturity: reliable
 editors:
   - name: Adin Schmahmann
     github: aschmahmann
+    affiliation:
+        name: Protocol Labs
+        url: https://protocol.ai/
   - name: Marcin Rataj
     github: lidel
     url: https://lidel.org/
+    affiliation:
+        name: Protocol Labs
+        url: https://protocol.ai/
 xref:
   - path-gateway
   - trustless-gateway
-tags: ['httpGateways', 'lowLevelHttpGateways']
-order: 1
+tags: ['httpGateways', 'lowLevelHttpGateways', 'exchange', 'transport']
+order: 3
 ---
 
-The libp2p Transport Gateway specification describes how Gateway APIs can be used over [libp2p](https://github.com/libp2p/specs) transports.
+## Introduction
 
-libp2p has a [http+libp2p specification](https://github.com/libp2p/specs/pull/508) that describes how to use HTTP semantics over stream transports, as well as how to do discovery of what protocols are available (and where they are mounted), on a given endpoint. In order for a given HTTP protocol like the IPFS Gateway API to work in this environment it requires a protocol identifier.
+This specification describes how HTTP Gateway semantics
+and APIs can be used over [libp2p](https://github.com/libp2p/specs) transports.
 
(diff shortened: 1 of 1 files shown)


src/http-gateways/libp2p-gateway.md
```diff
@@ -0,0 +1,24 @@
...
+    github: lidel
+    url: https://lidel.org/
+xref:
+  - path-gateway
+  - trustless-gateway
+tags: ['httpGateways', 'lowLevelHttpGateways']
+order: 1
+---
+
+The libp2p Transport Gateway specification describes how Gateway APIs can be used over libp2p transports.
```
hacdias, line 20:
```suggestion
The libp2p Transport Gateway specification describes how Gateway APIs can be used over [libp2p](https://github.com/libp2p/specs) transports.
```

src/http-gateways/libp2p-gateway.md
```diff
@@ -0,0 +1,24 @@
...
+
+libp2p has has a [http+libp2p specification](https://github.com/libp2p/specs/pull/508) that describes how to use HTTP semantics over stream transports as well as how to do discovery of what protocols are available (and where they are mounted) on a given endpoint. In order for a given HTTP protocol like the IPFS Gateway API to work in this environment it requires a protocol identifier.
```
hacdias, line 22:
```suggestion
libp2p has a [http+libp2p specification](https://github.com/libp2p/specs/pull/508) that describes how to use HTTP semantics over stream transports, as well as how to do discovery of what protocols are available (and where they are mounted), on a given endpoint. In order for a given HTTP protocol like the IPFS Gateway API to work in this environment it requires a protocol identifier.
```

Removing the double `has`. Commas at your discretion.

src/http-gateways/libp2p-gateway.md
```diff
@@ -0,0 +1,24 @@
+---
+title: libp2p Transport Gateway Specification
+description: >
+  The libp2p Transport Gateway specification describes how Gateway APIs can be used over libp2p transports
```
hacdias, line 4:
```suggestion
  The libp2p Transport Gateway specification describes how Gateway APIs can be used over libp2p transports.
```

To be consistent with other specifications.

hacdias (approved):

lidel (approved):
Pushed cosmetic cleanup and added example of the mapping file which is expected for this to work.

@aschmahmann do you think we should merge it, or should this wait until we have a reference client in Kubo that leverages this, when available? (my vote is to merge, as hard to tell when we have time to work on client, probably after Istanbul, right?)

aschmahmann:
Yeah, I suspect an implementation in kubo will have to wait although we could probably make others (in either Go or JS) fairly easily given that we already have the tooling (we've had for blocks for a long time and for DAGs with the bifrost-gateway and saturn js-client work).

WDYT about merging this one but decreasing the spec maturity from "reliable" until it becomes more load bearing (i.e. there is at least one deployed and used pair of clients and servers that's relying on it)?

lidel:
I've switched to `draft` to set proper expectation while the ecosystem can start implementing and refining the spec in follow-up PRs, merging.

PR #435, opened 2023-09-04
gateway: document Retry-After header

lidel:
This is a cosmetic change to Gateway spec that 
- adds missing HTTP 502
  - this is used in projects like Saturn/Rhea for indicating content routing error for requested content path (making it distinct from 504 which is timeout during retrieval phase)
- adds missing guidance around `Retry-After` header, which should be leveraged for throttling (429) and in 502 and 504

(We don't need IPIP as this is filling a gap in existing specs to describe production behavior, not adding/changing anything new.)

### DIFF
--- a/src/http-gateways/path-gateway.md
+++ b/src/http-gateways/path-gateway.md
@@ -291,10 +291,9 @@ sent by the client.
 
 ### `429` Too Many Requests
 
-A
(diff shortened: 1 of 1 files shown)


willscott (approved):
having this defined as 'SHOULD' seems very reasonable.

hacdias (approved):
I think this is reasonable, and better for compliance.

PR #436, opened 2023-09-05
gateway: clean Cache-Control and Last-Modified

hacdias:
See https://github.com/ipfs/boxo/pull/459#pullrequestreview-1609790478. This is more of a clarification/cleanup and therefore requires no IPIP.

### DIFF
--- a/src/http-gateways/path-gateway.md
+++ b/src/http-gateways/path-gateway.md
@@ -380,7 +380,8 @@ Returned directive depends on requested content path and format:
   - Implementations MAY place an upper bound on any TTL received, as
     noted in Section 8 of :cite[rfc2181].
   - If TTL value is unknown, implementations SHOULD not send a `Cache-Control`
-    header and instead send a [`Last-Modified`](#last-modified-response-header) header.
+  - No matter if TTL value is known or not, implementations SHOULD always
+    send a [`Last-Modified`](#last-modified-response-header) header with the timestamp of the record resolution.
 
 ### `Last-Modified` (response header)
 
--- a/src/http-gateways/path-gateway.md
+++ b/src/http-gateways/path-gateway.md
@@ -370,17 +370,17 @@ in caches.
 
 Returned directive depends on requested content path and format:
 
-- `Cache-Control: public, max-age=29030400, immutable` must be returned for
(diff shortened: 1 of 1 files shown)


src/http-gateways/path-gateway.md
```diff
@@ -370,17 +370,17 @@ in caches.
...
+- `Cache-Control: public, max-age=<ttl>` SHOULD be returned for mutable
+  resources under `/ipns/{id-with-ttl}/` namespace; `max-age=<ttl>` SHOULD
+  indicate remaining TTL of the mutable pointer such as :cite[ipns-record] or DNSLink
   TXT record.
   - Implementations MAY place an upper bound on any TTL received, as
     noted in Section 8 of :cite[rfc2181].
-  - If TTL value is unknown, implementations SHOULD set it to a static
-    value, but it SHOULD not be lower than 60 seconds.
+  - If TTL value is unknown, implementations SHOULD not send a `Cache-Control`
+    header and instead send a [`Last-Modified`](#last-modified-response-header) header.
```
lidel, line 383: @hacdias we may want to KEEP `Last-Modified`

```suggestion
  - If TTL value is unknown, implementations SHOULD not send a `Cache-Control`
  - No matter if TTL value is known or not, implementations SHOULD always
    send a [`Last-Modified`](#last-modified-response-header) header with the timestamp of the record resolution.
```

Rationale:

>  RFC9110 prefers that servers send both ETag and Last-Modified for a 200 response if possible. During cache revalidation, if both If-Modified-Since and If-None-Match are present, then If-None-Match takes precedence for the validator. If you are only considering caching, you may think that Last-Modified is unnecessary. However, Last-Modified is not just useful for caching; it is a standard HTTP header that is also used by [..]  **systems to display the last-modified time, by crawlers to adjust crawl frequency**, and for other various purposes. So considering the overall HTTP ecosystem, it is better to provide both ETag and Last-Modified. – [HTTP Caching at MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching)

lidel (changes requested):

src/http-gateways/path-gateway.md
hacdias: Seems reasonable to me.

lidel (approved):
Thanks, lgtm.

PR #437, opened 2023-09-19
Publish existing Bitswap specs at specs.ipfs.tech

hacdias:
As per what I discussed with @lidel, this PR moves the current Bitswap specs to the website. Merge should be done using rebase or merge commit in order to preserve the history for the Bitswap specifications.

Note that I put this under `architecture`. We may want to either create a new section, or put it somewhere else.


lidel:
I've created `/exchange` category and moved Bitswap spec  to `/exchange/bitswap` (which will produce permalink  https://specs.ipfs.tech/exchange/bitswap/).

Preview: 
- https://bafybeibutkluyz3u2cm65zftu7cxtasbrtltz5s3hapsx2wehg2uxyrl5u.on.fleek.co/
  - https://bafybeibutkluyz3u2cm65zftu7cxtasbrtltz5s3hapsx2wehg2uxyrl5u.on.fleek.co/exchange/bitswap/

@aschmahmann  any concerns or ok to merge?

src/routing/index.html
```diff
@@ -1,8 +1,8 @@
 ---
 title: Routing
 description: |
-  Content routing is the way to determine where to find a given CID on the network;
-  specifically, which network peers provide specific CIDs. 
+  Routing is the way to determine where to find a given CID, Peer, or IPNS Record on the network;
+  specifically, which network peers provide specific CIDs.
```
aschmahmann, line 5: should delete this line if expanding from content routing to more general routing.

src/routing/index.html
```diff
@@ -1,8 +1,8 @@
 ---
 title: Routing
 description: |
-  Content routing is the way to determine where to find a given CID on the network;
-  specifically, which network peers provide specific CIDs. 
+  Routing is the way to determine where to find a given CID, Peer, or IPNS Record on the network;
```
aschmahmann, line 4:
```suggestion
  Routing is the way to determine where to find a given resource (e.g. Data, Peer, or IPNS record).
```

The above might not be quite right but my two notes here are:
1. It might be better to define the category more broadly and give examples. Although also fair to just describe exactly what's in there now and broaden as we add more
2. Technically we don't look for CIDs, we have CIDs and look for some of the data they refer to (e.g. the block itself, the full DAG underneath it, etc.)

Not blockers though

src/exchange/bitswap.md
```diff
@@ -0,0 +1,247 @@
...
+  }
+
+  Wantlist wantlist = 1;
+  repeated Block payload = 3;
+  repeated BlockPresence blockPresences = 4;
+  int32 pendingBytes = 5;
+}
+```
+
+## Implementations
```
aschmahmann, line 242: Can be done in a follow up, but noticed we're missing a Rust implementation here. Anyone know what implementations of Bitswap are best supported in Rust? cc @mxinden @b5

aschmahmann (changes requested):
Mostly LGTM, one small change requested (removing the unnecessary text after switching the section from content routing -> routing). The rest are optional.

Thanks for doing this 🙏

src/exchange/bitswap.md
mxinden: I know several, but I am not familiar enough with either to be able to endorse one.

aschmahmann: If there are multiple I think fine to include them. 

Just would prefer to not add in abandoned, non functional, etc. implementations where people will likely be better off not relying on them. People can still find those with a search engine if they want to build an implementation and see where others left off.

lidel: We can add them in future PR, right now I don't have bandwidth to evaluate how useful https://docs.rs/iroh-bitswap/0.2.0/iroh_bitswap/ is

src/routing/index.html
lidel: Makes sense, I've applied a shortened version of your suggestion.

aschmahmann (approved):

BigLep:
2023-10-17 maintainer conversation: @aschmahmann will approve and @lidel will merge

lidel (approved):

PR #438, opened 2023-09-19
chore: remove empty MFS spec

hacdias:
Simply removes the empty MFS spec file. It's not adding any information, and only links to an archive repository.


lidel (approved):
Thanks! This is effectively userland, ok with removing it.

Issue #444, opened 2023-10-09
WebSeed specifications


Jorropo:
This is a meta issue discussing required features to enable [webseed](https://www.bittorrent.org/beps/bep_0019.html) like features for the IPFS network.

A webseed is qualified as a server:
- Server that implements a version of the HTTP protocol. Or protocols with equivalent `GET` semantics. (like FTP)
- Is able to serve static files.
- Supports [RFC 7233](https://datatracker.ietf.org/doc/html/rfc7233) bytes HTTP `Range` request on such files. Or protocols with equivalent ranging semantics. (like FTP)
- Does not contain and do not run IPFS aware piece of code.
- And finally can be used by IPFS clients to download content without compromising on the IPFS features (content addressed incremental verification).

The endgame here would be to upload some file to such server which would contain both the merkle dag cryptographic proofs and an index which would allow clients to implement `O(log n)` seeking and random read behaviors.
Here for unixfs we have 2½ usecases which are useful to separate:
- 1\. IPFS proofs and index packed together with the leaves node. This mode would be blind to content of the blocks and could work with any IPFS application.
  Potential applications target IPFS native applications which want to leverage extensive HTTP compatible storage services which do not understand IPFS.
- 2\. IPFS proofs and index stored separately from the de-serialized leaves of the tree. This is usefull for applications which also want to serve their content over non IPFS means, this allows to reuse existing non IPFS architectures and have IPFS clients download the same files as non ipfs clients for the `raw-leaves` of the file.
- 2½. Similar as point 2 however it allows for using non webseeds (servers that run IPFS aware code) for downloading the proofs.
  This include cases where the proofs are packed into an archive and shared OOB (see `.torrent` files for example).

I can think of at least one other possible implementation but I am not including it here due to dubious value, could be added if this interest peoples.

# Milestones
## 2½
- [ ] `skip-leaves` trustless gateway option allowing to download from a gateway
- [ ] Figure out signaling and expected semantics of webseeds over content routers. 

## 2
- [ ] TODO

## 1
- [ ] TODO

lidel:
If we need a test data for proper webseeds (2), Wikipedia snapshots at https://download.kiwix.org/zim/wikipedia/ provide both good use case and variety (biggest archive is 96GiB).

The web-based ZIM reader described in https://github.com/ipfs/distributed-wikipedia-mirror/issues/140 could even utilize the MVP from (2½) – fetch proof from one gateway, and then range-requests for deserialized raw bytes from others.

PR #445, opened 2023-10-11
ipip(1337): add skip-leaves IPIP

Jorropo:
For #444


src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,89 @@
...
+
+Setting the default value of the `skip-leaves` parameter to `n` ensures
+backward compatibility with existing clients and systems that are unaware
+of this new flag.
+
+### Prevention of Amplification Attacks and Efficient Server Operation
+
+By utilizing the `raw` (`0x55`) codec servers can trivially determine whether
+to fetch or skip a block without having to learn any new information.
+Altho more limited and not able to handle unixfs file using dag-pb for their
```
rvagg, line 68:
```suggestion
Although more limited and not able to handle unixfs file using dag-pb for their
```

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,89 @@
...
+tags: ['ipips']
+---
+
+## Summary
+
+Introduce `skip-leaves` flag for the :cite[trustless-gateway].
+
+## Motivation
+
+Allow clients to read a stream which only contains the unixfs proofs.
```
rvagg, line 24: I'd like to see some details in here that address the non-unixfs impact of this, or at least make it clear that this is compatible with non-unixfs graphs (or not?) since our current trustless design only special-cases unixfs, it doesn't assume it; this is written with what sounds like an assumption and could be adjusted.

Jorropo: This is the `Motivation` field, I don't think it need to be an accurate depiction of all the possible usecases.
I don't have any non-unixfs motivations, could you give me an example of an other data-structure which can use raw cids for the bulk of their data in a way that would be interesting here ? (bottom heavy graphs, download candidates discovery for bitswap, ...)

rvagg:
I can see the rationale for this; although it's quite similar to `dag-scope=entity`, so I wouldn't mind a treatment in the documentation why `entity` can't properly cover the desired use case, and if it can't, why `dag-scope=skip-leaves` (or similar) wouldn't be a better approach?

I think we should have a conversation about evolution of trustless now that it's deployed and being actively used in a number of scopes. We have multiple implementations of trustless now, and this adds an extension that is not necessarily in the scope or interests of classes of users or servers, or even if it is, implementation may be delayed or deferred for various reasons. It might be helpful to have some method of feature sniffing built into the spec, perhaps with an `OPTIONS` pattern so that clients that know they are using newer, less widely deployed, features of trustless, can first sniff before getting a rejection. Alternatively, we could just do a reject with `415` and send a pre-defined header, such as `X-Supported-Content-Types`, that lists variants of `Accept` that won't error. If that were in the spec it would be easy to implement; I'd probably take existing `Accept` parsing that's lax and make it strict, because currently the 2 trustless server implementations I'm responsible for would ignore this new parameter and give you the whole thing; I wouldn't mind tightening that up but we need a better signalling mechanism.

src/ipips/ipip-0445.md
rvagg: sorry, I just tagged here because it's the first mention of `unixfs`, I just think discussion would be good so we're clear that this is interoperable with other types of graphs.

two things come to mind:

1. you could just be clear that this would still work with non-unixfs, it's entirely dependent on codec in the CID
2. there are some novel cases where you could _use_ this to provide alternate addressing of existing data. Perhaps you have a standard unixfs graph but it's deep and complicated, so just bundle it up in some custom dag-cbor, linking to all the `raw` blocks, and you have a custom kind of proof pointing to the same file data; i.e. this could unlock custom proof and/or addressing mechanisms .. maybe (not claiming it will, or even that I think this is worthwhile, just talking off the top of my head).

Jorropo:
> I can see the rationale for this; although it's quite similar to `dag-scope=entity`, so I wouldn't mind a treatment in the documentation why entity can't properly cover the desired use case, and if it can't, why `dag-scope=skip-leaves` (or similar) wouldn't be a better approach?

Because you need to be able to express a request that is both `dag-scope=entity&entity-bytes=42:1337` and has `skip-leaves` set. (/ `entity-bytes` implies `dag-scope=entity`, so can't do `dag-scope=skip-leaves&entity-bytes=42:1337`):

For example let's say I have a `.webm` file stored in deserialized form on a non IPFS HTTP server and I am implementing a client that downloads the proofs from a gateway and the leaves from the non IPFS server.
Even in this situation features like seeking and `ReadAt` needs to work if the user wants to seek in the video.

This is great feedback I'll add that to the docs later thx.

---

I agree on multiplicity of options that could become an issue and we need a way to do signaling. I think #425 is a good place to have this discussion.
Imo this IPIP is fine to do without #425 because it's generally useful, the implementation difficulty range from trivial to easy and we are still at a point where the number of implementations is low.
It is also not very hard for clients to workaround a server that does not implement this feature by doing block by block requests. In my webseed usecase this is less worst than it sounds since the bulk of the traffic wont go through the gateway and thus wont be done block by block.

src/ipips/ipip-0445.md
Jorropo: I see thx, I've updated the text, first sentence don't mention unixfs and I've moved it to the sentence where I link #444

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,105 @@
...
+  video.
+- A client is verifying an OOB transfered directory in deserialized form,
+  then `dag-scope=all` with  `skip-leaves=y` makes sense.
+
+### Alternatives
+
+An alternative approach would be to request blocks individually.
+However it adds extra round trips and more per HTTP request overhead
+and thus is undesireable.
+
```
rvagg, line 94:
```suggestion
1. An alternative approach would be to request blocks individually.
However it adds extra round trips and more per HTTP request overhead
and thus is undesireable.

2. An alternative implementation may be to either specify codecs to
include, or codecs to exclude. e.g. `exclude-codec=0x55` or
`include-codec=0x70` to achieve similar results. There may be
broader utility in such an approach, but the use-cases beyond that
proposed above for `skip-leaves` are not clear.
```

:shrug: this _is_ an option, is it better? not sure, I'm not even sure it'd be very different on the implementation level and may afford some other possibilities. It also clears up the "leaves" terminology, which is very unixfs+CIDv1 specific.

Jorropo: Intresting idea. I don't think the current API is unixfs specific. I hope `raw` blocks are children-less bytes in all formats that use them.
I guess you could have an ADL that parse the bytes node out of the file using cbor but then you are using IPLD wrong IMO, should really be a cbor cid then.

`exclude-codec` and `include-codec` are interesting however they complexity the server implementation.
I have an idea on how we could combine a small index and a CARv1 DFS body with offloading and zerocopy to make a 400Gbps gateway with extremely low CPU usage implementation and having user input surface complexify the index this situation because I don't just need to map ranges in the unixfs file to ranges in the car, I also need to mark and compute prefix sums for all the codecs in the car.
So given this might anoy me one day and I don't need it, I wouldn't do it unless someone else wants this feature right now.

Jorropo: Also one weird thing, if I have a dag that is `dag-json → dag-cbor → raw` and I exclude the `dag-cbor` codec, is the gateway expected to send me 1 or 2 blocks ? Because if the answer is 2 blocks then the gateway has to process every skipped blocks which means the gateway does work which does not reflect on the client. And if the answer is 1 then it does not follow least surprise principle.

hannahhoward:
All that said, I'm fundamentally a yes on this IPIP, with the non-blocking suggestion to change the parameter name to "proof-only" instead of "skip-leaves"

Jorropo:
> All that said, I'm fundamentally a yes on this IPIP, with the suggestion to change the parameter name to "proof-only" instead of "skip-leaves"

Many formats store data in non `raw` leaves, so this would be making it really unixfs but with `--raw-leaves` enabled specific.

What do people think about `skip-raw` instead, it's self descriptive and correct ?

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,105 @@
...
+        url: https://protocol.ai/
+relatedIssues:
+  - https://github.com/ipfs/specs/issues/444
+order: 445
+tags: ['ipips']
+---
+
+## Summary
+
+Introduce `skip-leaves` flag for the :cite[trustless-gateway].
```
lidel, line 20: _(warning: obligatory rename thread ;)_

A slight concern here is that **the description does not match what the name suggests** :upside_down_face: 
It is not skipping _all_ leaves, only `raw` ones. 

Either we really skip all leaves (and have spec define them clearly, e.g. nodes without children / only with link to parent) or rename this to reflect what is really happening (skipping  raw blocks).

:point_right: Would either of the below be acceptable?
- `skip-raw-blocks=y|n` 
- `skip-raw=y|n`

src/http-gateways/trustless-gateway.md
```diff
@@ -214,7 +214,7 @@ The Body hash MUST match the Multihash from the requested CID.
 
 A CAR stream for the requested
 [application/vnd.ipld.car](https://www.iana.org/assignments/media-types/application/vnd.ipld.car)
-content type (with optional `order` and `dups` params), path and optional
+content type (with optional `order`, `dups` and `skip-leaves` params), path and optional
```
lidel, line 217: :warning: nit: this does not belong to content type. It belongs to URL params.

Rationale:
-  `order` and `dups` are parameters of CAR response format. they do not control the scope of user data that is beign returned, only how the user data is represented in CAR format
- `skip-leaves` proposed here CHANGES the scope of user data returned in response (some blocks won't be returned, not even once), which makes it the same abstraction as `dag-scope` and `entity-bytes`

lidel (changes requested):
Thank you for submitting this IPIP @Jorropo
Quick initial feedback inline.

src/ipips/ipip-0445.md
lidel: (bit unprocessed thought) If we are framing this around DAG traversal, `exclude-codec=0x55` could be a sensible attempt at future-proofing, but we can always add it later.

Rationale:

The  `raw` codec provides an interoperable way to act as traversal stop: ensure the child block that was referred will not be used for any further traversal / deserialization, but returned as an opaque blob. Allows for creating non-UnixFS dags that are compatible with existing IPFS ecosystem.

If we add generic  `exclude-codec` then its semantic meaning should also control when the traversal stops: there could be a real world utility for excluding DAG branches based on their root codec, e.g. when they can't be retrieved with usual transports and require special handling. Creating a knob for increasing retrieval interop of data that is partially encoded in proprietary codecs, but now sure how real world need there is for this _right now_.  

One thing that comes to mind is Filecoin which iirc can't be pinned or fetched recursively with standard IPFS tools due to the use of unsupported codecs (cc @hsanjuan @aschmahmann  if i misremembered this, I think we've hit that problem a while ago).

Is there any other use case that could benefit from open-ended `exclude-codec` instead of a flag that skips raw blocks?

src/http-gateways/trustless-gateway.md
```diff
@@ -301,6 +301,26 @@ of their presence in the DAG or the value assigned to the "dups" parameter, as
...
+A gateway MUST NOT assume this field is `y` if unspecified.
+When not specified it always MUST be understood as `n`.
+
+:::note Notes for implementers
+
+A request which is rooted at a `raw` block and has `skip-leaves=y` does not
+make sense and SHOULD NOT be sent by clients, it is fair for servers to
+error in this situation.
+
+:::
```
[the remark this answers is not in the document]
Jorropo: Something 4XX something.
My lazy side tell me that I don't want to do this because then I save on code and I can return an empty car in boxo gateway instead of implementing one more edge case.
I can do it if it feels important to someone but given sending a 200 and returning an empty car is fine outcome I'm not sure we need to do this.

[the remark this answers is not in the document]
rvagg: I'm not OK with empty CARs for the error case, and I don't particularly like that boxo/gateway is already doing this for other error cases and would rather not embed it as a standard practice here. `415` please.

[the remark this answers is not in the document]
Jorropo: IMO it's different from the other cases in boxo/gateway. If you request a raw block with skip-leaves, and empty car is a correct answer. It is the complete dag minues all the raw blocks, which happen to be an empty set but that is the client's problem and there is nothing the gateway can do about it. This is not a transient error either.

src/ipips/ipip-0445.md
Jorropo: Switched to `skip-raw-blocks`

src/http-gateways/trustless-gateway.md
lidel: Moved in 131b29d2c39c0f210d1f4449d9d416793350538c

src/http-gateways/trustless-gateway.md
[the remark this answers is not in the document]
lidel: I've updated this section in f96a92a5262fba7c9f788192aaea3f37ab0d8f06 and made HTTP 400 (Bad Request) a MUST behavior that we will also enforce via conformance test. 


(error is not 415 as we moved `skip-raw-blocks` from content type to URL query params – see "Alternatives" / "Why not CAR content type parameter ?" section  in IPIP).

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,176 @@
...
+    with raw-leaves option, then export DAG without `raw` blocks (use go-car's
+    [`filter`](https://github.com/ipld/go-car/tree/master/cmd/car#readme) or
+    similar)
+    - Why? This goes extra mile, but ensures every conformant gateway
+      implementation is not doing useless work of fetching raw blocks which are
+      not required for fulfilling `skip-raw-blocks=y` requests). We did
+      similar thing for `entity-bytes` and it was the only way we could show
+      bugs in Saturn project's cache implementation at the time.
+
+:::
```
lidel, line 172: @Jorropo this is the minimal set of tests I've identified for this IPIP, lmk if you think it is sufficient, or if we need more.

The way we did this in the past, was to update test fixtures section at the very end:
1. create reference implementation in Boxo & Kubo PRs
2. then creating a [gateway-conformance](https://github.com/ipfs/gateway-conformance) PR with relevant fixtures that pass against branches from (2)
3. after reference implementation and tests are merged in respective repos, we update IPIP with final CID or link to a CAR fixture (examples: [428](https://specs.ipfs.tech/ipips/ipip-0428/#test-fixtures), [412](https://specs.ipfs.tech/ipips/ipip-0412/#test-fixtures))

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,176 @@
...
+   remaining raw block byte ranges fetched from non-IPFS system match expected
+   CIDs.
+
+### Compatibility
+
+Setting the default value of the `skip-raw-blocks` parameter to `n` ensures
+backward compatibility with existing clients and systems that are unaware
+of this new flag.
+
+### Alternatives
```
lidel, line 86: :information_source: updated this section with paths not taken, lmk if anything requires more elaboration.

src/http-gateways/trustless-gateway.md
```diff
@@ -183,6 +183,22 @@ returned:
     returned to the client, the HTTP status code has already been sent to the
     client.
 
+### :dfn[skip-raw-blocks] (request query parameter)
```
lidel, line 186: :information_source:  `skip-raw-blocks` is now a URL query parameter 
(rationale explained in "Alternatives" → "Why not CAR content type parameter" section)

lidel (commented):
Pushed some editorial changes, namely rename from `skip-leaves` content type param to `skip-raw-blocks` URL query param.
(rationale explained in "Alternatives" → "Why not CAR content type parameter" section)

Ratification requires reference implementation and conformance tests (details inline) but it is good enough to mention during IPFS Implementers call this Thursday for final reviews from wider ecosystem.

src/ipips/ipip-0445.md
lidel: Ok,  silence suggests we don't have any immediate need for `exclude-codec`.
Let's keep this IPIP limited to `skip-raw-blocks`,  open-ended filtering can be proposed in future IPIP.

src/ipips/ipip-0445.md
```diff
@@ -0,0 +1,176 @@
...
+   clients can quickly discover numerous candidate blocks without being
+   bottlenecked by the gateway's transmission of raw blocks.
+
+4. **Non-IPFS HTTP Mirrors Become Useful:** Legacy data that is already exposed
+   over HTTP in deserialized form can now act as sources for specific block
+   byte ranges, without having to support any IPFS specific APIs. Plain HTTP
+   Range Requests can be used for fetching remaining raw block data, and the
+   metadata read via `skip-raw-blocks=y` is enough for a client to verify the
+   remaining raw block byte ranges fetched from non-IPFS system match expected
+   CIDs.
```
Jorropo, line 78: I was trying to hint at that in 1 Verification Flexibility, this text goes in much more detail and I think should be merged with the first entry.

kernelogic:
Would like to see this implemented as well. This can be useful to quickly inspect (and validate, for Fil+) the files in a sector. Without downloading the whole 32GB sector.

The current dag-scope=entity only covers the first level, currently will have to recursively drill down to get more information - involving a lot more requests.

hannahhoward:
One other thing to consider, though probably not in this IPIP -- option to not traverse files while recursively following a UnixFS Directory tree.

rvagg:
well, that could be included in here because it's quite similar - `skip=none|files|leaves`, being able to `ls -alR` and only get the directory entries would be pretty neat

PR #451, opened 2023-11-09
web-redirects: clarifications

### LINKED ISSUE #450: _redirects spec clarification
The _redirects file specification:
[https://github.com/ipfs/specs/blob/main/src/http-gateways/web-redirects-file.md](https://github.com/ipfs/specs/blob/main/src/http-gateways/web-redirects-file.md)

seems to me like it could become a bit more explicit about what is or is not acceptable about the file contents. Perhaps even using something BNF-ish. For example, here are some things that came up when I implemented parsing it:

* Lines that begin with whitespace are completely illegal? (I'm ignoring leading whitespace)
* Same for trailing whitespace aside from the optional carriage return (I'm ignoring all trailing ws)
* Blank lines are illegal? (I'm ignoring them)
* If the file does not end in a line feed, that's illegal? (I'm treating the last line as if it had a line ending)
* What whitespace is acceptable separating `from`, `to`, and/or `status`? (I'm being perhaps too flexible on this, but I would imagine the real answer is 1-or-more spaces and/or `\t`)
* "Rules MUST only be evaluated when hosted on a Subdomain or DNSLink Gateway, so that we have Same-Origin isolation." 
   -  I think what's intended here is that it should only be evaluated with same-origin isolation. 
   - ipfs:// URLs can do that, too, but if you just rearrange the sentence to rely on same-origin stuff, any future implementation can figure out if it applies to them
* Placeholder substitution isn't actually described, just given an example. 
   - My main question here is whether it's acceptable to use the same placeholder multiple times in the `to` expression. I went with yes.

lidel:
@John-LittleBearLabs this PR aims to apply clarifications for questions you've raised and close #450.
Lmk if it's good enough or if we should adjust something :pray:


src/http-gateways/web-redirects-file.md
```diff
@@ -75,7 +75,11 @@ For example:
 
 This rule will redirect a URL like `/posts/06/15/2022/hello-world` to `/articles/2022/06/15/hello-world`.
 
-### Splat
+Implementation MUST error when the same placeholder name is used more than once in `from`.
```
John-LittleBearLabs, line 78: I think I violate this one currently. /:a/:a would I think substitute the latter into to

But that's confusing-at-best. Error makes more sense.

John-LittleBearLabs: I think this is saying that not only that directive becomes a 500 response... but that any request going through the file becomes a 500, right?

John-LittleBearLabs (approved):
Thanks!

src/http-gateways/web-redirects-file.md
lidel: Yes, a duplicate `:placholder` in `from` is kind of a syntax error, and we want to HTTP 500 when those happen, so website owner fixes the problem, and not relies on unspecified behavior.

Issue #368, opened 2023-01-19
Gateway/UnixFS: specify/unify  symlink handling


lidel:
(this is placeholder)

Ref. https://github.com/ipfs/kubo/issues?q=is%3Aissue+is%3Aopen+symlink

lidel:
Unblocked. #331 landed in August 2025, and the UnixFS spec now defines both [`Symlink`](https://specs.ipfs.tech/unixfs/#dag-pb-symlink) and [symlink path resolution](https://specs.ipfs.tech/unixfs/#dag-pb-symlink-path-resolution).

The gateway side is still unspecified, and the two implementations already disagree:

- **boxo**, so Kubo and Rainbow, serves the target path as the response body with `Content-Type: inode/symlink`, and does not follow it (`gateway/handler_unixfs_file.go`). A symlink is terminal during traversal, so `/ipfs/cid/link/file.txt` does not resolve.
- **`@helia/verified-fetch`**, which [ipfs/service-worker-gateway](https://github.com/ipfs/service-worker-gateway#readme) runs, returns an empty body. The UnixFS exporter maps `symlink` to a content function that yields nothing, so the target never reaches the caller.

Same CID, a target string from one gateway and zero bytes from the other.

So the [path gateway spec](https://specs.ipfs.tech/http-gateways/path-gateway/) needs to say three things: what the body is, which `Content-Type` to use, and whether pathing through a symlink is allowed, including what happens when the target escapes the content root. #284 is the concrete request for resolving them server-side, but that needs to be argued as an IPIP i think.

PR #453, opened 2023-11-12
[wip] Web Pathing Specification: initial outline with TODOs

### LINKED ISSUE #432: Gateway: specification for Web Pathing
This is a placeholder for writing a specification on how `/ipfs` and `/ipns` paths are intended to work on HTTP Gateways.

### Scope 

- Should be added to https://specs.ipfs.tech/http-gateways/
- This is not about UnixFS pathing. Web Pathing  is a way wider topic, and includes behaviors not specific to UnixFS.
- This is not only about HTTP URLs for Gateways, but will also inform behavior of native `ipfs://` and `ipns://` (e.g. Brave, ipfs-chromium), which is expected to be the same on both.

### Things the specs MUST cover

- [ ] non-ascii characters (percent-encoding of unicode and arbitrary binary data)
- [ ] UnixFS pathing
  - [ ] traversing HAMTs
  - [ ] traversing symlinks
- [ ] DAG-CBOR/JSON pathing
  - [ ] `/ipfs/cbor-cid/unixfs-file`
  - [ ] `/ipfs/unixfs-dir-cid/dag-cbor-file/cbor-field`  (boxo/gateway errors on  this ([spec→traversing-cbor notes](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor)), but we should specify behavior when someone wants to support this)
- [ ] handling non-ascii characters
  - [ ] edge case: handling filenames that already look percent-encoded https://github.com/ipfs/gateway-conformance/issues/115
- [ ] handling redundant slashes  `///`  (301 to resolved URL? `path.Clean`?)
- [ ] handling `.` and `..` (301 to resolved URL? `path.Clean`?)
- [ ] trailing slash `/` required for enumerable map-like entities (UnixFS dir, DAG-CBOR document?)
- [ ] length limit for entire path
- [ ] length limit for a path segment
- [ ]  :warning:   [gateway-conformance](https://github.com/ipfs/gateway-conformance) tests for all of the above  :warning:
 - we really, really need this to ensure uniform behavior across implementations and contexts such as gateways vs `ipfs://` in browsers

cc @hacdias @aschmahmann @rvagg @ribasushi 
Feel free to comment about edge cases youa re aware of,  or add  things to the list.

lidel:
The goal of this specification is to close #432 and define a subset of possible content paths that ensures compatibility with existing HTTP and Web Platform standards, and have clear MUSTs and SHOULDs that we can use when discussing implementation details of projects like [ipfs-chromium](https://github.com/little-bear-labs/ipfs-chromium/)'s [Intent to Prototype: Verifying IPFS client](https://groups.google.com/a/chromium.org/g/blink-dev/c/RPdRi1ZzQKU/m/H84ERvmHAQAJ).

Pushing an extremely early draft of the scope to get early feedback.

Everyone is invited to comment on the PR, focusing on TODOs, MUSTs and SHOULDs and suggest improvements, especially if something is missing :pray:

### DIFF
--- a/src/web-pathing.md
+++ b/src/web-pathing.md
@@ -1,0 +1,143 @@
+---
+title: Web Pathing Specification
+description: >
+  Specification defines a subset of possible content paths that ensures
+  compatibility with existing HTTP and Web Platform standards.
+date: 2023-11-12
+maturity: wip
+editors:
+  - name: Marcin Rataj
+    github: lidel
+    url: https://lidel.org/
+    affiliation:
+        name: Protocol Labs
+        url: https://protocol.ai/
+tags: ['architecture', 'httpGateways', 'webHttpGateways']
+---
+
+Web Pathing Specification defines a subset of possible content paths
+that ensures compatibility with existing HTTP and Web Platform standards.
+
+## Introduction
+
+TODO: Clearly explain why the specification exists, what is the problem solved here.
+
+This document specifies details of pathing for content paths that start with
+`/ipfs` and `/ipns` namespaces, and why a logical content root included in a
+content path can facilitate security isolation and relative pathing in web
+contexts.
+
+Specification includes guidance around aspects such as hash functions,
+multibases, CID versions, codecs, and how they impact implementation's ability
+to translate pathing into traversal of a DAG.
+
+The goal of this specification is to enable competing and interoperable
+implementations, all while ensuring seamless traversal of paths within the web
+ecosystem.
+
+## Specification
+
+TODO: Explain things in depth.
+The resulting specification should be detailed enough to allow competing,
+interoperable implementations.
+
+### TODO: things to cover
+
+- TODO: why it's called "web pathing": ensuring pathing is interoperable with how existing http and web platform works; covers both /ipfs and /ipns namespace semantics; defines logical content root CID that can be mapped to URL / root which enables subdomain/dnslink gateways and ipfs:// and ipns:// protocol handlers to load existing datasets, websites, and assets with relative pathing without the need for modifying them; 
+
+- TODO: how web pathing is applied to CLI Tools; path gateways; and origin contexts: subdomain/dnslink, ipfs:// ipns:// URIs
+
+- TODO: MUSTs, SHOULDs and MAYs in relation to
+
+  - TODO: multihash functions
+    - MUSTs
+      - `sha2-256` (`0x12`)
+      - `blake2b-256` (`0xb220`)
+      - `blake3` (`0x1e`)
+      - `identity` (`0x00`) (i.e. the data itself inlined in place of a hash)
+          - TODO: Identity CIDs MUST NOT generate network I/O such as bitswap, http request, since the data is always available in Multihash itself
+    - SHOULDs
+      - `sha2-384` (`0x20`, aka SHA-384; as specified by [FIPS 180-4](https://csrc.nist.gov/pubs/fips/180-4/upd1/final)) TODO: where is this used? why is this on the list?
+      - sha3-512 TODO: code for such label does not exist, a typo in prior notes? follow up required
+
+  - TODO: mutlibases
+    - MUSTs
+      * f - base16
+      * b - base32
+      * k - base36
+      * z - base58btc (case-sensitive!)
+      * u - base64url (case-sensitive!)
+    - SHOULDs
+      * F - base16 (uppercase)
+      * B - base32 (uppercase)
+      * K - base36 (uppercase)
+
+  - TODO: cid versions
+    - MUST:
+      - CIDv1 (`0x01`)
+      - CIDV0 (Multihash encoded with `base58btc`, with implicit dag-pb `0x70` codec)
+
+  - TODO: multicodecs that are required to facilitate path traversal
+    - DAG-PB
+    - RAW
+    - libp2p-key (for IPNS names)
+    - DAG-CBOR
+    - DAG-JSON
+
+- TODO: MUST support UnixFS pathing
+  - TODO: traversing HAMTs
+  - TODO: traversing symlinks
+  - TODO: make sure [UnixFS spec draft](https://github.com/ipfs/specs/pull/331) includes relevant descriptions, only refer to them from here, dont duplicate content
+
+- TODO MUST support DAG-CBOR/JSON pathing
+  - TODO `/ipfs/cbor-cid/unixfs-file`
+  - TODO `/ipfs/unixfs-dir-cid/dag-cbor-file/cbor-field`  (boxo/gateway errors on  this ([spec→traversing-cbor notes](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor)), but we should specify behavior when someone wants to support this)
+  - TODO make it clear if both DAG variants of CBOR and JSON are a MUST, or if JSON is a SHOULD (right now conformance tests require both as a MUST).
+
+- TODO: MUST what happens when we can't traverse part of the path
+  - TODO: separate errors for traversal errors due to missing codec vs missing content
+    - TODO: `/ipfs/valid-cid-dag-pb/invalid-path` (logical "not found", translates to HTTP 404 to indicate content does not exist, mention implicit http caching of 404 vs 500 – )
+    - TODO: `/ipfs/cid/unknown-codec-block/some/path` is requested (logical "path parser error", translates to  HTTP 500 error page due to missing decoder)
+
+- TODO: MUST describe handling of non-ascii characters
+  - TODO: dont invent anything new, refer to URL percent-encoding, like we did in [IPIP-383](https://github.com/ipfs/specs/pull/383)
+    - TODO: non-ascii characters (percent-encoding of unicode and arbitrary binary data)
+    - TODO: MUST: explicitly cover Unicode and that UTF-8 is implicit default
+    - TODO: have an answer for non-UTF-8 (e.g. UTF-16) code points (a MAY and error if are not supported? or error since this is web pathing, and web URL encoding uses UTF-8?)
+    - TODO: edge case: handling filenames that already look percent-encoded https://github.com/ipfs/gateway-conformance/issues/115
+      - TODO/TBD notes for implementers: mixing percent-encoded and raw paths is a very very comon case across the stack, writing down a sane MUST rule of thumb for implementers could improve resiliency across systems (e.g. if path includes `%` and produced 404, retry with percent-decoded value?)
+
+- TODO: path normalization
+  - TODO: note that paths are equivalent, but HTTP 301 SHOULD be used in HTTP context to ensure clients always end up on normalized paths
+  - TODO: handling redundant slashes  `///`  (301 to resolved URL? `path.Clean`?)
+  - TODO: handling `.` and `..` (301 to resolved URL? `path.Clean`?)
+  - TODO: trailing slash `/` required for enumerable map-like entities (UnixFS dir, DAG-CBOR document?)
+  - TODO: CID normalization (to canonical text respresentation version and multibase)
+    - /ipfs to CIDv1 in base32
+    - /ipns to CIDV1 with libp2p-key in base36
+
+### Test fixtures
+
+TODO: List relevant CIDs. Describe how implementations can use them to determine
+specification compliance.
+
+TODO: [gateway-conformance](https://github.com/ipfs/gateway-conformance) tests for all MUSTs in this spec
+This ensure uniform behavior across implementations and contexts such as gateways vs `ipfs://` in browsers
+
+### Security
+
+TODO: Explain the security implications/considerations relevant to the spec.
+
+TODO: length limit for entire path
+TODO: length limit for a path segment
+TODO: content path normalization should be performed before comparing paths
+TODO: mention how arbitrary content paths can be blocked via denylists defined in [IPIP-383](https://github.com/ipfs/specs/pull/383)
+
+### Privacy and User Control
+
+TODO: Note if there are any privacy or user control considerations that should be
+taken into account by the implementers.
+
+## Copyright
+
+Copyright and related rights waived via [CC0](https://creativecommons.org/publicdomain/zero/1.0/).


src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+implementations, all while ensuring seamless traversal of paths within the web
+ecosystem.
+
+## Specification
+
+TODO: Explain things in depth.
+The resulting specification should be detailed enough to allow competing,
+interoperable implementations.
+
+### TODO: things to cover
```
lidel, line 44: cc @Stebalien @dignifiedquire @hacdias @aschmahmann @Jorropo @rvagg @ribasushi @alanshaw @2color @autonome  @darobin for visibility and sourcing early feedback on the scope of this spec.

Feel free to drop a comment about any tricky/painful pathing edge cases you've encountered over the years that we should clarify web behavior for by including them in this spec :pray:

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+  - TODO: multihash functions
+    - MUSTs
+      - `sha2-256` (`0x12`)
+      - `blake2b-256` (`0xb220`)
+      - `blake3` (`0x1e`)
+      - `identity` (`0x00`) (i.e. the data itself inlined in place of a hash)
+          - TODO: Identity CIDs MUST NOT generate network I/O such as bitswap, http request, since the data is always available in Multihash itself
+    - SHOULDs
+      - `sha2-384` (`0x20`, aka SHA-384; as specified by [FIPS 180-4](https://csrc.nist.gov/pubs/fips/180-4/upd1/final)) TODO: where is this used? why is this on the list?
+      - sha3-512 TODO: code for such label does not exist, a typo in prior notes? follow up required
```
lidel, line 61: @John-LittleBearLabs  these two were included in your draft for WICG proposal, do you remember the reason/source?

John-LittleBearLabs: What is meant by 'label'? 

I don't recall, no. sha2-384 doesn't ring a bell - perhaps it was one of the comments that's now deleted (hackmd doesn't seem to let me mark things as resolved/hidden). As for sha3-512... it was probably not a good source; I think what it was was I found someone somewhere was talking about future-proofing hashes and I looked for one of the recommendations that also was marked as permanent in [the table](https://github.com/multiformats/multicodec/blob/master/table.csv#L11).

I'm definitely open to this list being altered.

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Specification
+
+TODO: Explain things in depth.
+The resulting specification should be detailed enough to allow competing,
+interoperable implementations.
+
+### TODO: things to cover
+
+- TODO: why it's called "web pathing": ensuring pathing is interoperable with how existing http and web platform works; covers both /ipfs and /ipns namespace semantics; defines logical content root CID that can be mapped to URL / root which enables subdomain/dnslink gateways and ipfs:// and ipns:// protocol handlers to load existing datasets, websites, and assets with relative pathing without the need for modifying them;
```
John-LittleBearLabs, line 46: > why it's called "web pathing"

I'm curious about this myself. It doesn't strike me as being particularly web-specific, at least not immediately.

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+  - TODO: traversing symlinks
+  - TODO: make sure [UnixFS spec draft](https://github.com/ipfs/specs/pull/331) includes relevant descriptions, only refer to them from here, dont duplicate content
+
+- TODO MUST support DAG-CBOR/JSON pathing
+  - TODO `/ipfs/cbor-cid/unixfs-file`
+  - TODO `/ipfs/unixfs-dir-cid/dag-cbor-file/cbor-field`  (boxo/gateway errors on  this ([spec→traversing-cbor notes](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor)), but we should specify behavior when someone wants to support this)
+  - TODO make it clear if both DAG variants of CBOR and JSON are a MUST, or if JSON is a SHOULD (right now conformance tests require both as a MUST).
+
+- TODO: MUST what happens when we can't traverse part of the path
+  - TODO: separate errors for traversal errors due to missing codec vs missing content
```
John-LittleBearLabs, line 98: :+1:

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+  - TODO: CID normalization (to canonical text respresentation version and multibase)
+    - /ipfs to CIDv1 in base32
+    - /ipns to CIDV1 with libp2p-key in base36
+
+### Test fixtures
+
+TODO: List relevant CIDs. Describe how implementations can use them to determine
+specification compliance.
+
+TODO: [gateway-conformance](https://github.com/ipfs/gateway-conformance) tests for all MUSTs in this spec
```
John-LittleBearLabs, line 124: :+1: :+1:

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+  - TODO: multicodecs that are required to facilitate path traversal
+    - DAG-PB
+    - RAW
+    - libp2p-key (for IPNS names)
+    - DAG-CBOR
+    - DAG-JSON
+
+- TODO: MUST support UnixFS pathing
+  - TODO: traversing HAMTs
+  - TODO: traversing symlinks
```
John-LittleBearLabs, line 89: I have questions, and not sure if I should be instead commenting [here](https://github.com/ipfs/specs/issues/368) ?

There's a few basic forms I could imagine this working in, and they're not necessarily incompatible:

* /ipfs/cid1/a = "/ipfs/cid2/c" : /ipfs/cid1/a/b -> /ipfs/cid2/c/b 
  - Replace all left of and including current path element with link contents.
  - IIRC I believe the gateway conformance test has this, so I'm guessing this is the real thing.
  - Are we allowed to link to /ipns/ namespace?
  - If so... even DNSLink? The link would still be immutable, but fully resolved what looks like part of your tree now depends on your local DNS setup?
* /ipfs/cid1/a = "c" : /ipfs/cid1/a/b -> /ipfs/cid1/c/b (i.e. not starting with /)
  - Replace current path element with link contents.
  - I read someone talking about converting tar to car and if so there's an important special case...
  - "../b" : If allowed we might need rules about this.
* /ipfs/cid1/a/b = "/c" : /ipfs/cid1/a/b -> /ipfs/cid1/c
  - Replace current path element and everything between the root and current element.
  - Need rules about DAGs that contain a directory under root named either /ipfs/ or /ipns/ etc.
  - I don't love features that break DAG symmetry, but others seem to :shrug:

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+
+  - TODO: cid versions
+    - MUST:
+      - CIDv1 (`0x01`)
+      - CIDV0 (Multihash encoded with `base58btc`, with implicit dag-pb `0x70` codec)
+
+  - TODO: multicodecs that are required to facilitate path traversal
+    - DAG-PB
+    - RAW
+    - libp2p-key (for IPNS names)
```
John-LittleBearLabs, line 83: Is verfying an IPNS record outside the scope of this document? It's not exactly pathing, even if that's where it my codebase it happens to show up.

[This](https://specs.ipfs.tech/ipns/ipns-record/#key-types) makes me think there really are only 4 we care about, and 2 of them are MAY, and none of them are listed as permanent [here](https://github.com/multiformats/multicodec/blob/master/table.csv).

src/web-pathing.md
John-LittleBearLabs: Also, how does this interact with _redirects (since it both has to be in root and its redirects can be relative to the root)?
Site A has a _redirects with splat to /a.html
Site B has a symlink (called link) to a's root, and its own redirects splat to /b.html
`ipfs://B/link/notfound.html`
becomes what exactly?

In my current [PR](https://github.com/little-bear-labs/ipfs-chromium/pull/75) it would redirect to `ipfs://B/link/a.html` (e.g. it respects A's _redirects file, and does it relative to A's root). But if A did not have a redirect, it would be not found (e.g. B's _redirects is ignored).

Feels weird.

src/web-pathing.md
lidel: Good question. My initial idea is to refer to [IPNS spec](https://specs.ipfs.tech/ipns/ipns-record/#key-types) which states that only Ed25519 is a MUST (RSA is SHOULD, other key types are MAY).

src/web-pathing.md
lidel: @John-LittleBearLabs _(I've realized we've discussed this during one of sync calls but did not reply here)_

- symlinks are generally underspecified and not used much. I would mark this as unspecified behavior in this spec until we land https://github.com/ipfs/specs/pull/331 
- that being said, if you already implemented symlink support, its ok, only caveat is that following symlink should not allow for going beyond the content root (/ipfs/cid), so `/ipfs/cid1/a` pointing at `/ipfs/cid2/b` or `../cid2/b` must error 
- rules from `_redirects` are executed only when requested content path is missing within same origin (based on root CID).  in scenario you described you operate under origin B and it is not aware of _redirects from origin A (so _redirects is not executed)

src/web-pathing.md
bumblefudge: would URL-safe pathing or web-compatible pathing be more precise? it isn't pathing FOR or OF the web, but rather a web-compatible subset of the pathing currently possible with the tech to date, right?

src/web-pathing.md
SgtPooki: I think it's important to clarify how this differs, overlaps, or varies from pathing defined in https://specs.ipfs.tech/http-gateways/path-gateway, https://specs.ipfs.tech/http-gateways/trustless-gateway, & https://specs.ipfs.tech/http-gateways/subdomain-gateway.

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+
+### TODO: things to cover
+
+- TODO: why it's called "web pathing": ensuring pathing is interoperable with how existing http and web platform works; covers both /ipfs and /ipns namespace semantics; defines logical content root CID that can be mapped to URL / root which enables subdomain/dnslink gateways and ipfs:// and ipns:// protocol handlers to load existing datasets, websites, and assets with relative pathing without the need for modifying them; 
+
+- TODO: how web pathing is applied to CLI Tools; path gateways; and origin contexts: subdomain/dnslink, ipfs:// ipns:// URIs
+
+- TODO: MUSTs, SHOULDs and MAYs in relation to
+
+  - TODO: multihash functions
```
SgtPooki, line 52: Is the intention of this section to clarify baseline multihash & codecs that must be supported to provide content for libraries such as `@helia/verified-fetch`?

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+      * z - base58btc (case-sensitive!)
+      * u - base64url (case-sensitive!)
+    - SHOULDs
+      * F - base16 (uppercase)
+      * B - base32 (uppercase)
+      * K - base36 (uppercase)
+
+  - TODO: cid versions
+    - MUST:
+      - CIDv1 (`0x01`)
```
SgtPooki, line 77: if we MUST support CIDv1, we should call out the multibase/hash/codecs that aren't guaranteed to be supported by web-pathing spec implementers.

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+      - CIDV0 (Multihash encoded with `base58btc`, with implicit dag-pb `0x70` codec)
+
+  - TODO: multicodecs that are required to facilitate path traversal
+    - DAG-PB
+    - RAW
+    - libp2p-key (for IPNS names)
+    - DAG-CBOR
+    - DAG-JSON
```
SgtPooki, line 85: we should MUST raw JSON as well, or is the intent to use RAW for that?

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+specification compliance.
+
+TODO: [gateway-conformance](https://github.com/ipfs/gateway-conformance) tests for all MUSTs in this spec
+This ensure uniform behavior across implementations and contexts such as gateways vs `ipfs://` in browsers
+
+### Security
+
+TODO: Explain the security implications/considerations relevant to the spec.
+
+TODO: length limit for entire path
```
SgtPooki, line 131: Are we limiting to browser URLs, or do we want to support longer lengths? https://stackoverflow.com/a/417184/592760 is a really thorough answer talking about variants.

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+TODO: length limit for a path segment
```
SgtPooki, line 132: I don't think we should limit path segment lengths, but we should prevent `/` in path segments opposite of [IPLD pathing](https://ipld.io/docs/data-model/pathing/)

src/web-pathing.md
```diff
@@ -0,0 +1,143 @@
...
+  - TODO: make sure [UnixFS spec draft](https://github.com/ipfs/specs/pull/331) includes relevant descriptions, only refer to them from here, dont duplicate content
+
+- TODO MUST support DAG-CBOR/JSON pathing
+  - TODO `/ipfs/cbor-cid/unixfs-file`
+  - TODO `/ipfs/unixfs-dir-cid/dag-cbor-file/cbor-field`  (boxo/gateway errors on  this ([spec→traversing-cbor notes](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor)), but we should specify behavior when someone wants to support this)
+  - TODO make it clear if both DAG variants of CBOR and JSON are a MUST, or if JSON is a SHOULD (right now conformance tests require both as a MUST).
+
+- TODO: MUST what happens when we can't traverse part of the path
+  - TODO: separate errors for traversal errors due to missing codec vs missing content
+    - TODO: `/ipfs/valid-cid-dag-pb/invalid-path` (logical "not found", translates to HTTP 404 to indicate content does not exist, mention implicit http caching of 404 vs 500 – )
```
lidel, line 99: TODO: a browser/HTTP specific section with additional behaviors that are possible when HTTP redirects can be executed:
- [ ] checking NFC and NFC normalized filenames if not found (https://github.com/ipfs/specs/issues/457)
- [ ] checking _redirects in not found (https://specs.ipfs.tech/http-gateways/web-redirects-file/)
- [ ] index.html when requesting directory  with `Accept` that has `text/html` (this one does not belong here, it should live in gateway specs instead, but writing down here so we dont forget documenting it)

lidel:
Thanks @John-LittleBearLabs, @SgtPooki, @bumblefudge for the feedback here.

Closing this. It is a 2023 Protocol Labs era draft, and most of what it outlined has since landed in specs that did not exist when it was written:

- UnixFS pathing: https://specs.ipfs.tech/unixfs/
- gateway behavior: https://specs.ipfs.tech/http-gateways/
- `ipfs://` and `ipns://`: #544

Once all three have landed, anyone is welcome to open something more modern that references them, rather than reviving this outline. TBD, but one could argue there is no longer a case for a separate `/web-pathing/` document, which also answers @SgtPooki's overlap question and @bumblefudge's naming one.

Where the TODOs went:

- `.` and `..`, escaping past the content root, restricted names: [Relative Path Components](https://specs.ipfs.tech/unixfs/#relative-path-components), [Restricted Names](https://specs.ipfs.tech/unixfs/#restricted-names)
- HAMT and symlink traversal: [HAMTDirectory](https://specs.ipfs.tech/unixfs/#dag-pb-hamtdirectory-path-resolution), [Symlink](https://specs.ipfs.tech/unixfs/#dag-pb-symlink-path-resolution)
- DAG-JSON and DAG-CBOR pathing, and failure to traverse: [Traversing through DAG-JSON and DAG-CBOR](https://specs.ipfs.tech/http-gateways/path-gateway/#traversing-through-dag-json-and-dag-cbor), [Handling traversal errors](https://specs.ipfs.tech/http-gateways/path-gateway/#handling-traversal-errors)
- trailing slash and the 301 to a normalized directory URL: [Use in directory URL normalization](https://specs.ipfs.tech/http-gateways/path-gateway/#use-in-directory-url-normalization)
- CID normalization, case-insensitive multibase, origin isolation: [Subdomain Gateway](https://specs.ipfs.tech/http-gateways/subdomain-gateway/) and #544
- multibase baseline: https://specs.ipfs.tech/cid/

Two answers that outlived the draft: symlink pathing is still unsettled, tracked in #368, and IPNS key codecs are left per implementation on purpose in #544's `ipns-uri.md`, so post-quantum keys can land without a spec revision.

Four things from this outline are written down nowhere, so #432 stays open for them:

- which multihash functions an implementation must support, and that an `identity` CID must never trigger network I/O
- path escaping, which [UnixFS explicitly leaves undefined](https://specs.ipfs.tech/unixfs/#path-escaping); Unicode normalization forms are separate, in #457
- length limits for a whole path and for a single segment
- whether the `json` codec is pathable the way `dag-json` is

PR #455, opened 2023-11-20
chore: remove Kubo userland specifications

### LINKED ISSUE #439: Remove Kubo (userland) specifications
A few of the specifications in this repository are not part of the "IPFS" specifications. Some are simply the specification of what certain IPFS implementations do. I'm opening this issue because I want to remove those from this repository and move them to the correct places.

These are the ones I identified:

- https://github.com/ipfs/specs/blob/main/REPO.md seems to be `kubo` and https://github.com/ipfs/js-ipfs-repo specific. I think we should move it to `kubo`, as I don't think there's anything we can call an "IPFS repository".
- https://github.com/ipfs/specs/blob/main/REPO_FS.md, same as above, but for specific implementation.
- https://github.com/ipfs/specs/blob/main/KEYSTORE.md is `kubo` specific

Am I missing anything? cc @lidel

hacdias:
Closes #439.

See also https://github.com/ipfs/kubo/pull/10226.


lidel (approved):
Thank you for cleaning this up, merged the kubo one, this one lgtm as well.

PR #460, opened 2024-01-25
gateway: clarify entiry-bytes out of range

lidel:
This PR applies cosmetic spec clarification based on discussion that happened in  https://github.com/ipfs/boxo/pull/523 and was released in [gateway-conformance@v0.5](https://github.com/ipfs/gateway-conformance/releases/tag/v0.5.0) 

cc @rvagg @hannahhoward

### DIFF
--- a/src/http-gateways/trustless-gateway.md
+++ b/src/http-gateways/trustless-gateway.md
@@ -155,7 +155,7 @@ The following additional values are supported:
 A Gateway MUST augment the returned `Etag` based on the passed `entity-bytes`.
 
 A Gateway SHOULD return an HTTP 400 Bad Request error when the requested range
-cannot be parsed as valid offset positions.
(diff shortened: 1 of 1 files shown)


src/http-gateways/trustless-gateway.md
```diff
@@ -155,16 +155,16 @@ The following additional values are supported:
 A Gateway MUST augment the returned `Etag` based on the passed `entity-bytes`.
 
 A Gateway SHOULD return an HTTP 400 Bad Request error when the requested range
-cannot be parsed as valid offset positions.
+is outside of valid offset positions in full, and the gateway knows that upfront.
```
rvagg, line 158: soft suggestion, my brain tripped up on "in full" when reading this on the first pass.

```suggestion
is entirely outside of the entity's byte range and the server is able to determine this
upfront.
```

rvagg (approved):
sgtm

PR #462, opened 2024-02-16
IPIP-0461: Ipfs-Path-Affinity on Gateways

lidel:
## TLDR

Extends Gateway specs with optional  `Ipfs-Path-Affinity` request header.
That is all.

If header is present in request, gateway can leverage this optional hint to improve content routing.

## Background

Endpoints that implement https://specs.ipfs.tech/http-gateways/trustless-gateway/ may receive requests for a single block, or a CAR request sub-DAG of a biger tree.

Not every CID is announced today, some providers limit announcements to top-level root CIDs.
Over time, both clients and servers should get smarter about the concept of "affinity": when processing request for a content path that is deeper than a root CID, leverage parent segments as additional hint for content routing lookup.

cc https://github.com/ipfs/kubo/issues/8676 https://github.com/ipfs/kubo/issues/10251


src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Security
+
+The client is in control when the affinity information is sent in the header,
+and implementation SHOULD allow end user to disable it in context where parent
+content path information is considered sensitive information.
+
+### Alternatives
+
+N/A
```
aschmahmann, line 72: Might be worth explaining alternatives here like:
- Why not just an arbitrary identifier the user could use to establish a relationship between requests?
- Why only one  path affinity rather than more?

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+What this mean for ecosystem? It should adapt. Over time, both clients and
+servers should leverage the concept of "affinity".
+
+## Detailed design
+
+Introduce `Ipfs-Path-Affinity` HTTP request header to allow HTTP client to
+inform gateway about the context of block/CAR request.
+
+Client asking gateway for a block SHOULD provide a hint about the DAG the block
+belongs to, if such information is available.
```
aschmahmann, line 40: Is it only a strict IPLD DAG where we'd recommend this? It seems like you could plausibly do this for a set of related data that aren't explicitly linked via IPLD (e.g. a website that has HTML that loads jpegs from within the same or a different root DAG).

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+This IPIP adds gateway support for optional `Ipfs-Path-Affinity` HTTP request header.
+
+## Motivation
+
+Endpoints that implement :cite[trustless-gateway] may receive requests for a
+single block, or a CAR request sub-DAG of a biger tree.
+
+Not every CID is announced today, some providers limit announcements to
+top-level root CIDs due to time and cost.
```
aschmahmann, line 29: Might be worth adding a reminder here that every piece of data that is supposed to be able to be accessed independently should be advertised.

Want to remind people to not walk themselves into silly problems like having data only accessible when passing the affinity header and then being confused when they can't get the data when it's included in another DAG. (e.g. I have the IPFS logo inside the DAG for docs.ipfs.tech but not advertised, but then someone uses the CID for the logo on explore.ipld.io but can't get the data because the affinity is wrong).

aschmahmann (changes requested):

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+What this mean for ecosystem? It should adapt. Over time, both clients and
+servers should leverage the concept of "affinity".
+
+## Detailed design
+
+Introduce `Ipfs-Path-Affinity` HTTP request header to allow HTTP client to
+inform gateway about the context of block/CAR request.
```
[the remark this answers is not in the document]
Stebalien: IMO, it should be `/ipfs/...` (or `ipfs://cid`), not just a CID.

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+  efficiently, resume downloads faster.
+- Gateway operators are able to leverage the hint and save resources related to
+  provider lookup.
+- Content providers are able to implement smarter announcement mechanisms,
+  without worrying that internal blocks are not announced.
+
+### Compatibility
+
+This is an optional HTTP header which makes it backward-compatible with
+existing ecosystem of HTTP clients and IPGS Gateways.
```
2color, line 62:
```suggestion
existing ecosystem of HTTP clients and IPFS Gateways.
```

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+This IPIP adds gateway support for optional `Ipfs-Path-Affinity` HTTP request header.
+
+## Motivation
+
+Endpoints that implement :cite[trustless-gateway] may receive requests for a
+single block, or a CAR request sub-DAG of a biger tree.
+
+Not every CID is announced today, some providers limit announcements to
+top-level root CIDs due to time and cost.
+
```
2color, line 30:
```suggestion
The introduction of an optional `Ipfs-Path-Affinity` header can increase the success rate of the gateway retrieving the request block, especially if the requested block is not advertised. 
```

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+## Detailed design
+
+Introduce `Ipfs-Path-Affinity` HTTP request header to allow HTTP client to
+inform gateway about the context of block/CAR request.
+
+Client asking gateway for a block SHOULD provide a hint about the DAG the block
+belongs to, if such information is available.
+
+Gateway being unable to find providers for internal block should be
```
2color, line 42:
```suggestion
Gateways unable to find providers for internal block should be
```

OR 

```suggestion
A gateway unable to find providers for internal block should be
```

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+### User benefit
+
+When supported by both client and server:
+
+- Light clients are able to use trustless HTTP gateway endpoints more
+  efficiently, resume downloads faster.
+- Gateway operators are able to leverage the hint and save resources related to
+  provider lookup.
+- Content providers are able to implement smarter announcement mechanisms,
```
2color, line 56: What exactly do you mean by "smarter announcement mechanisms"? 
Is it just a matter of whether only roots or all CIDs are announced?

2color:
What would be the general expectation of a server when a client requests, for example, a binary leaf block that isn't announced, and the affinity header of an announced CID is passed?

A client requests `bafk..ufqy` and passes `bafy...nkuq` in the `Ipfs-Path-Affinity` header. 
image

src/ipips/ipip-0462.md
2color: > Every piece of content that needs to be independently addressable should be advertised.

Yes, that makes sense in theory, but if we go to the `tar.gz` example:

>So at the very least if you make a block-request for the middle of a tar.gz file (where no part of the file really needs to be addressed on its own) you should be able to find it even if the provider has only advertised the root of the file.

Here, the only independently addressable content is the tar.gz file and for range block-requests in the middle, you'd **need** to pass the affinity header to be able to fetch that. 

If we're on the same page thus far, what would be smarter announcement mechanisms/strategies? Is the idea for those to be codec-aware in the sense that you could tell the node to advertise all UnixFS Files and Directories?

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Motivation
+
+Endpoints that implement :cite[trustless-gateway] may receive requests for a
+single block, or a CAR request sub-DAG of a biger tree.
+
+Not every CID is announced today, some providers limit announcements to
+top-level root CIDs due to time and cost.
+
+What this mean for ecosystem? It should adapt. Over time, both clients and
```
SgtPooki, line 31:
```suggestion
What does this mean for the ecosystem? It should adapt. Over time, both clients and
```

src/ipips/ipip-0462.md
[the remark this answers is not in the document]
SgtPooki: I would much prefer an ipfs:// protocol instead of pathing, but i guess whatever the users is likely to have is better ux

src/ipips/ipip-0462.md
SgtPooki: This reminds me of some chats we've had about it being likely that a provider of "bafyFoo1" would likely also have "bafyFoo2"

src/ipips/ipip-0462.md
SgtPooki: Prefer the later

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+### User benefit
+
+When supported by both client and server:
+
+- Light clients are able to use trustless HTTP gateway endpoints more
+  efficiently, resume downloads faster.
+- Gateway operators are able to leverage the hint and save resources related to
+  provider lookup.
+- Content providers are able to implement smarter announcement mechanisms,
+  without worrying that internal blocks are not announced.
```
SgtPooki, line 57:
```suggestion
  without worrying that some blocks are not announced unintentionally.
```

Speaks to the foot-gun problem Adin has commented about?

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Compatibility
+
+This is an optional HTTP header which makes it backward-compatible with
+existing ecosystem of HTTP clients and IPGS Gateways.
+
+### Security
+
+The client is in control when the affinity information is sent in the header,
+and implementation SHOULD allow end user to disable it in context where parent
```
SgtPooki, line 67:
```suggestion
and implementations SHOULD allow an end user to disable it in context where parent
```

Or

```suggestion
and an implementation SHOULD allow an end user to disable it in context where parent
```

src/ipips/ipip-0462.md
```diff
@@ -0,0 +1,80 @@
...
+Endpoints that implement :cite[trustless-gateway] may receive requests for a
+single block, or a CAR request sub-DAG of a biger tree.
+
+Not every CID is announced today, some providers limit announcements to
+top-level root CIDs due to time and cost.
+
+What this mean for ecosystem? It should adapt. Over time, both clients and
+servers should leverage the concept of "affinity".
+
+## Detailed design
```
SgtPooki, line 34: I really feel like this is missing some guidance on what the values should or can be.

With existing wording, it's too vague and could result in significant client pain of having to implement different values for different providers.

Also: we should really speak to the how of implementing this server on the client and server sides: at least some best practices. E.g. how are we going to implement this in `@helia/verified-fetch` and rainbow?

src/ipips/ipip-0462.md
lidel: @2color  we have some ideas for smarter announcement mechanisms/strategies in https://github.com/ipfs/kubo/ issues/8676 and more actionable https://github.com/ipfs/kubo/issues/10365  (which has wip implementation).

ps. we have a concept of `entity`, it was introduced in [IPIP-402](https://specs.ipfs.tech/ipips/ipip-0402/), we can use it in discussions like this.

src/ipips/ipip-0462.md
lidel:
```suggestion
  without worrying that some internal blocks are not announced (intentionally or unintentionally).
```

src/ipips/ipip-0462.md
2color: It sounds to me like "smarter" in this case means "frugal but smart" in the sense that it involves potentially less announcements that are effective enough for routing to work.

src/ipips/ipip-0462.md
lidel: Reworded Motivation section in de0b231ba97fa8fd223ddc1b8627b48c84138f8b a bit, but I think we need to have a page about advertisement and content routing on https://docs.ipfs.tech rather than specs.

---- 

While I get what the logo example was supposed to illustrate, it won't be a real world scenario: even when only top dir is announced, and no files, whoever authored explore.ipld.io would have the logo cached locally, and once the DAG is published, the  affinity of /ipns/explore.ipld.io will be enough to read it, even if it is still not announced directly.

Saying "announce things you want to access independently" sounds good on paper but does not work for real world use cases such as HTTP Range requests for video streaming or resuming downloads of big files. 

Streaming 32GiB video is an example of something where we can mostly agree we should not announce every internal CID, even tho one could say those internal CIDs are "accessed independently" when I paused and resumed the player (generated new range request for specific internal CIDs).

IPFS clients should leverage knowledge of affinity path, and ask providers of the root CID of the parent entity (in this case, the root of UnixFS DAG, or a directory it is in) for the data.

src/ipips/ipip-0462.md
lidel: Clarified format in `trustless-gateway.md`, the gist is: you put the content path that you try to load via block or CAR request.

https://en-wikipedia--on--ipfs-org.ipns.inbrowser.dev/wiki/Books fetching necessary blocks from https://trustless-gateway.link it would have `encodeURIComponent('/ipfs/bafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze/wiki/Books')` value:

```
Ipfs-Path-Affinity: %2Fipfs%2Fbafybeiaysi4s6lnjev27ln5icwm6tueaw2vdykrtjkwiphwekaywqhcjze%2Fwiki%2FBooks
```

in every `?format=raw` request.

src/ipips/ipip-0462.md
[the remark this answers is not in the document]
lidel: Because UnixFS can have arbitrary bytes in labels, we have to percent-encode, so it does not matter much. 

Clarified format in `trustless-gateway.md` (35a5eedf181ac1036562d0ae7be6a2275556d404)

src/ipips/ipip-0462.md
lidel: The way field format is specified, these can be arbitrary content paths. It is up to the client to provide a meaningful hint.

Most of the time it will be the content path the client tries to load, but it could also be `/ipns/other-website`.

src/ipips/ipip-0462.md
lidel: Exactly, there are "frugal" things we could do on both client and server to announce less, but be as efficient at retrieval in real world usage patterns (website browsing, video streaming, download resume or parallel downloads etc).

src/ipips/ipip-0462.md
lidel: Thanks. Added answer for the first one in 61518c604b17716221f3215808f2da2114afb75e
The second one is no longer relevant, because I've updated `trustless-gateway.md` we allow more than one header to be sent in a request.

lidel:
- Addressed some feedback, added description of field format in `trustless-gateway.md`
- Opened https://github.com/ipfs/boxo/pull/592  with poc implementation for `boxo/gateway`

PR #466, opened 2024-03-07
/routing/v1: add missing HTTP cache headers

hacdias:
Request from https://github.com/ipfs/boxo/pull/584/.

### DIFF
--- a/src/routing/http-routing-v1.md
+++ b/src/routing/http-routing-v1.md
@@ -74,7 +74,7 @@ This API uses a standard version prefix in the path, such as `/v1/...`. If a bac
 #### Response Headers
 
 - `Content-Type`: the content type of this response, which MUST be `application/json` or `application/x-ndjson` (see [streaming](#streaming)).
-- `Last-Modified`: the timestamp of the resolution, allowing CDNs and load balancers to support inexpensive update checks via `If-Modified-Since`
+- `Last-Modified`: an HTTP-date timestamp ([RFC9110, Section 5.6.7](https://www.rfc-editor.org/rfc/rfc9110#section-5.6.7)) of the resolution, allowing HTTP proxies and CDNs to support inexpensive update checks via `If-Modified-Since`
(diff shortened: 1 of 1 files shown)


src/routing/http-routing-v1.md
```diff
@@ -71,6 +71,13 @@ This API uses a standard version prefix in the path, such as `/v1/...`. If a bac
 - `404` (Not Found): must be returned if no matching records are found.
 - `422` (Unprocessable Entity): request does not conform to schema or semantic constraints.
 
+#### Response Headers
+
+- `Content-Type`: the content type of this response, which MUST be `application/json` or `application/x-ndjson` (see [streaming](#streaming)).
+- `Last-Modified`: the timestamp of the resolution.
+- `Cache-Control: public, max-age={TTL}`: cache TTL returned with the response. When present, it SHOULD be short for responses that do whose resolution ended in no results (e.g. 15 seconds), and MAY be longer for responses that have results (e.g. 5 minutes).
```
SgtPooki, line 78:
```suggestion
- `Cache-Control: public, max-age={TTL}`: cache TTL returned with the response. When present, it SHOULD be short for responses whose resolution ended in no results (e.g. 15 seconds), and MAY be longer for responses that have results (e.g. 5 minutes).
```

src/routing/http-routing-v1.md
```diff
@@ -108,6 +115,13 @@ represented as a CIDv1 encoded with `libp2p-key` codec.
...
```
SgtPooki, line 122:
```suggestion
- `Cache-Control: public, max-age={TTL}`: cache TTL returned with the response. When present, it SHOULD be short for responses whose resolution ended in no results (e.g. 15 seconds), and MAY be longer for responses that have results (e.g. 5 minutes).
```

lidel (approved):
Thanks! Added some references to RFC and the rationale behind TTL.
Merging (we can refine later, but better to have HTTP caching headers in the spec).

PR #471, opened 2024-04-17
gateways: document Content-Location

hacdias:
Document `Content-Location` as part of the work in https://github.com/ipfs/boxo/pull/603.

Context: https://github.com/ipfs/boxo/issues/606

### DIFF
--- a/src/http-gateways/path-gateway.md
+++ b/src/http-gateways/path-gateway.md
@@ -499,6 +499,17 @@ Returned only when request was a [`Range`](#range-request-header) request.
 
 See Section 14.4 of :cite[rfc9110].
 
+### `Content-Location` (response header)
+
+Returned when a non-default content format has been negotiated with the
+[`Accept` header](#accept-request-header). The value of this field SHOULD include
+the URL of the resource with the `format` query parameter included, so that
+generic HTTP caches can store deserialized, CAR, and block responses separately.
+
+For example, a request to `/ipfs/{cid}` with `Accept: application/vnd.ipld.raw`
+SHOULD return a `Content-Location: /ipfs/{cid}?format=raw` header in order for
+this request to be able to be successfully cached.
+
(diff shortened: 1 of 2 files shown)


src/http-gateways/path-gateway.md
```diff
@@ -242,6 +242,16 @@ These are the equivalents:
...
 
+A Client SHOULD include the `format` query parameter in the request URL, in
+addition to the `Accept` header. This provides the best interoperability and
+ensures consistent HTTP cache behavior across various gateway implementations.
+
+A Gateway SHOULD include the
+[`Content-Location`](#content-location-response-header) header in the response when:
+- the request contains an `Accept` header specifying a well-known response
+  format, but the URL does not include the `format` query parameter
+- the `format` parameter is present, but does not match the format from `Accept`
```
lidel, line 253: @hacdias I've added these clarifications as SHOULDs, the goal here is to push everyone towards including explicit `?format` in requests – haves a lot of headache in debugging interop/caching problems. 

We've always been doing this internally  (including both `format` and `Accept`), this writes it down as suggested practice.

lidel (approved):
Lgtm, thank you for documenting this.

Added some additional content, if looks good, feel free to merge.
(no IPIP as this is documenting/filling spec gaps, and not changing existing specs)

src/http-gateways/path-gateway.md
aschmahmann: Is it problematic then that some information in accept headers cannot be expressed in the format parameters (e.g. dfs and duplicates for CARs)

lidel: Potentially, could cause issues at some point in the future when a single gateway supports something more than "the Kubo default" and clients actually request non-default.

To avoid cache override problems, we add `car-*` URL params in #472 and use in `Content-Location` in https://github.com/ipfs/boxo/pull/603

PR #472, opened 2024-04-17
gateway: car-* request query parameters

hacdias:
For https://github.com/ipfs/boxo/pull/604.


lidel (approved):
Thanks!

Rebased and clarified how they impact use of `Content-Location`.

For anyone following: 

- Documenting use of `Content-Location` in https://github.com/ipfs/specs/pull/471 allowed us to introduce these query parameters safely, in a way that is backward compatible (`Accept` header taking priority over `format` query parameter, `Content-Type` of response being the final source of truth for the client). 
- No IPIP as these were already present in `Accept` and `Content-Type`, we don't introduce anything new here, we document query version in specification to allow exposing them in `Content-Location`, to ensure unique cache key for requests with parameters other than the implicit defaults.

PR #476, opened 2024-06-06
http-routing: add "get closest peers" operation

achingbrain:
Adds a new HTTP endpoint that can be used to request records for the closest peers to a given key that the routing implementation knows about.

The use-case for this is browser nodes performing random walks to find peers that they can make a circuit relay reservation on, without having to be DHT clients to perform the walk which can be undesirable given all the connection/processing overhead that entails.

I've tried to avoid defining what "closest" means to leave it up to the routing implementation, only that the responses should be "closer" than the optional `closerThan` parameter which should mean the caller gets useful results for other scenarios.


src/routing/http-routing-v1.md
```diff
@@ -106,6 +106,60 @@ Each object in the `Providers` list is a record conforming to a schema, usually
 
 ## Peer Routing API
 
+### `GET /routing/v1/closest-peers/{peer-id}?[closerThan]&[count]`
```
lidel, line 109: @achingbrain I think we could prototype this in https://github.com/ipfs-shipyard/someguy and deploy to [https://delegated-ipfs.dev](https://docs.ipfs.tech/concepts/public-utilities/#delegated-routing). 

Priority-wise, and for the purpose of IPIP we will write for this at some point, does the utility here stops at finding Circuit Relay servers viable in browsers without being a DHT client? 



Would it be useful to have Amino-specific API where the entire hash space could be queried for closest value, not just PeerIDs?
I wonder if this be generalized into delegated `/routing/v1/closest/{cid}?[than]&[count]`, or be limited to peers.

achingbrain: > does the utility here stops at finding Circuit Relay servers viable in browsers without being a DHT client?

No, there are other uses - you may wish to use this to populate your routing tables, or to find peers that can host provider records.  It would be useful to perform these operations from environments that have limited resources or cannot open many concurrent connections, for example.

> Would it be useful to have API where the entire hash space could be queried for closest value, not just PeerIDs?

Yes, plus here the PeerIDs are represented as CIDs, so the above is something that can be performed.

To make this clearer it might be better specced as `{cid}` rather than `{peer-id}` I guess.

src/routing/http-routing-v1.md
```diff
@@ -106,6 +106,60 @@ Each object in the `Providers` list is a record conforming to a schema, usually
...
+
+#### Path Parameters
+
+- `peer-id` is a [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) represented as a CIDv1 encoded with `libp2p-key` codec.
+
+#### Query Paramters
+
+- `closerThan` is an optional [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) represented as a CIDv1 encoded with `libp2p-key` codec.
+  - Returned peer records must be closer to `peer-id` than `closerThan`.
+  - If omitted the routing implementation should use it's own [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
```
guillaumemichel, line 119: If the query is expected to return the absolute closest peers to the target CID, the only use case I can see for `closerThan` is to find peers that are under eclipse attacks, but there are no current plans to implement it. Otherwise it will just limit the number of returned peers.

If there are other reasons to keep this parameter, I would recommend against using the receiver's (routing implementation) own peer id, when no value is provided. The default value should be no filter at all, return all the `count` closest peers. Otherwise, if a peer is looking for the `20` closest peers to some CID, and it reaches the 5th closest peer, this one will respond only with a list of 4 peers (because all the other are further away compare to the 5th closest peer). So in order to get the 20 closest peers, the requester would have to come up with a `closerThan` parameter that is _far enough_ so that at least 20 peers are closer to the target CID.

src/routing/http-routing-v1.md
```diff
@@ -106,6 +106,60 @@ Each object in the `Providers` list is a record conforming to a schema, usually
...
+- `count` is an optional number that specifies how many peer records the requester desires.
+  - Minimum 1, maximum 100, default 20.
```
guillaumemichel, line 121: If the `closerThan` parameter is set, and `count` is omitted, the default value should be the `maximum`, otherwise some peers closer than `closerThan` may be filtered out. Both parameters could however be combined if both are set.

src/routing/http-routing-v1.md
aschmahmann, line 109: I'm not sure this makes sense at the `/routing/v1` namespace since it's fairly specific to systems like the Amino DHT.

Having an HTTP API for this kind of thing makes sense though, perhaps as `/amino` or `/kad` or something. It could be nested under `/routing/v1/<amino, etc.>` as well if that makes folks happier.
- Note: Given that this is quite close to being the HTTP equivalent of the `FIND_NODE` RPC something like this could potentially end up being worked into that protocol itself (e.g. to enable http-based request/response).

aschmahmann: > does the utility here stops at finding Circuit Relay servers

Note: In many cases you'd want relay servers you'd also want peer routing (i.e. finding your address based on your peerID) which in the case of Amino means having recently pinged the closest peers to your peerID in XOR space. Ideally one of those (20) nodes is also acting as your relay which means the number of connections and queries you'd need are even smaller.

aschmahmann (changes requested):

src/routing/http-routing-v1.md
achingbrain: I think my assumption here is that the peers would be sorted by closeness value before the limit was applied.

Would this not be the case?

src/routing/http-routing-v1.md
achingbrain: My hope was that we could have it be routing-agnostic but maybe that's not possible, since the actual results would very much depend on what was being queried and how.

I'm happy with:

`GET /routing/v1/amino/closest-peers/{peer-id}?[closerThan]&[count]`

though if it's all amino, perhaps we should just have done with it and use the RPC names?

- `GET /routing/v1/amino/find-node/{multibase}`

...then maybe later:

- `GET /routing/v1/amino/get-value/{multibase}`
- `GET /routing/v1/amino/get-providers/{cid}`

though the second is redundant given the existing HTTP endpoints and the first is only partially redundant since we could use it to search for more than just IPNS records.

...and even later `PUT_VALUE` and `ADD_PROVIDER`?

src/routing/http-routing-v1.md
guillaumemichel: > I think my assumption here is that the peers would be sorted by closeness value before the limit was applied.

Yes that is correct. My point is that both `closerThan` and `count` can limit the count of peers that are returned, the parameters may conflict with each other.

E.g `peer-id=key0&closerThan=key1` should return all the peers between `key0` and `key1`, but if there are more than `default=20` peers between `key0` and `key1` this query will only return the `default` closest, and not all of them up to the `maximum`.

If none of `count` or `closerThan` parameters is set, the default `count` should be `default=20`. However, if `closerThan` is set and not `count`, the default `count` should be set to the `maximum=100`. This allows to return as many peers as possible between `key0` and `key1`, and if the caller wants to set a limit, they should combine `closerThan` and `count`.

Also, when setting `count` and not `closerThan` using the node's peer id as default `closerThan` parameter can be problematic. E.g I want to provide a CID to the 20 closest nodes, and I am the 12th closest peer to this CID. I will receive only 11 closer nodes, and will be unable to find the `20` closest nodes. Hence, if `count` is set and not `closerThan` the `closerThan` default should be the opposite of the target `peer-id` (`peer-id XOR 111...11`). Maybe the opposite of the target `peer-id` is a better default for `closerThan` than the requester's peer id, allowing the default response to contain a constant number of peers.

src/routing/http-routing-v1.md
achingbrain: In this case yes, they can query for their own peer id as a CID to get better results.

SgtPooki:
Just connecting some dots. This problem this solves seems somewhat related to https://github.com/libp2p/specs/issues/222

src/routing/http-routing-v1.md
achingbrain: @aschmahmann what do you think?  If we're going to expose "amino" endpoints, should they reflect the actual KAD-DHT RPC methods?

guillaumemichel: Agree with @aschmahmann, `/routing/v1/amino/` would make more sense, since the distance metric to define the _closeness_ is specific to kademlia.

---

IMO we shouldn't expose the actual KAD-DHT RPC methods, but the HTTP client should expect the HTTP server to do the DHT client's work.

e.g the `ADD_PROVIDER` RPC asks 1 DHT server to add the provided entry as provider. The API should expose a function where the HTTP server will find the 20 closest peers to the key (`FIND_NODE`), and then send them all a `ADD_PROVIDER` RPC.

The interface exposed should be the same as what is exposed by a kademlia implementation to its user, but not directly kademlia RPCs.

achingbrain:
Refs https://github.com/ipshipyard/roadmaps/issues/16

src/routing/http-routing-v1.md
lidel: I'm thinking about this via lens of `/routing/v1`, we already have `/routing/v1/ipns` namespace, which somehow  optional, so `/amino` proposal is not that controversial.

I guess to make sure the spec does what we want, next step here is a Someguy PR that exposes `/routing/v1/kad-dht/closest-peers/{peer-id}?[closerThan]&[count]` and then if we are happy with API we could deploy it to `delegated-ipfs.dev`.

(`kad-dht` because "dht" makes it self-explanatory + people may want to deploy Someguy in non-Mainnet context, and that instance would no longer be "Amino" – following distinction suggested in https://github.com/ipfs/specs/pull/497)

guillaumemichel: > (not feeling strongly about name, but leaning towards kad-dht because "dht" makes it self-explanatory + people may want to deploy Someguy in non-Mainnet context, and that instance would no longer be "Amino" – following distinction suggested in https://github.com/ipfs/specs/pull/497)

If someguy is hooked up to multiple dht swarms (e.g Amino and private swarms), then it doesn't make sense to return a _global_ list of closest peers, the caller would be interested in the closest peers for a specific dht swarm (if specific swarm is supported by that someguy instance).

We could use `/routing/v1/dht/{swarm}/closest-peers/{peer-id}?[closerThan]&[count]`. `swarm` needs to be a unique identifier for a swarm of DHT peers, e.g `amino`. Which means that we probably need to add `/routing/v1/dht/list-swarms` (or similar) to get the list of supported swarm identifiers (and descriptions?).

lidel: Triage notes from the public Boxo/GO  triage today (cc @hsanjuan): 
- https://github.com/ipfs/specs/pull/497 exists now
  - it defines AminoDHT as instance of kad-dht with specific libp2p protocol (`/ipfs/kad/1.0.0`)
  - people may use this API for delegating DHT instance different than the AminoDHT, so we may want to go with generic `kad-dht` and not `amino`
  - do not overengineer the MVP, go with `/routing/v1/dht/closest/{peerid}` , rationale:
    - less is more
    - peerid already conveys information we mean closest peers
    - we wont have more than one public DHT any time soon, and if people run private ones, unlikely they need public one as well, so ok to just go with `/routing/v1/dht/`
    - if we ever need to support more than one, we can always have different routing servers for each swarm: `dht1.example.com/routing/v1/dht` and `other-dht.example.com/routing/v1/dht`
    - or use explicit query parameter that requires explicit swarm using its libp2p protocol name (we dont invent any extra dictionaries/mappings): `?swarm=/custom/ipfs/kad/1.0.0`
    - if we have something that does not follow the above KAD-DHT spec, that would be a different endpoint anyway
    
@hsanjuan if you want to prototype this in https://github.com/ipfs/boxo/issues/1004, let's go with simple MVP:
  -  `/routing/v1/dht/closest/{peer-id}`

lidel:
@achingbrain FYSA to make this bit more formal, generated IPIP-0476 documenting why we need this: it includes the motivation from the PR (browser nodes needing lightweight peer discovery), the specification, rationale, and future-proofing considerations discussed in the PR comments.

Feel free to add any context to the document you find useful.

hsanjuan (approved):

src/ipips/ipip-0476.md
```diff
@@ -0,0 +1,145 @@
...
+
+Add a new HTTP endpoint to the Delegated Routing API that allows clients to find the closest peers to a given peer ID without being full DHT clients.
+
+## Motivation
+
+Browser nodes and other resource-constrained clients need to perform peer discovery operations without the overhead of being full DHT clients. The primary use case is for browser nodes performing random walks to find peers that they can make circuit relay reservations on.
+
+Currently, to find peers close to a particular key in the DHT keyspace, a node must:
+1. Be a full DHT client with all the associated overhead
+2. Maintain connections to many peers
```
guillaumemichel, line 33: In theory this isn't necessary.

It is true that many connections are necessary to perform a lookup, but a client technically doesn't need to maintain any connection when not performing a lookup actively. It should remember at least some addresses of DHT servers (e.g bootstrappers are enough).

I think we just miss light DHT client implementations.

src/ipips/ipip-0476.md
```diff
@@ -0,0 +1,145 @@
...
+
+This IPIP introduces a new "DHT Routing API" section to the Delegated Routing V1 HTTP API specification with the following endpoint:
+
+### `GET /routing/v1/dht/closest/peers/{peer-id}`
+
+#### Path Parameters
+
+- `peer-id`: The target peer ID to find closest peers for, represented as a CIDv1 encoded with `libp2p-key` codec
+
+#### Query Parameters
```
guillaumemichel, line 51: Do we have any specific use cases in mind that would require query parameters?

IIUC Go and JS DHT implementations only support returning the `20` closest peers. Since the `FIND_NODE` RPC returns at most `20` peers, it is quite complex to lookup for more than `20` peers.

The query parameters could be used to filter the results and return less than 20 peers, but not to get more peers.

lidel: @achingbrain @hsanjuan would it be ok to remove `closer-than` and  `count`?

The only implementation (kad-dht) doesn't support these parameters:
- go-libp2p-kad-dht
  - [GetClosestPeers only takes (ctx, key string)](https://github.com/libp2p/go-libp2p-kad-dht/blob/4152c7c0b5a1ca862b4fd6b8034a4f4edd7f2530/lookup.go#L23) 
  - Always returns K (20) closest peers
  - No filtering by closer-than

Imo there is no point in listing them in spec and implementing filtering in userland.

src/routing/http-routing-v1.md
```diff
@@ -244,6 +244,62 @@ The content body must be a [`application/vnd.ipfs.ipns-record`][application/vnd.
 - `400` (Bad Request): the provided :ref[IPNS Record] or :ref[IPNS Name] are not valid.
 - `406` (Not Acceptable): submitted content type is not supported. Error message returned in body should inform the user to retry with `Content-Type: application/vnd.ipfs.ipns-record`.
 
+## DHT Routing API
+
+### `GET /routing/v1/dht/closest/peers/{peer-id}`
+
+#### Path Parameters
+
+- `peer-id` is a [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) represented as a CIDv1 encoded with `libp2p-key` codec.
```
guillaumemichel, line 253: I realize now that we could extend the API beyond peer ids, e.g to look for the closest peers to a CID or an IPNS key.

Instead of naming this field `peer-id` we could name it `cid` (or `target`?) and expect a CID (could be a CIDv1 encored with `libp2p-key` codec). This would allow to find the closest DHT servers to either 1) a peer ID, 2) a cid or 3) an IPNS key, which roughly cover all IPFS use cases of the Amino DHT.

---

Currently in Go land, our interfaces simply accept a string as input and will look for the closest DHT servers to that string, using this [conversion function](https://github.com/libp2p/go-libp2p-kbucket/blob/0004129afd078b55497b3b977212ff4329c75147/util.go#L49-L53) to get a 256-bit kademlia identifier.

In [go-libp2p-kad-dht](https://github.com/libp2p/go-libp2p-kad-dht), the 3 identifiers passed to [`GetClosestPeers(ctx, string)`](https://github.com/libp2p/go-libp2p-kad-dht/blob/861573b39723ff0522535b9c751a4f723d582a87/lookup.go#L23) are the string cast from 1) a peer id for `FIND_NODE`, 2) the multihash (`bytes`) contained in a CID for `ADD_PROVIDER` and `FIND_PROVIDERS` and 3) the multihash contained in the peer id of the IPNS key for `PUT` and `GET` (IPNS).

guillaumemichel (commented):
It would be useful to extend the scope of the API to allow getting the closest DHT servers to CIDs and IPNS keys in addition to Peer IDs.

src/routing/http-routing-v1.md
lidel: Triage notes:
- switch to `{cid}` and double-check if code is compatible with that
- test end-to-0end in someguy first, before merging any PRs

src/routing/http-routing-v1.md
lidel: **RESOLVED in implementation**: The endpoint shipped as `/routing/v1/dht/closest/peers/{key}` in someguy v0.11.0 (released 2025-11-20).

After discussion, we decided on `/dht/` rather than `/amino/` or `/kad-dht/` because this is a DHT operation, not specific to Amino DHT deployment. Allows usage in non-Mainnet DHT contexts where "Amino" or Kademlia specific semantics wouldn't apply

See implementation in:
- boxo v0.35.2: https://github.com/ipfs/boxo/pull/1021
- someguy v0.11.0: https://github.com/ipfs/someguy/releases/tag/v0.11.0

lidel (approved):
## Ratified and shipped in someguy v0.11.0

Thank you to everyone else who contributed to this IPIP through design discussions and implementation work.

The reference implementation of IPIP-476 has shipped in **someguy v0.11.0** (released 2025-11-20) and is now deployed to the public delegated routing endpoint at `https://delegated-ipfs.dev/routing/v1` (documented at
https://docs.ipfs.tech/concepts/public-utilities/#delegated-routing-endpoint).

**What shipped:**
- Endpoint: `GET /routing/v1/dht/closest/peers/{key}`
- Accepts both CID and Peer ID formats
- Returns up to 20 DHT-closest peers (Amino DHT bucket size)
- Peers sorted by XOR distance
- No query parameters in MVP (simplified for initial deployment)

**Implementation references:**
- boxo v0.35.2: https://github.com/ipfs/boxo/pull/1021
- someguy v0.11.0: https://github.com/ipfs/someguy/releases/tag/v0.11.0

The specs have been updated to reflect what actually shipped. 
I've addressed the unresolved discussion threads above with resolution notes.

Merging as ratified – further changes can be made by opening a new IPIP.

PR #480, opened 2024-07-29
fix(gateway): clarify cache-control of generated html on /ipfs and TTL on /ipns

lidel:
This PR clarifies and recommends `Cache-Control`  value to be returned on generated HTML responses  such as UnixFS Directory listings, raised by @acejam in https://github.com/ipfs/boxo/pull/639 & https://github.com/ipfs/gateway-conformance/pull/219.

No IPIP since this is a gap is specs surface, not a change to existing spec.


src/http-gateways/path-gateway.md
```diff
@@ -426,14 +426,19 @@ Returned directive depends on requested content path and format:
 
 - `Cache-Control: public, max-age=29030400, immutable` MUST be returned for
   every immutable resource under `/ipfs/` namespace.
+  - Generated `/ipfs/` responses such as UnixFS directory listings are not
+    byte-for-byte immutable and SHOULD have a meaningful expiration that allows
+    for generated responses to be updated over time. Suggested value is one week:
```
lidel, line 431:
```suggestion
    for generated responses to be updated over time. Suggested value is one week (or longer):
```

Bonvi28 (approved):

lidel:
Pushed a couple more commits. Working through ipfs/boxo#1166 made the nuance here clearer: a fixed `stale-while-revalidate` can keep a cached IPNS record alive past its EOL, where the signature has already expired and clients reject it (those sporadic 500s in the service worker gateway).

So the `/ipns/` guidance now says `max-age` plus any stale window has to stay within the record's validity, with `no-store` once it's expired. I also kept the fixed generated-HTML window scoped to `/ipfs/`, since a directory listing reached through `/ipns/` should follow the IPNS TTL/EOL rules rather than the 4-week immutable one.

Same fix applied to `GET /routing/v1/ipns/{name}`, which overshot the same way.

Bonvi28 (approved):

PR #482, opened 2024-08-23
denylists: clarify doublehash way of supporting IPNS

lidel:
This PR adds a short mention how legacy doublehash notation can be used for representing IPNS and DNSLink block rules.

## Rationale

What prompted me to adding this to specs is the fact that IPNS entries (double-hashed CIDv1 with libp2p-key codec) already are on https://badbits.dwebops.pub/ and we also have cases where a very bad ENS DNSLink needs to be blocked for legal reasons. 

No IPIP because it does not change spec, only clarifies how to use existing notation in mutable context, which was missing from the spec.

Ref. internal issues with additional context (sadly too sensitive, can't be public):
- https://github.com/ipshipyard/waterworks-infra/issues/209
- https://github.com/protocol/badbits.dwebops.pub/pull/40002

### DIFF
--- a/src/compact-denylist-format.md
+++ b/src/compact-denylist-format.md
@@ -2,7 +2,7 @@
 title: Compact Denylist Format
 description: >
   How content blocking rules can be represented as a .deny file.
-date: 2023-10-25
+date: 2023-08-24
 maturity: reliable
 editors:
   - name: Hector Sanjuan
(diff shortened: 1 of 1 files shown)


hsanjuan (approved):
LGTM... This does not need any implementation NoPFS changes right?

lidel:
@hsanjuan unsure (focused on legacy infra first), for nopfs filled https://github.com/ipfs-shipyard/nopfs/issues/40 to make sure we check `/ipns`.

hsanjuan:
Thanks, I missed the notification!

PR #483, opened 2024-09-04
chore: grammatical fix in compact-denylist-format

SgtPooki:
I was reading through this spec and noticed a slight grammatical error. read through the rest and didn't spot anything


lidel (approved):
Thanks!

PR #484, opened 2024-09-06
IPIP-484: Opt-in Filtering on Routing V1 HTTP API

2color:


src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,104 @@
...
+- The value of the `network-transports` parameter is a comma-separated list of network transport protocol _name strings_ as defined in the [multiaddr protocol registry](https://github.com/multiformats/multiaddr/blob/master/protocols.csv), e.g. `?network-transports=webtransport`.
+- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs.
+- Only multiaddrs that pass the filter are included in the response.
+- If there are no multiaddrs that match the passed transports, the provider is omitted from the response.
+- Negative filtering is done by prefixing the protocol name with `!`, e.g. `?network-transports=!quic-v1,!tcp`.
+- Filtering is case-insensitive.
+- If no parameter is passed, the default behavior is to not apply filtering by network transports.
+
+:::note
+Filtering out providers without any matching transport, will cause providers without any multiaddrs to be filtered out, even if they _might_ support the requested transports. This is due to the fact that provider records are independent of peer records, and it's pretty common to have provider records without up-to-date multiaddrs for that peer.
```
lidel, line 59: I wonder if we should support `unknown` here, as a special value that helps with this and allows returning empty list, so the client can still store such peers and use them as a last resort fallback, if  better ones fail.

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,104 @@
...
+:::
+
+### Transfer Protocol Filtering
+
+The proposed change is to add a `?transfer-protocols` parameter to the `GET /routing/v1/providers/{cid}` endpoint of :cite[http-routing-v1]:
+
+- Add a `?transfer-protocols=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` that indicates which transfer protocols to return by filtering the `Protocol` field of the [Peer schema].
```
lidel, line 66:
```suggestion
- Add a `?transfer-protocols=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` that indicates which transfer protocols to return by filtering the `Protocol` field of the [Peer schema] (logical OR).
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Detailed design
+
+### Network Transport Protocol Filtering
+
+The proposed change is to add a `?network-transports` parameter to the `GET /routing/v1/providers/{cid}` endpoint of :cite[http-routing-v1]:
+
+- Add a `?network-transports=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` that indicates which network transports to return by filtering the multiaddrs in the `Addrs` field of the [Peer schema].
+- The value of the `network-transports` parameter is a comma-separated list of network transport protocol _name strings_ as defined in the [multiaddr protocol registry](https://github.com/multiformats/multiaddr/blob/master/protocols.csv), e.g. `?network-transports=webtransport`.
+- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs.
```
lidel, line 51:
```suggestion
- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs (logical OR).
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+
+However, there are many cases where most of the results from the Delegated Routing v1 API are not actionable by clients, because the client does not support either the **network transport protocol** or the **transfer protocol** of the provider.
+
+For instance, web browsers are limited to a specific set of network transport protocols, namely HTTPS, Secure WebSockets, WebTransport (emerging), and WebRTC. Consequently, providing information about providers that exclusively support TCP and/or UDP is not beneficial for browser-based clients, as they are unable to utilize such connections.
+
+Moreover, [Helia](https://github.com/ipfs/helia/), the most actively maintained browser IPFS implementation, supports block retrieval by CID with Bitswap and Trustless Gateways, but does not support GraphSync.
+
+This means that returning providers that only support GraphSync from the Delegated Routing API is not useful for browser clients, and results in unnecessary network traffic for browser clients.
+
+## Note on terminology
```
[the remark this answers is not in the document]
2color: > How do you feel about something like:
> 
> * `filter-addrs` → filter applied to multiaddrs from `Addrs` field
> * `filter-protocols` → filter applied to optional `Protocols` field

Love it!

src/ipips/ipip-0484.md
2color: Yeah good idea. Retain the same semantics as transfer protocol filtering.

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,102 @@
...
+
+- Add a `?filter-addrs=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` that indicates which network transports to return by filtering the multiaddrs in the `Addrs` field of the [Peer schema].
+- The value of the `filter-addrs` parameter is a comma-separated list of network transport protocol _name strings_ as defined in the [multiaddr protocol registry](https://github.com/multiformats/multiaddr/blob/master/protocols.csv), e.g. `?filter-addrs=webtransport`.
+- `unknown` can be be passed to include providers whose multiaddrs are unknown, e.g. `?filter-addrs=unknown`. This allows filtering providers whose multiaddrs are unknown at the time of filtering.
+- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs (logical OR).
+- Only multiaddrs that pass the filter are included in the response.
+- If there are no multiaddrs that match the passed transports, the provider is omitted from the response.
+- Negative filtering is done by prefixing the protocol name with `!`, e.g. `?filter-addrs=!quic-v1,!tcp`.
```
2color, line 55: negative filtering with logical OR will cause problems. 

For example, if we want to get all non `p2p-circuit` that are `webtransport` and `webrtc-direct` maddrs,  i.e. `?filter-addrs=!p2p-circuit,webtransport,webrtc-direct` we get something that looks like

`if (!maddr.has(p2p-circuit) || maddr.has(webtransport) || maddr.has(webrtc-direct)) { include maddr }`

which basically returns all non p2p-circuit addresses in addition to webtransport

Whereas we actually want something more like:

`if (!maddr.has(p2p-circuit) && (maddr.has(webtransport) || maddr.has(webrtc-direct)))  { include maddr }`

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,102 @@
...
+- `unknown` can be be passed to include providers whose transfer protocol is unknown, e.g. `?filter-protocols=unknown`. This allows filtering providers returned from the DHT that do not contain explicit transfer protocol information.
+- Providers are filtered by checking if the transfer protocol name appears in the `Protocols` array (logical OR).
+- If the provider doesn't match any of the passed transfer protocols, the provider is omitted from the response.
+- If a provider passes the filter, it is returned unchanged, i.e. the full set of protocols is returned including protocols that not included in the filter. (note that this is different from `filter-addrs` where only the multiaddrs that pass the filter are returned)
+- Negative filtering is done by prefixing the protocol name with `!`, e.g. `?transport=!transport-graphsync-filecoinv1`.
```
2color, line 70: I'm starting the question whether negative filtering for transfer protocols is actually needed or just unnecessary complexity. Clients should know which protocols they no. Not sure if it makes sense to implement this.

2color:
```suggestion
```

masih: Perhaps skip it for now and leave for possible extension?

masih (approved):
Overall I for one welcome this. Thank you for putting this one together @2color 🙏 

In therms of filter extensibility, my vote would be to start with the simplest most demanded one and expand later. In that pile, I can think of a more compact filtering using multicodec codes, bitfields, etc. all of which can be postponed for later.

2color:
> In that pile, I can think of a more compact filtering using multicodec codes, bitfields, etc. all of which can be postponed for later.

Thanks, good to have these optimisation suggestions on the record here. 

My take is that there are diminishing returns for these:
- multicodec codecs would be smaller, at the cost of reducing human debug-ability 
- bitfields would require assigning explicit bits for each protocol. This breaks the ability to treat filter values as opaque (to allows future evolution without regressions), because it requires assigning new bits as protocols are added.

masih:
> there are diminishing returns for these

The context for a more compact filtering suggestion on my part (which we should definitely ignore for now because of the "O"-word) is that I envision a world where these APIs would mostly be called by machines - not humans.

2color:
Open question: should we also add this to the `GET /routing/v1/peers/PeerID` endpoint?

src/ipips/ipip-0484.md
SgtPooki: would it be too complex to support grouping with `[]` or `()`?

SgtPooki (approved):
I'm definitely in favor of this one, and it looks good as a starting point.

src/ipips/ipip-0484.md
2color: Can you think of examples where you would want that? Otherwise, I'd say YAGNI

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,103 @@
...
+
+The Delegated Routing v1 API empowers resource constrained clients like web browsers by significantly reducing the number of network connections necessary to fetch content addressed data directly from provider peers.
+
+However, there are many cases where most of the results from the Delegated Routing v1 API are not actionable by clients, because the client does not support either the **network transport protocol** or the **transfer protocol** of the provider.
+
+For instance, web browsers are limited to a specific set of network transport protocols, namely HTTPS, Secure WebSockets, WebTransport (emerging), and WebRTC. Consequently, providing information about providers that exclusively support TCP and/or UDP is not beneficial for browser-based clients, as they are unable to utilize such connections.
+
+Moreover, [Helia](https://github.com/ipfs/helia/), the most actively maintained browser IPFS implementation, supports block retrieval by CID with Bitswap and Trustless Gateways, but does not support GraphSync.
```
gammazero, line 31: Change "GraphSync" to "Graphsync", as spelled in [spec](https://github.com/ipld/specs/blob/master/block-layer/graphsync/graphsync.md#graphsync)

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Add a `?filter-addrs=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` that indicates which network transports to return by filtering the multiaddrs in the `Addrs` field of the [Peer schema].
+- The value of the `filter-addrs` parameter is a comma-separated list of network transport protocol _name strings_ as defined in the [multiaddr protocol registry](https://github.com/multiformats/multiaddr/blob/master/protocols.csv), e.g. `?filter-addrs=webtransport`.
+- `unknown` can be be passed to include providers whose multiaddrs are unknown, e.g. `?filter-addrs=unknown`. This allows filtering providers whose multiaddrs are unknown at the time of filtering.
+- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs (logical OR).
+- Negative filtering is done by prefixing the protocol name with `!`, e.g. `?filter-addrs=!quic-v1,!tcp`. Note that negative filtering is done by checking if the protocol name does not appear in any of the multiaddrs (logical AND).
```
gammazero, line 53: Is there a use case for negative filtering? Why would a client want to exclude providers that had support for a particular protocol?

gammazero (approved):
see question

src/ipips/ipip-0484.md
2color: It's not strictly necessary. Clients should generally know which protocols/transports they support. 

This was @lidel's recommendation, which I thought would be useful without carefully thinking it through.

src/ipips/ipip-0484.md
2color, line 31:
```suggestion
Moreover, [Helia](https://github.com/ipfs/helia/), the most actively maintained browser IPFS implementation, supports block retrieval by CID with Bitswap and Trustless Gateways, but does not support Graphsync.
```

src/ipips/ipip-0484.md
lidel: Another example would be ipv4-only client that does not want to waste bandwidth for receiving ipv6 addrs. Perhaps we could improve example in the spec so the utility is more obvious?

```suggestion
- Negative filtering is done by prefixing the protocol name with `!`, e.g. to skip ipv6 and QUIC addrs: `?filter-addrs=!ip6,!quic-v1`. Note that negative filtering is done by checking if the protocol name does not appear in any of the multiaddrs (logical AND).
```

2color:
@lidel Is there anything left to address before merging?

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
+---
+title: 'IPIP-0484: Delegated Routing V1 Support for Querying Specific Network Transport and Transfer Protocols'
```
lidel, line 2: @2color we have to shorten it so it looks better on main page listings etc
```suggestion
title: 'IPIP-0484: Opt-in Provider and Peer Filtering on Routing V1 HTTP API'
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+- Multiaddrs are filtered by checking if the protocol name appears in any of the multiaddrs (logical OR).
+- Negative filtering is done by prefixing the protocol name with `!`, e.g. to skip IPv6 and QUIC addrs: `?filter-addrs=!ip6,!quic-v1`. Note that negative filtering is done by checking if the protocol name does not appear in any of the multiaddrs (logical AND).
+- If no parameter is passed, the default behavior is to return the original list of addresses unchanged.
+- If only negative filters are provided, addresses not passing any of the negative filters are included.
+- If positive filters are provided, only addresses passing at least one positive filter (and no negative filters) are included.
+- If both positive and negative filters are provided, the address must pass all negative filters and at least one positive filter to be included.
+- If there are no multiaddrs that match the passed transports, the provider is omitted from the response.
+- Filtering is case-insensitive.
+
+### Transfer Protocol Filtering
```
lidel, line 61: I realized these are IPFS-specific, and as we discussed, may not be limited to transfer, so let's highlight that.

```suggestion
### IPFS Protocol Filtering
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+
+The proposed change is to add a `?filter-protocols` parameter to the `GET /routing/v1/providers/{cid}` and `GET /routing/v1/peers/{peer-id}` endpoints of :cite[http-routing-v1]:
+
+- Add a `?filter-protocols=<comma-separated-list>` optional parameter to `GET /routing/v1/providers/{CID}` to filter providers based on the `Protocol` field of the [Peer schema].
+- The `filter-protocols` parameter is a comma-separated list of transfer protocol names, e.g. `?filter-protocols=transport-bitswap`.
+- Transfer protocols names should be treated as opaque strings and have a max length of 63 characters. A non-exhaustive list of transfer protocols are defined per convention in the [multicodec registry](https://github.com/multiformats/multicodec/blob/3b7b52deb31481790bc4bae984d8675bda4e0c82/table.csv#L149-L151).
+- Implementations MUST ignore unknown transfer protocol names.
```
lidel, line 68: `unknown` has a special meaning (empty protocol list), so let's rephrase:

```suggestion
- Implementations MUST preserve all transfer protocol names when returning a positive result that matches one or more of them.
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+- `unknown` can be be passed to include providers whose transfer protocol is unknown, e.g. `?filter-protocols=unknown`. This allows filtering providers returned from the DHT that do not contain explicit transfer protocol information.
```
lidel, line 69: This clarifies `unknown` means empty:
```suggestion
- A special `unknown` name can be be passed to include providers whose transfer protocol list is empty (unknown), e.g. `?filter-protocols=unknown`. This allows for including providers returned from the DHT that do not contain explicit transfer protocol information.
```

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+- Providers are filtered by checking if the transfer protocol name appears in the `Protocols` array (logical OR).
+- If the provider doesn't match any of the passed transfer protocols, the provider is omitted from the response.
+- If a provider passes the filter, it is returned unchanged, i.e. the full set of protocols is returned including protocols that not included in the filter. (note that this is different from `filter-addrs` where only the multiaddrs that pass the filter are returned)
+- Filtering is case-insensitive.
+- If no parameter is passed, the default behavior is to not filter by transfer protocol.
+
+:::note
+Even though existing transfer protocol names start with `transport`, e.g. `transport-bitswap`, `transport-graphsync-filecoinv1`, and `transport-ipfs-gateway-http`. This is not to be confused with the network transport protocols, which are filtered using the `filter-addrs` parameter.
```
lidel, line 77: nit: I think we renamed things a bit since this was written, maybe clarify?

```suggestion
Even though some of existing IPFS transfer protocol names start with `transport`, e.g. `transport-bitswap`, `transport-graphsync-filecoinv1`, and `transport-ipfs-gateway-http`, they should not to be confused with the network transport protocols used in peer addresses, which are filtered using the `filter-addrs` parameter.
```

lidel (changes requested):
@2color I think we are happy with this spec and unlikely to change any further (boxo implementation got merged: https://github.com/ipfs/boxo/pull/671).

The next steps here:

- [x] Update this PR to add "Request Query Parameters" sections to `/providers` and `/peers` endpoint docs at https://specs.ipfs.tech/routing/http-routing-v1/ (`src/routing/http-routing-v1.md`), and document the newly added parameters there.
   - Fine to copy-paste content, just follow section/header naming convention established in other HTTP specs so we have meaningful permalinks, example: https://specs.ipfs.tech/http-gateways/path-gateway/#request-query-parameters
   - Fine to keep content duplicated in IPIP, IPIP acts as a snapshot in time, while the spec document is a living document
- [x] Ship both client and server, to ensure spec truly works end-to-end and we've identified all the gaps
  - [x] release and deploy server (someguy at https://delegated-ipfs.dev/)
  - [x] release helia; it  should be asking only for actionable peers when running in browser context.
- [ ] Merge this PR (The specs / IPIP always land last, allowing community to engage over long window of time)

src/ipips/ipip-0484.md
```diff
@@ -0,0 +1,105 @@
...
+
+The term **"transport"** is overloaded in the IPFS ecosystem.
+
+In the context of this IPIP, we refer to the network layer transport protocol, e.g. TCP, QUIC, WebTransport, as **"network transport protocol"** to avoid ambiguity.
+
+**"Transfer protocol"** refers to data transfer protocols, i.e. content-addressed block retrieval protocols, e.g. Bitswap, Graphsync, HTTP.
+
+## Detailed design
+
+### Network Transport Protocol Filtering
```
lidel, line 45: maybe this allows us to avoid having Transport in both?

```suggestion
### Network Address Filtering
```

2color:
> release helia query for action

What do you by that?

lidel:
Ah, it cut-off, I meant helia should be asking only for actionable peers when running in browser context.

lidel (approved):
Thank you all for fleshing this out @2color.
No concerns were raised since September, applied final ratification edits, ok to be merged.

Reference implementation shipped in:
- https://github.com/ipfs/boxo/releases/tag/v0.24.1 (client/server)
- https://github.com/ipfs/rainbow/releases/tag/v1.8.1 (client)
- https://github.com/ipfs/someguy/releases/tag/v0.5.0 (client/server from boxo)
- https://github.com/ipfs/helia-delegated-routing-v1-http-api/releases/tag/%40helia%2Fdelegated-routing-v1-http-api-client-4.1.0 (client)
- and this will ship in kubo 0.32.0-rc1 this week (client)

PR #486, opened 2024-11-04
chore: clarify accepted Peer IDs

2color:
Clarify that the delegated routing peer routing endpoint accepts both CID and Multihash encoded PeerIDs

### DIFF
--- a/src/routing/http-routing-v1.md
+++ b/src/routing/http-routing-v1.md
@@ -52,7 +52,7 @@ As such, human-readable encodings of types are preferred. This specification may
 
 - CIDs are always string-encoded using a [multibase]-encoded [CIDv1].
 - Multiaddrs are string-encoded according to the [human-readable multiaddr specification][multiaddr].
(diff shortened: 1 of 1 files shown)


src/routing/http-routing-v1.md
```diff
@@ -148,7 +148,7 @@ Each object in the `Providers` list is a record conforming to a schema, usually
 #### Path Parameters
 
 - `peer-id` is the [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) to fetch peer records for,
-represented as a CIDv1 encoded with `libp2p-key` codec.
+represented as either a Multihash in Base58btc, or a CIDv1 with libp2p-key (`0x72`) codec.
```
lidel, line 151:
```suggestion
represented as either a Multihash in Base58btc, or a CIDv1 with libp2p-key (`0x72`) codec (in Base36 or Base32).
```

lidel (approved):

PR #491, opened 2025-02-12
trustless gateway spec: add probing path (identity cid).

hsanjuan:
Add section about status codes.

Add section about Retry-After.

### DIFF
--- a/src/http-gateways/trustless-gateway.md
+++ b/src/http-gateways/trustless-gateway.md
@@ -39,6 +39,24 @@ The minimal implementation means:
 
 A subset of "HTTP API" of :cite[path-gateway].
 
+## `GET /ipfs/bafkqaaa`
+
+`bafkqaaa` is the identity empty CID. This endpoint can be used to probe that that the endpoint corresponds to a trustless gateway.
+
+For block requests (`?format=raw`), when supported, it must return `200 OK` and an empty body.
+
+For CAR requests (`?format=car`), when supported, it must return `200 OK` and CAR file with root set to `bafkqaaa` and a single `bafkqaaa` block (which is empty).
+
+We recommend supporting only this specific identity CID and not random identity CIDs.
+
+## `HEAD /ipfs/bafkqaaa`
+
+`bafkqaaa` is the identity empty CID. This endpoint can be used to probe that that the endpoint corresponds to a trustless gateway that correctly supports `HEAD` requests.
+
+It must return `200 OK` in all cases.
+
+We recommend supporting only this specific identity CID and not random identity CIDs.
+
 ## `GET /ipfs/{cid}[/{path}][?{params}]`
 
 Downloads verifiable, content-addressed data for the specified **immutable** content path.
@@ -235,6 +253,21 @@ In case both are present in the request, the value from the [`Accept`](#accept-r
 
 Below MUST be implemented **in addition** to "HTTP Response" of :cite[path-gateway].
 
+## Status codes
+
+Trustless gateways MUST return reasonable status codes.
+
+* `200` for successful requests.
+* `4xx` for client errors.
+* `5xx` for server errors, server unavailability.
+
+For example:
+
+* A CID non retrievable from this gateway SHOULD return `404`
+* A blocked CID SHOULD return `410` or `451`.
+* A request with a malformed or unsupported CID SHOULD return `400`.
+* No `5xx` status should be used for client errors, like requesting an unexistent path under a CID.
+
 ## Response Headers
 
 ### `Content-Type` (response header)
@@ -256,6 +289,15 @@ MUST be returned and set to `attachment` to ensure requested bytes are not rende
 Same as in :cite[path-gateway], SHOULD be returned when Trustless Gateway
 supports more than a single response format and the `format` query parameter is
 missing or does not match well-known format from `Accept` header.
+
+### `Retry-After` (response header)
+
+In conjuction with status code `429` (for rate limiting) and `503` (maintenance), the `Retry-After` header can be used to signal how long the user agent should wait before making a follow-up request. This is a [standard response header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) with the following syntax:
+
+```
+Retry-After: <http-date>
+Retry-After: <delay-seconds>
+```
 
 # Block Responses (application/vnd.ipld.raw)
 


guillaumemichel (approved):
Seems reasonable 👍🏻

src/http-gateways/trustless-gateway.md
```diff
@@ -39,6 +39,24 @@ The minimal implementation means:
 
 A subset of "HTTP API" of :cite[path-gateway].
 
+## `GET /ipfs/bafkqaaa`
+
+`bafkqaaa` is the identity empty CID. This endpoint can be used to probe that that the endpoint corresponds to a trustless gateway.
+
+For block requests (`?format=raw`), when supported, it must return `200 OK` and an empty body.
+
+For CAR requests (`?format=car`), when supported, it must return `200 OK` and CAR file with root set to `bafkqaaa` and a single `bafkqaaa` block (which is empty).
```
lidel, line 48: go-car (used by most gateways) will skip identity blocks, which means CAR will be empty (no `Data`, just `Header`:

```suggestion
For CAR requests (`?format=car`), when supported, it must return `200 OK` and a valid, empty CAR file with CAR Header `roots` set to `bafkqaaa`.
```

src/http-gateways/trustless-gateway.md
```diff
@@ -235,6 +253,21 @@ In case both are present in the request, the value from the [`Accept`](#accept-r
 
 Below MUST be implemented **in addition** to "HTTP Response" of :cite[path-gateway].
 
+## Status codes
```
lidel, line 256: We should refer to https://specs.ipfs.tech/http-gateways/path-gateway/#response-status-codes and avoid duplicating code list. 

Follow convention from https://specs.ipfs.tech/http-gateways/trustless-gateway/#content-location-response-header and https://specs.ipfs.tech/http-gateways/trustless-gateway/#http-response where we refer to path-gateway spec where possible.

src/http-gateways/trustless-gateway.md
```diff
@@ -257,6 +290,15 @@ Same as in :cite[path-gateway], SHOULD be returned when Trustless Gateway
 supports more than a single response format and the `format` query parameter is
 missing or does not match well-known format from `Accept` header.
 
+### `Retry-After` (response header)
```
lidel, line 293: Update https://specs.ipfs.tech/http-gateways/path-gateway/#retry-after-response-header and refer to that.

src/http-gateways/trustless-gateway.md
```diff
@@ -39,6 +39,24 @@ The minimal implementation means:
 
 A subset of "HTTP API" of :cite[path-gateway].
 
+## `GET /ipfs/bafkqaaa`
```
lidel, line 42: Move GET/HEAD probing with identity CID to "notes for implementers" section at the end of document. Maybe call it "Inexpensive Health Check for Clients" ?

lidel (changes requested):
Thanks. Quick feedback inline.

src/http-gateways/trustless-gateway.md
hsanjuan: @lidel can you double check?

The reason I wrote it like this is because `ipfs dag export bafkqaaa > bafkqaaa.car` seemed to return a CAR with a single empty block in it (as in, https://pkg.go.dev/github.com/ipld/go-car/util#ReadNode reads a node in the car, so it doesn't seem to be just the root being set in the car header. It seems to include an empty block).

src/http-gateways/trustless-gateway.md
hsanjuan: I think a "dedicated probe paths" section would be enough ? See now, as it is RECOMMENDED from the description of GET/HEAD requests.

src/http-gateways/trustless-gateway.md
lidel: Yeah we have a discrepancy in behavior. I think spec should only require valid CAR with `roots` – we can't depend on `Data` because different implementations will produce different CARs. They may be logically equivalent, but won't be byte-for-byte identical.

You can inspect with `car` CLI (`go install github.com/ipld/go-car/cmd/car@latest`):

CAR produced by `boxo/gateway` ([code](https://github.com/ipfs/boxo/blob/7c3ac0d8f9e1718ebe0d637c04ab597a54a93569/gateway/backend_blocks.go#L284)) seem to correctly skip identity CID:

```console
$ curl "http://127.0.0.1:8080/ipfs/bafkqaaa?format=car" -o bafkqaaa-gw.car -s
$ car inspect bafkqaaa-gw.car
Version: 1
Roots: bafkqaaa
Root blocks present in data: No
Block count: 0
...
```

CAR produced by `ipfs dag export` ([code](https://github.com/ipfs/kubo/blob/f00e115e1ed094d47b3abe30d86fea40c6ca2f60/core/commands/dag/export.go#L24)) CLI does not not skip:

```console
$ ipfs dag export bafkqaaa > bafkqaaa-cli.car
$ car inspect bafkqaaa-cli.car
Version: 1
Roots: bafkqaaa
Root blocks present in data: Yes
Block count: 1
```


Which one is "more correct"? 
On the principle of not wasting space, identity blocks should be skipped, but they also not hurt anyone if present – just waste space because data is included twice.

src/http-gateways/path-gateway.md
```diff
@@ -359,23 +373,46 @@ See: [Denylists](#denylists)
...
 
-This error response SHOULD include [`Retry-After`](#retry-after-response-header) HTTP header to indicate how long the client should wait before retrying.
+This error response SHOULD include
+[`Retry-After`](#retry-after-response-header) HTTP header to indicate how long
+the client should wait before retrying.
+
+Gateways MUST return `404` instead of `502` when the content is known to be
+unretrivable: for example, when the Gateway is
+[non-recursive](#recursive-vs-non-recursive-gateways) and the content is known
+to not be available.
```
lidel, line 395: nit: MUST is not realistic

```suggestion
Gateways SHOULD return `404` instead of `502` when the content is known to be
unretrievable: for example, when the Gateway is
[non-recursive](#recursive-vs-non-recursive-gateways) and the content is known
to not be available.
```

src/http-gateways/path-gateway.md
```diff
@@ -359,23 +373,46 @@ See: [Denylists](#denylists)
...
+
+This error response SHOULD include
+[`Retry-After`](#retry-after-response-header) HTTP header to indicate how long
+the client should wait before retrying.
 
-There is no generic timeout, Gateway implementations SHOULD set timeouts based on specific use cases.
+Gateways MUST return `404` instead of `504` when the content is known to be
+unretrivable: for example, when the Gateway is
+[non-recursive](#recursive-vs-non-recursive-gateways) and the content is known
+to not be available.
```
lidel, line 414:
```suggestion
Gateways SHOULD return `404` instead of `504` when the content is known to be
unretrievable: for example, when the Gateway is
[non-recursive](#recursive-vs-non-recursive-gateways) and the content is known
to not be available.
```

src/http-gateways/path-gateway.md
```diff
@@ -835,6 +878,21 @@ The usual optimizations involve:
   - The downside of this approach is that it will always be slower than
     skipping child block resolution.
 
+## Recursive vs non-recursive gateways
+
+A *recursive Gateway* is a gateway which generally attempts to fetch content
+from a third party it does not control by triggering lookups and retrievals. A
+recursive Gateway may not know in advance whether it can obtain and return a
+piece of content as the availability of it is out of its control. It may also
+suggest that clients retry failed requests later via `502` and `504` respones
```
lidel, line 887:
```suggestion
suggest that clients retry failed requests later via `502` and `504` responses
```

src/http-gateways/path-gateway.md
```diff
@@ -835,6 +878,21 @@ The usual optimizations involve:
...
+status codes.
+
+A *non-recursive Gateway* is gateway which accesses a known content-set and,
+under normal operation conditions, knows with certainity whether content
```
lidel, line 891:
```suggestion
under normal operation conditions, knows with certainty whether content
```

src/http-gateways/trustless-gateway.md
```diff
@@ -47,10 +47,14 @@ Optional `path` is permitted for requests that specify CAR format (`?format=car`
 
 For block requests (`?format=raw` or `Accept: application/vnd.ipld.raw`), only `GET /ipfs/{cid}[?{params}]` is supported.
 
+It is RECOMMENDED to additionally implement the [`GET` probe path](#dedicated-probe-paths).
```
lidel, line 50:
```suggestion
Client and Server implementations SHOULD include support for the [`GET` probe path](#dedicated-probe-paths).
```

src/http-gateways/trustless-gateway.md
```diff
@@ -47,10 +47,14 @@ Optional `path` is permitted for requests that specify CAR format (`?format=car`
...
+
 ## `HEAD /ipfs/{cid}[/{path}][?{params}]`
 
 Same as GET, but does not return any payload.
 
+It is RECOMMENDED to additionally implement the [`HEAD` probe path](#dedicated-probe-paths).
```
lidel, line 56:
```suggestion
Client and Server implementations SHOULD include support for the [`HEAD` probe path](#dedicated-probe-paths).
```

src/http-gateways/trustless-gateway.md
```diff
@@ -438,3 +444,34 @@ returned as [application/vnd.ipfs.ipns-record](https://www.iana.org/assignments/
...
+
+`bafkqaaa` is the identity empty CID. This endpoint can be used to probe that
+that the endpoint corresponds to a trustless gateway.
+
+For block requests (`?format=raw`), when supported, it must return `200 OK`
+and an empty body.
+
+For CAR requests (`?format=car`), when supported, it must return `200 OK` and
+CAR file with root set to `bafkqaaa` and a single `bafkqaaa` block (which is
+empty).
```
lidel, line 464: Given discrepancy in https://github.com/ipfs/specs/pull/491#discussion_r1962065874 this should be adjusted to require CID in header, but make data trully optional:

```suggestion
For CAR requests (`?format=car`), when supported, it MUST return `200 OK` and a valid CAR file with CAR Header `roots` set to `bafkqaaa`. Identity block MAY be skipped in the CAR Data section.
```

lidel (changes requested):
second pass, lgtm, but some MUSTs and phrasing should be relaxed to match reality (details/suggestions inline)

src/http-gateways/trustless-gateway.md
hsanjuan: Ok, I take your fix.

src/http-gateways/path-gateway.md
```diff
@@ -641,9 +677,15 @@ Optional, present in certain response types:
...
-Gateway returns this header with error responses such as [`429 Too Many Requests`](#429-too-many-requests) or [`504 Gateway Timeout`](#504-gateway-timeout).
+Gateway SHOULD return this header with error responses such as [`429 Too Many Requests`](#429-too-many-requests), [`504 Gateway Timeout`](#504-gateway-timeout) or `503` (server maintainance).
 
-The "Retry-After" header indicates how long the user agent ought to wait before making a follow-up request.
+The "Retry-After" header indicates how long the user agent ought to wait before making a follow-up request. It uses the following syntax:
+
+```
+Retry-After: <http-date>
+// or
+Retry-After: <delay-seconds>
```
lidel, line 687: :thought_balloon:   Date is virtually never used in real world, most services (Nginx, AWS, Cloudflare, Caddy) default to using delay in seconds for simplicity and practicality. 

Putting it here introduces unnecessary complexity (similar to range requests that are for more than one range – they are in spec, but nothing implements them correctly enough for it to be used). 

Let's simplify

```suggestion
Retry-After: <delay-seconds>
```

src/http-gateways/trustless-gateway.md
```diff
@@ -438,3 +444,32 @@ returned as [application/vnd.ipfs.ipns-record](https://www.iana.org/assignments/
...
+## Dedicated Probe Paths
+
+Trustless gateways SHOULD provide probing endpoints as described below.
+
+### `GET /ipfs/bafkqaaa`
+
+`bafkqaaa` is the identity empty CID. This endpoint can be used to probe that
+that the endpoint corresponds to a trustless gateway.
+
+For block requests (`?format=raw`), when supported, it MUST return `200 OK`
```
lidel, line 459:
```suggestion
For block requests (signaled by `?format=raw` and `Accept: application/vnd.ipld.raw`), when supported, it MUST return `200 OK`
```

src/http-gateways/trustless-gateway.md
```diff
@@ -438,3 +444,32 @@ returned as [application/vnd.ipfs.ipns-record](https://www.iana.org/assignments/
...
+and an empty body.
+
+For CAR requests (`?format=car`), when supported, it MUST return `200 OK` and a valid CAR file with CAR Header `roots` set to `bafkqaaa`. Identity block MAY be skipped in the CAR Data section.
```
lidel, line 462:
```suggestion
For CAR requests (signaled by `?format=car` and `Accept: application/vnd.ipld.car`), when supported, it MUST return `200 OK` and a valid CAR file with CAR Header `roots` set to `bafkqaaa`. Identity block MAY be skipped in the CAR Data section.
```

src/http-gateways/trustless-gateway.md
```diff
@@ -438,3 +444,32 @@ returned as [application/vnd.ipfs.ipns-record](https://www.iana.org/assignments/
 A Client MUST confirm the record signature match `libp2p-key` from the requested IPNS Name.
 
 A Client MUST [perform additional record verification according to the IPNS specification](https://specs.ipfs.tech/ipns/ipns-record/#record-verification).
+
+# Notes for implementers
```
lidel, line 448:
```suggestion
# Appendix: Notes for implementers
```

lidel (approved):
Lgtm, small editorials inline

PR #497, opened 2025-03-18
kad: adding DHT spec

guillaumemichel:
Long overdue

Fixes https://github.com/ipfs/specs/issues/345

### DIFF
--- a/src/routing/kad-dht.md
+++ b/src/routing/kad-dht.md
@@ -336,7 +336,12 @@ more Peer IDs providing the corresponding content. Instead of storing the
 content itself, the DHT stores provider records pointing to the peers hosting
 the content.
 
-A Provider Record is identified by the multihash contained by the CID.
+A Provider Record is identified by the multihash contained by the CID. It
+functions as an append-only list, where multiple providers can add themselves
+as content hosts. Since strict consistency across the network is not required,
+different DHT servers MAY store slightly different sets of providers, but the
+lookup mechanism ensures that clients can still discover multiple sources
+efficiently.
 
 ### Content Kademlia Identifier
 
@@ -398,22 +403,97 @@ for the Client
 ### Content Provider Lookup
 
 To find providers for a given CID, a node initiates a lookup using the
-GET_PROVIDERS RPC. This process follows the same approach as a FIND_NODE
+`GET_PROVIDERS` RPC. This process follows the same approach as a `FIND_NODE`
 lookup, but with one key difference: if a DHT server holds a matching provider
 record, it MUST include it in the response.
 
 Clients MAY terminate the lookup early if they are satisfied with the returned
 providers. If a node does not find any provider records and is unable to
-discover closer DHT servers after querying the β closest reachable servers, the
-request is considered a failure.
+discover closer DHT servers after querying the `β` closest reachable servers,
+the request is considered a failure.
 
-## DHT Record Storage
+## Value Storage and Retrieval
+
+The IPFS Kademlia DHT allows users to store and retrieve records directly
+within the DHT. These records serve as key-value mappings, where the key and
+value are defined as arrays of bytes. Each record belongs to a specific
+keyspace, which defines its type and structure.
+
+The IPFS Kademlia DHT supports two types of records, each stored in its own
+keyspace:
+
+1. **Public Key Records** (`/pk/`) – Used to store public keys that cannot be
+   derived from Peer IDs.
+2. **IPNS Records** (`/ipns/`) – Used for decentralized naming and content
+   resolution.
+
+Records MUST meet validity criteria specific to their record type before being
+stored or updated. DHT Servers MUST verify the validity of each record before
+accepting it.
+
+### Routing
+
+The Kademlia Identifier of a record is derived by applying the SHA256 hash
+function to the record’s key and using the resulting digest in binary format.
+
+To store a value in the DHT, a client first finds the `k` closest peers to the
+record’s Kademlia Identifier using `GetClosestPeers`. The client then sends a
+`PUT_VALUE` RPC to each of these peers, including the `key` and the `record`.
+DHT servers MUST validate the record based on its type before accepting it.
+
+Retrieving values from the DHT follows a process similar to provider record
+lookups. Clients send a `GET_VALUE` RPC, which directs the search toward the
+`k` closest nodes to the target `key`. If a DHT Server holds a matching
+`record`, it MUST include it in its response. The conditions for terminating
+the lookup depend on the specific record type.
+
+### Public Keys
+
+Some public keys are too large to be embedded within libp2p Peer IDs ([keys
+larger than 42
+bytes](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md#peer-ids)).
+In such cases, the Peer ID is derived from the hash of the public key, but the
+full key still needs to be accessible. To facilitate retrieval, public keys MAY
+be stored directly in the DHT under the `/pk/` keyspace.
+
+1. Key: `/pk/<PeerID>` (binary Peer ID format).
+2. Value: The full public key (in binary format).
+
+#### Validation
+
+DHT servers MUST verify that the Peer ID derived from the full public key
+matches the Peer ID encoded in the key. If the derived Peer ID does not match,
+the record MUST be rejected.
 
 ### IPNS
 
-### Resiliency
+IPNS (InterPlanetary Naming System) allows peers to publish mutable records
+that point to content in IPFS. These records MAY be stored in the DHT under the
+`/ipns/` namespace.
 
-### Validators
+Record format and validation is documented in the [IPNS
+specification](https://specs.ipfs.tech/ipns/ipns-record/). IPNS records are
+limited in size to
+[10KiB]((https://specs.ipfs.tech/ipns/ipns-record/#record-size-limit)).
+
+#### Quorum
+
+A quorum is the minimum number of distinct responses a client must collect from
+DHT Servers to determine a valid result. Since different DHT Servers may store
+different versions of an IPNS record, a client fetches the record from multiple
+DHT Servers to increase the likelihood of retrieving the most recent version.
+
+For IPNS lookups, the default quorum value is `16`, meaning the client attempts
+to collect responses from at least `16` DHT Servers out of `20` before
+determining the best available record.
+
+#### Entry Correction
+
+Because some DHT servers may store outdated versions of a record, clients need
+to ensure that the latest valid version is propagated. After obtaining a
+quorum, the client MUST send the most recent valid record to any of the `k`
+closest DHT Servers to the record’s Kademlia Identifier that did not return the
+latest version.
 
 ## RPC Messages
 
@@ -570,6 +650,8 @@ Make a `FIND_NODE` request and inspect response before adding node to RT. Follow
 * Go: [`libp2p/go-libp2p-kad-dht`](https://github.com/libp2p/go-libp2p-kad-dht)
 * JS: [libp2p/kad-dht](https://github.com/libp2p/js-libp2p/tree/main/packages/kad-dht)
 * Rust: [libp2p-kad](https://github.com/libp2p/rust-libp2p/tree/master/protocols/kad)
+
+---
 
 ## References
 
--- a/src/routing/kad-dht.md
+++ b/src/routing/kad-dht.md
@@ -113,8 +113,8 @@ identifier for incoming streams.
 
 ## Kademlia Keyspace
 
-Kademlia [0] operates on a binary keyspace defined as $\{0, 1\}^m$. In
-particular, the IPFS Kademlia DHT uses a keyspace of length $m=256, containing
+Kademlia [0] operates on a binary keyspace defined as $\lbrace 0,1 \rbrace^m$. In
+particular, the IPFS Kademlia DHT uses a keyspace of length $m=256$, containing
 all bitstrings of 256 bits. The distance between any pair of keys is defined as
 the bitwise XOR of the two keys, resulting in a new key representing the
 distance between the two keys. This keyspace is used for indexing both nodes
--- a/src/routing/kad-dht.md
+++ b/src/routing/kad-dht.md
@@ -6,10 +6,11 @@ description: >
   InterPlanetary File System (IPFS). It extends the libp2p Kademlia DHT
   specification, adapting and adding features to support IPFS-specific
   requirements.
-date: 2022-08-26
+date: FIXME
 maturity: reliable
 editors:
   - name: Guillaume Michel
+    url: https://guillaume.michel.id
     github: guillaumemichel
     affiliation:
       name: Shipyard
@@ -28,6 +29,10 @@ requirements.
 
 FIXME:
 
+Distributed Key-Value Store
+
+Goal of DHT is to find the closest peers to some key (in a specific geometry). Once this routing to the closest nodes is possible, nodes can interact with these nodes in various ways, including in asking them to store and serve data.
+
 ### Relation to [libp2p kad-dht](https://github.com/libp2p/specs/tree/master/kad-dht)
 
 The IPFS Kademlia DHT specification is a specialization of the libp2p Kademlia DHT.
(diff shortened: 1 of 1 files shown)


src/routing/kad-dht.md
```diff
@@ -0,0 +1,666 @@
...
+
+---
+
+## References
+
+[0]: Maymounkov, P., & Mazières, D. (2002). Kademlia: A Peer-to-Peer
+Information System Based on the XOR Metric. In P. Druschel, F. Kaashoek, & A.
+Rowstron (Eds.), Peer-to-Peer Systems (pp. 53–65). Berlin, Heidelberg: Springer
+Berlin Heidelberg. [DOI](https://doi.org/10.1007/3-540-45748-8_5)
+[pdf](https://www.scs.stanford.edu/~dm/home/papers/kpos.pdf)
```
guillaumemichel, line 666: not sure what is the best format for to cite papers with the `spec-generator`

src/routing/kad-dht.md
```diff
@@ -0,0 +1,666 @@
+---
+title: Kademlia DHT
+description: >
+  The IPFS Distributed Hash Table (DHT) specification defines a structured
+  overlay network used for peer routing and content routing in the
+  InterPlanetary File System (IPFS). It extends the libp2p Kademlia DHT
+  specification, adapting and adding features to support IPFS-specific
+  requirements.
+date: 2025-03-18
+maturity: reliable
```
guillaumemichel, line 10: Set to `reliable` since this spec transcribes the network behavior as it currently is

src/routing/kad-dht.md
```diff
@@ -0,0 +1,707 @@
...
+NOT return any information about unresponsive nodes.
+
+In public DHT swarms, DHT Servers MUST filter out private and loopback
+multiaddresses, and MUST NOT include peers whose only addresses are private or
+loopback.
+
+DHT Servers SHOULD NOT return their own Peer ID in responses to `FIND_NODE`
+queries. However, they MUST include information about the requester, if and
+only if the requester is a DHT Server in its routing table and it is among the
+`k` closest nodes to the target key.
```
guillaumemichel, line 265: In most cases, returning information about self (DHT Server) or requester (DHT Client) will be useless for the request, since the DHT Client already knows about both. The main argument of not including these addresses is to save bytes on the wire.

I don't see a use case where it would be useful that the DHT Server sends information about itself, since listen addresses and supported protocols should be exchanged using libp2p identify.

In some cases, it could be useful for the client (if it is a DHT Server) to know whether or not it is included in the Server's routing table, or which addresses are advertised. Information about the requester will essentially be present when a node is looking up itself when refreshing its routing table, and having this information provides a guarantee that the peer is actually routable. In a network that is large enough, it is very unlikely that the requester (being a DHT Server) will be among the `k` closest nodes to a key (other than self) being looked up. 

Alternatively is possible to get this information by starting a fresh DHT Client and requesting about the initial peer, but it is a bit cumbersome. 

also see https://github.com/libp2p/specs/issues/535

lidel (commented):
Thank you @guillaumemichel for documenting this :heart: 

Made a first pass and pushed small cosmetics + some questions / suggestions inline.

I think its ready for wider feedback, and if no concerns landing it in a few weeks.

src/routing/kad-dht.md
lidel: yeah, sadly `xref` system only works for web specs.  I've renamed section to "Bibliography" for now, and made manual `[0]` links – we will make this better at some point, but dont want t block this PR on it.

src/routing/kad-dht.md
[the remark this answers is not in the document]
guillaumemichel: Most of the "recommended" are for values/strategies that can be defined freely by implementers, they are not part of the protocol. In this case, "recommended" is more of a suggestion for values/strategies already used in implementations, e.g a safe default. Implementers are free to choose different values/strategies if they have other needs.

Hence the nuance should (not pun intended) be less than `SHOULD` but more than `MAY`. I think `SHOULD` is closer though.

I can either:
* Replace the occurrences with `SHOULD`
* Rephrase to _suggested_ or _safe default_.

src/routing/kad-dht.md
```diff
@@ -0,0 +1,700 @@
...
+
+The IPFS Kademlia DHT supports two types of records, each stored in its own
+keyspace:
+
+1. **Public Key Records** (`/pk/`) – Used to store public keys that cannot be
+   derived from Peer IDs.
+2. **IPNS Records** (`/ipns/`) – Used for decentralized naming and content
+   resolution.
+
+Records MUST meet validity criteria specific to their record type before being
```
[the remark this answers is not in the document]
guillaumemichel: Records with unknown prefixes are currently rejected ([source](https://github.com/libp2p/go-libp2p-kad-dht/blob/8c96f0ae5f746f16dad911f35f6cdacc202824e7/internal/config/config.go#L159-L178)).

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+`/ipfs/kad/1.0.0`.
+
+:::note
+The Amino DHT is utilized by multiple IPFS implementations, including
+[`kubo`](https://github.com/ipfs/kubo) and
+[`helia`](https://github.com/ipfs/helia)
+and can be joined by using the [public good Amino DHT Bootstrappers](https://docs.ipfs.tech/concepts/public-utilities/#amino-dht-bootstrappers).
+:::
+
+### Client and Server Mode
```
aschmahmann, line 93: Does this make sense in this spec rather than the libp2p one? Is the advice around client / server resource usage not generally applicable?

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+identify
+protocol](https://github.com/libp2p/specs/blob/master/identify/README.md). In
+addition DHT Servers MUST accept incoming streams using the libp2p Kademlia protocol
+identifier.
+
+DHT Clients MUST NOT advertise support for the libp2p Kademlia
+protocol identifier nor offer the libp2p Kademlia protocol
+identifier for incoming streams.
+
+## Kademlia Keyspace
```
aschmahmann, line 121: This is general, right?

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+in the network share shorter prefix lengths with the local node. Conversely,
+buckets beyond a certain index usually remain empty, since it's statistically
+unlikely that any node will have an identifier sharing a very long common
+prefix with the local node. For more information see [bucket population
+measurements](https://github.com/probe-lab/network-measurements/blob/master/results/rfm19-dht-routing-table-health.md#peers-distribution-in-the-k-buckets).
+
+The IPFS Kademlia DHT uses a bucket size of `k = 20`. This corresponds to the
+`k` value as defined in the original Kademlia paper [`[0]`](#bibliography). The `k` value is also
+used as a replication factor and defines how many peers are returned to a
+lookup request.
```
aschmahmann, line 172: This paragraph seems specific to IPFS, is the rest of this section?

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+implementation, rather than the one detailed here. This document specifically
+outlines all protocol customizations and adaptations required for participation
+in the [Amino DHT](#relation-to-the-amino-dht). If you're designing a new
+Kademlia-based DHT for use with IPFS, some details in this specification may
+appear overly specific or prescriptive.
+
+### Relation to the Amino DHT
+
+Nodes participating in the public [Amino DHT Swarm](#amino-dht) MUST implement the
+IPFS Kademlia DHT specification. The IPFS Kademlia DHT specification MAY be
```
aschmahmann, line 55: Why not just own that this document is the Amino DHT specification and that similar variants have been known to be deployed elsewhere (e.g. LANs, CJDNS, etc.) with different parameters (particularly where IP filtering, etc. comes into play).

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+* **IPv6 addresses** are grouped by ASN.
+* **IPv4 addresses** are grouped by `/16` prefixes, except for [legacy Class A
+blocks](https://en.wikipedia.org/wiki/List_of_assigned_/8_IPv4_address_blocks),
+which are grouped by `/8` prefixes.
+
+Since a single node can advertise multiple addresses, a peer MUST NOT be added
+to the routing table if any of its addresses already exceed the allowed
+representation within the table.
+
+### Routing Table Refresh
```
aschmahmann, line 213: This section seems general with the exception of any numbers being recommended.

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+recently connected peers or by executing a `FIND_NODE` [RPC
+message](#rpc-messages) with a randomly generated Peer ID matching the bucket.
+`FIND_NODE` requests should only be run for buckets up to the last non-empty
+bucket.
+
+Finally, the refresh process concludes by executing a `FIND_NODE` request for
+the local node's Peer ID, ensuring the DHT Server maintains up-to-date
+information on its closest peers.
+
+## Lookup Process
```
aschmahmann, line 236: Seems shared/general

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+and interact with nodes that are closest to a given key based on the network's
+distance metric. Once a node has identified the closest peers, it can either:
+
+* **Locate a specific peer** in the network
+* **Find content providers** serving content associated with a CID
+* **Store and retrieve values** directly within the DHT, such as IPNS names
+
+### Relation to libp2p kad-dht
+
+The IPFS Kademlia DHT specification is a specialization of the [libp2p Kademlia DHT](https://github.com/libp2p/specs/tree/master/kad-dht).
```
aschmahmann, line 43: It seems like an awful lot of this document isn't describing anything specialized or specific to IPFS or the Amino DHT and is general to the libp2p kad spec.

What's the approach we're trying to take here:
1. Only add IPFS / Amino -specific info
2. Copy-paste an independent document so IPFS / Amino users don't need to look at the libp2p kad spec
3. Make a mostly independent IPFS / Amino doc so users only have to look at the libp2p one for levels of detail exceeding some threshold
4. Something else

This paragraph makes it sound like 1 is the approach we're taking but the doc looks like 2 or 3.

src/routing/kad-dht.md
```diff
@@ -0,0 +1,719 @@
...
+It is possible to use an alternative DHT specification alongside an IPFS
+implementation, rather than the one detailed here. This document specifically
+outlines all protocol customizations and adaptations required for participation
+in the [Amino DHT](#relation-to-the-amino-dht). If you're designing a new
+Kademlia-based DHT for use with IPFS, some details in this specification may
+appear overly specific or prescriptive.
+
+### Relation to the Amino DHT
+
+Nodes participating in the public [Amino DHT Swarm](#amino-dht) MUST implement the
```
aschmahmann, line 54: Thoughts on how we should handle parts of the network that evolve over time?

For example:
- Maybe we mandate TCP + Yamux + Noise for all DHT server nodes, or we mandate TCP + Yamux + Noise and QUIC for all DHT client nodes and say servers only need to support one. We could then encourage with SHOULDs other transports like WSS, WT, WebRTC, etc. and change from SHOULD to MUST as we see adoption rise past a certain point where we'd feel comfortable having clients rely on the new behavior from the servers.
- Same if we decide to add new RPC methods / behaviors or change some server behavior without bumping the protocol version (e.g. IP / ASN limiting, etc.)

aschmahmann (commented):
Thanks for starting here, it's much needed 🙏. Most of my comments fall into:
1. What should be here vs libp2p kad spec
2. If we've covered all the things we need to here

src/routing/kad-dht.md
guillaumemichel: Technically, the Amino DHT is just a swarm of peers. It is an instantiation of this spec, and not a spec by itself. There could be other swarms of peers implementing the {IPFS Kademlia DHT, Amino DHT} spec, but they would NOT be the Amino DHT (the only difference would be the libp2p protocol identifier).

Also, some details about the LAN DHTs are included as part of this spec, since LAN DHTs are (or should be) part of the IPFS Specs. There is a large overlap in the behavior of the Amino DHT with the LAN DHTs (only difference is probably addresses filtering).

If it makes things clearer, I can rename this doc "Amino DHT Spec", and specify that using this spec and changing the libp2p protocol identifier is possible and would create a new DHT swarm. We could then add a dedicated LAN DHT spec, describing the diff with the Amino DHT Spec.

src/routing/kad-dht.md
guillaumemichel: Good points!

* I will add transports requirements
* For new RPC methods/behaviors, I suggest we deal with it in a subsequent PR/IPIP since it will require some discussion

src/routing/kad-dht.md
lidel: There is a benefit of keeping it as "IPFS Kademlia DHT" ans then showing it can multiple independent instances/swarms. 

We really want the spec to show there is a surface for rolling their own DHT swarm by deploying own protocol identifier, to avoid accidental deployments of useless nodes to Amino (real risk, we had at least one incident in recent years).

I think we should clarify "libp2p Protocol Identifier" section by moving "Relation to" sections under it, and:
- rephrase "Relation to Amino DHT" section to note "Amino DHT" is an instance of the "IPFS Kademlia DHT" spec mounted under `/ipfs/kad/1.0.0` libp2p protocol
- add "Relation to the Amino LAN DHT" section and note that LAN-specific version of Amino is 
mounted under `/ipfs/lan/kad/1.0.0` libp2p protocol
- add section "Creating Custom DHT Swarm"  and suggest convention `/<swarm-prefix>/kad/<version>` there

src/routing/kad-dht.md
lidel: I think this is the similar tradeoff to "RPC messages" – duplicating important content here is intentional, aims to improve readability, creates coherent document, and protects us from feature creep.

The need for jumping from spec HTML into markdown in libp2p/specs repo should be minimized.

Long term, protection from unintended feature creep is important:  someone changes libp2p specs and adds new RPCs, we dont want that to magically apply to Amino DHT.

As prior art example, I copied part of PeerID specs and placed them in https://specs.ipfs.tech/ipns/ipns-record/#ipns-keys to ensure that if libp2p adds new key type, or extends protobuf, that key type does not automatically become part of IPNS specs, but needs an explicit IPIP to add it.

src/routing/kad-dht.md
lidel: I'm ok with 2-3, for reasons listed in https://github.com/ipfs/specs/pull/497#discussion_r2019187655

Perhaps go further and change this paragraph to state "The IPFS Kademlia DHT specification extends [libp2p Kademlia DHT](https://github.com/libp2p/specs/tree/master/kad-dht) with features related to CID, IPNS, and content providing."

src/routing/kad-dht.md
guillaumemichel: Clarified in https://github.com/ipfs/specs/pull/497/commits/cd85a7f92069fc41210e1cb2dfaabaf1d14d6507.

Note that I left the "Relation to" sections since the goal is to inform the reader early about the nature of this document vs notion they may already have of the libp2p/Amino DHT. Happy to change if confusing.

src/routing/kad-dht.md
guillaumemichel: https://github.com/ipfs/specs/pull/497/commits/27d2a5825816422c50ca1f462ddb4355784fc982

src/routing/kad-dht.md
```diff
@@ -0,0 +1,783 @@
...
+};
+
+message Message {
+    enum MessageType {
+        PUT_VALUE = 0;
+        GET_VALUE = 1;
+        ADD_PROVIDER = 2;
+        GET_PROVIDERS = 3;
+        FIND_NODE = 4;
+        PING = 5;
```
dozyio, line 659: Maybe add a comment mentioning that ping is deprecated

src/routing/kad-dht.md
```diff
@@ -0,0 +1,787 @@
...
+    // GET_VALUE, ADD_PROVIDER, GET_PROVIDERS
+    repeated Peer providerPeers = 9;
+}
+```
+
+These are the requirements for each `MessageType`:
+
+* `FIND_NODE`: In the request `key` must be set to the binary `PeerId` of the
+node to be found. In the response `closerPeers` is set to the DHT Server's `k`
+closest `Peer`s.
```
achingbrain, line 721: Can the DHT server include itself in the `closerPeers` response field?

Can it include the requester?

src/routing/kad-dht.md
```diff
@@ -0,0 +1,787 @@
...
+datastore `record` is set to the value for the given key.
+
+* `PUT_VALUE`: In the request `record` is set to the record to be stored and
+`key` on `Message` is set to equal `key` of the `Record`. The target node
+validates `record`, and if it is valid, it stores it in the datastore and as a
+response echoes the request.
+
+* `GET_PROVIDERS`: In the request `key` is set to the multihash contained in
+the target CID. The target node returns the known `providerPeers` (if any) and
+the `k` closest known `closerPeers`.
```
achingbrain, line 734: Same question as https://github.com/ipfs/specs/pull/497/files#r2092857416 about what peers can be in the `closerPeers` field

src/routing/kad-dht.md
achingbrain: Answered one of my questions:

> Can it include the requester?

No.  If this comment is accurate it's worth documenting - https://github.com/libp2p/go-libp2p-kad-dht/blob/010ef38a342f749a6fe380d2b67e651767850f3a/handlers.go#L265

achingbrain, line 721: go-libp2p-kad-dht appears to return [peers that are closer than the requester](https://github.com/libp2p/go-libp2p-kad-dht/blob/010ef38a342f749a6fe380d2b67e651767850f3a/handlers.go#L263C41-L263C48) so we should document this:

```suggestion
closest `Peer`s that are closer to the `key` than the requester's `PeerId`.
```

guillaumemichel: https://github.com/ipfs/specs/blob/802fd5547da496819f1f9217514dcb9a7e054fa8/src/routing/kad-dht.md?plain=1#L341-L344

src/routing/kad-dht.md
guillaumemichel: https://github.com/ipfs/specs/blob/802fd5547da496819f1f9217514dcb9a7e054fa8/src/routing/kad-dht.md?plain=1#L341-L344

If it helps, I can duplicate the information at the line you mentioned here.

src/routing/kad-dht.md
guillaumemichel: The suggested change doesn't seem right.

When a peer is looking for its closest peers, it will look for its own key. In this case no DHT server will be closest to the requested key than the requester itself. Nevertheless, the requester expects to discover its closest peers.

The response should contain the absolute `k` (number of peers to be returned during lookup) closest peers to the requested key (with exception of self (DHT server), and requester).

IIUC `go-libp2p-kad-dht` doesn't limit the returned peers to those closer than the requester (to the requested key).

src/routing/kad-dht.md
guillaumemichel: Now I think that including the DHT server is mostly useless and takes space on the wire. It isn't useful when looking for one specific key, or the `X` closest peers to a key.

The only use case I can see if for when DHT servers try to see whether they are included in another DHT server's routing table, and what information is stored about them. This would be useful only for crawlers to gain statistics on the network. The exact same information can be retrieved using another peer id.

Hence I would suggest that the `closestPeers` field shouldn't contain neither self (DHT server), nor the requester's information. It self AND/OR requester are among the `k` closest peers to the requested key in the DHT server's routing table, then they should be replaced by the next closest peer, so that the response always contains `k` peers (if network is large enough).

It means that the current implementations may not be compliant with this spec change, but it should be easy to address.

WDYT @achingbrain ?

Related issues:
* https://github.com/libp2p/specs/pull/609
* https://github.com/libp2p/specs/pull/644

src/routing/kad-dht.md
achingbrain: > IIUC go-libp2p-kad-dht doesn't limit the returned peers to those closer than the requester (to the requested key).

Ah yes, it just passes the requester's peer id in to filter it out and make sure it doesn't tell the requester about itself.

src/routing/kad-dht.md
guillaumemichel: Tried to clarify in https://github.com/ipfs/specs/pull/497/commits/23d4b91f904cd0cd909064417bd9cad9e1f1dcbd.

> iiuc, the current go-libp2p-kad-dht behavior:
> 
>     1. ✅ Already excludes requester - The code has if targetPid != from check

That is correct.

>     2. ❌ Does NOT exclude self - afaik there is no explicit check to exclude the DHT server's own ID?

Self is never returned by the routing table when looking for closest peers. See error if it happens [here](https://github.com/libp2p/go-libp2p-kad-dht/blob/548e361275f0a38bb219c218b1b630b49a9920c8/dht.go#L756-L783). The only situation where a provider record could be returned for self, is if self's Peer ID is directly requested.

>     3. ❌ No special handling for non-DHT peers - doesn't distinguish between DHT servers and clients in responses?

The protobuf message only has a single field to return peers. We cannot modify it, since it would be a breaking change. The only DHT client that can ever be returned in a response is if the requested Peer ID is a perfect match of a peer that is in the peerstore, but isn't a DHT Server (in the routing table). See [here](https://github.com/libp2p/go-libp2p-kad-dht/blob/548e361275f0a38bb219c218b1b630b49a9920c8/handlers.go#L266-L283).

> Should we approve spec clarification here + fix reference implementation of kad-dht in go and js for consistency?

https://github.com/libp2p/go-libp2p-kad-dht/pull/1158

src/routing/kad-dht.md
guillaumemichel: see https://github.com/ipfs/specs/pull/497#discussion_r2016997925

src/routing/kad-dht.md
guillaumemichel: https://github.com/ipfs/specs/pull/497#discussion_r2016997925

src/routing/kad-dht.md
guillaumemichel: https://github.com/ipfs/specs/pull/497#discussion_r2016997925

PR #504, opened 2025-05-26
IPIP-504: `provider` query parameter as hint for HTTP Gateways

vasco-santos:
Based on initial exploration https://github.com/vasco-santos/provider-hinted-uri/blob/main/EXPLORATION.md

### DIFF
--- a/IPIP/0504-provider-query-parameter.md
+++ b/IPIP/0504-provider-query-parameter.md
@@ -156,9 +156,98 @@ Once a CID has been successfully extracted, clients MAY parse `provider` paramet
   - Ignore all `provider` parameters (if unsupported).
   - Evaluate hints in order of appearance (left-to-right).
   - Evaluate hints in parallel.
-  - Apply their own prioritization or fallback strategies. If all hints fail, clients SHOULD fall back to default discovery strategies (e.g., DHT/IPNI), if available. Or even rely on discovery strategies in parallel.
 
 Note that the `multiaddr` string should point to the `origin` server where given CID is provided, and not include the actual CID in the Hint multiaddr as a subdomain/path.
+
+### Provider Hint as complement to Regular Routing
+
+#### Overview
+
+The `provider` query parameter is designed to enhance content retrieval by providing optional hints to clients about where content might be available. However, the order in which these hints are evaluated relative to regular routing systems (DHT, IPNI, etc.) MAY have implications for the long-term health of user experience, in the event that providers do not provide given content anymore, while the content is available in other providers.
+
+---
+
+#### Strategies and their Implications
+
+**Strategy A: Provider Hints First (Sequential Fallback)**
+
+In this approach, clients attempt to fetch content from `provider` hints first, only falling back to regular routing systems if all hints fail.
+
+**Implementation:** `provider` → (on failure) → regular routing
+
+**Trade-offs:**
+- 🔴 **User-hostile latency:** Every access to a URL with outdated or offline provider hints incurs unnecessary delay while waiting for failed requests to timeout
+- 🔴 **Economic hostage situation:** Creates perverse incentives where someone must continue paying for storage at specific providers to maintain good user experience if they can't update the URLs, even if better alternatives exist in the network
+- 🔴 **Permanent degradation:** If a hardcoded provider goes offline, the latency penalty persists forever
+- 🔴 **Discovery system stagnation:** Reduces incentives to improve and invest in decentralized routing systems since they're only used as a "last resort"
+- ✅ **Non-interactive retrievals are possible:** Content MAY be retrieval without any extra request
+- ✅ **Immediate latency benefits in cold-start**
+
+**Example scenario:** A viral piece of content is shared with `?provider=https://expensive-storage.com`. The content becomes embedded in blog posts, Wikipedia citations, and web crawlers index it. When the publisher stops paying for storage, every subsequent access experiences a timeout delay before falling back to the DHT/IPNI.
+
+---
+
+**Strategy B: Parallel Evaluation**
+
+In this approach, clients simultaneously attempt to fetch from provider hints and query regular routing systems, using whichever responds first.
+
+**Implementation:** `provider` ∥ regular routing → use fastest response
+
+**Trade-offs:**
+- 🟠 **Wasted bandwidth:** Always sends requests to potentially offline providers, consuming bandwidth and resources
+- 🟠 **Provider overhead:** Storage providers continue receiving requests for content they no longer host, creating ongoing operational costs
+- ✅ **No latency penalty:** Users don't experience delays from failed hint attempts
+- ✅ **Graceful degradation:** Still works when hints become stale
+
+---
+
+**Strategy C: Provider Hints as Fallback**
+
+In this approach, clients first attempt regular routing discovery, only trying provider hints if the routing system fails to find suitable providers or after a reasonable delay.
+
+**Implementation:** regular routing → (on failure or delay) → `provider` hints
+
+**Trade-offs:**
+- 🔴 **Non-interactive retrievals are not possible:** Several hops are required before bytes can start to be served
+- 🔴 **No immediate latency benefits in cold-start**
+- 🔴 **Requires robust timeout and retry logic:** Ensuring hints are eventually tried
+- ✅ **Ecosystem-friendly:** Reduces unnecessary load on hardcoded providers
+- ✅ **Encourages routing investment:** Maintains incentives to improve DHT, IPNI, and other discovery mechanisms
+- ✅ **Resource efficient:** Minimizes wasted requests to offline providers
+- ✅ **Natural load balancing:** Distributes requests across the network rather than concentrating them on hinted providers
+
+---
+
+#### Specification Recommendations
+
+The recommended approach (Strategy C) balances several competing concerns:
+
+1. **Long-term ecosystem sustainability:** By not making provider hints the primary resolution mechanism, we avoid creating dependencies on specific commercial providers
+2. **Decentralization preservation:** Regular routing systems remain the primary mechanism, preserving IPFS's decentralized nature
+3. **Economic fairness:** Prevents situations where specific entities bear disproportionate costs for serving popular content
+4. **User experience:** While not optimizing for immediate latency in all cases, provides better long-term user experience by avoiding permanent degradation
+
+---
+
+**Gateway Implementations**
+
+Gateway implementations that support the `provider` query parameter:
+
+- **SHOULD** evaluate provider hints as a fallback mechanism after attempting regular routing discovery
+- **MAY** implement parallel evaluation if they document the associated costs and implement appropriate rate limiting
+- **MUST NOT** implement provider-first evaluation as it MAY create long-term ecosystem harm
+
+---
+
+**Client Implementations**
+
+Client implementations that support the `provider` query parameter:
+
+- **SHOULD** prioritize regular routing systems (DHT, IPNI) over provider hints to maintain ecosystem health
+- **MAY** use provider hints in parallel with regular routing if they implement proper resource management
+- **MAY** allow a different strategy to be used via configuration out of the default path
+- **SHOULD** implement reasonable timeouts and retry logic when falling back to provider hints
+- **MUST** continue to function correctly when provider hints are ignored or fail
 
 ---
 
@@ -213,16 +302,6 @@ This distinction allows URI publishers to tailor behavior:
 Publishers of such URIs should consider the **security profile** and **trust assumptions** of their environment when deciding how to encode hints.
 
 This flexibility supports a spectrum of use cases—from fully local client-side fetch strategies to cooperative client-server resolution pipelines.
-
-## Design rationale
-
-TODO
-
-The rationale fleshes out the specification by describing what motivated
-the design and why particular design decisions were made.
-
-Provide evidence of rough consensus and working code within the community,
-and discuss important objections or concerns raised during discussion.
 
 ### User benefit
 
--- a/IPIP/0504-provider-query-parameter.md
+++ b/IPIP/0504-provider-query-parameter.md
@@ -129,15 +129,26 @@ If a CID is present in both a multi-dotted origin and in the path (even if they 
 - Value: Multiaddr string (`?provider=multiaddr`).
 - Interpretation: Optional hint for how to fetch and locate the content identified by the CID
 
+ Name: `provider`
+- Type: URI Query Parameter (repeating allowed)
+- Value: Either a Multiaddr string (`?provider=multiaddr`) or HTTP URL string (`?provider=http-url`) that can be transformed to Multiaddr.
+- Interpretation: An optional hint for how to locate and fetch the content identified by the CID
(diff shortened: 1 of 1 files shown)


IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,242 @@
...
+**4. Conflict Resolution**
+
+If a CID is present in both a multi-dotted origin and in the path (even if they match), the URI MUST be rejected as ambiguous.
+
+---
+
+#### Query Parameter: `provider`
+
+- Name: `provider`
+- Type: URI Query Parameter (repeating allowed)
```
SgtPooki, line 128: why repeating parameter instead of comma-delimited?

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,242 @@
...
+- Each `provider` parameter MUST be treated as an independent, optional provider hint.
+- Clients MAY ignore hints with invalid multiaddrs.
+
+**2. Evaluate hints**
+
+- Clients MAY:
+  - Ignore all `provider` parameters (if unsupported).
+  - Evaluate hints in order of appearance (left-to-right).
+  - Evaluate hints in parallel.
+  - Apply their own prioritization or fallback strategies. If all hints fail, clients SHOULD fall back to default discovery strategies (e.g., DHT/IPNI), if available. Or even rely on discovery strategies in parallel.
```
SgtPooki, line 148: > Or even rely on discovery strategies in parallel.

I feel like if there are provider hints, we SHOULD attempt to process them before IPNI/DHT to prevent any negative cascading network effort

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,242 @@
...
+
+Note that the `multiaddr` string should point to the `origin` server where given CID is provided, and not include the actual CID in the Hint multiaddr as a subdomain/path.
```
SgtPooki, line 150: Are CIDs valid in multiaddr at all (besides Peer ID)? might be worth linking to multiaddr spec: https://github.com/libp2p/specs/blob/master/addressing/README.md#multiaddr-in-libp2p

SgtPooki (commented):
first read through.. makes sense and seems like it could really benefit the ecosystem. One problem I can imagine is that people start encoding dynamic IP address providers in urls and they quickly become useless, so we should probably call that out.

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,242 @@
...
+tags: ['ipips']
+---
+
+## Summary
+
+A URI-based format for expressing content-addressed identifiers (such as IPFS CIDs) optionally augmented with one or more provider hints. This format aims to support a simple, unopinionated, transport-agnostic scheme to simplify data retrieval in content-addressable systems by introducing a clear, extensible, and broadly compatible URI format.
+
+## Motivation
+
+Content-addressable systems, such as IPFS, allow data to be identified by the hash of its contents (CID), enabling verifiable, immutable references. However, retrieving content typically relies on side content discovery systems (e.g. DHT, IPNI), even when a client MAY know one (or more) provider of the bytes. A provider in this context is any node, peer, gateway, or service that can serve content identified by a CID.
```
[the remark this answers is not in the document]
2color: I've seen many situations in which you want to share content addressed data with all the benefits of verification and p2p without necessarily caring about long-term persistence. If we don't formalise this in a spec, we end up pushing users to solving this in user/application space (see [example](https://github.com/ipfs-shipyard/ipfs-share-files/pull/206), leading to fragmentation, and no ecosystem benefits from a conventional approach.

> Application developers can hard-code additional routing systems into their applications. For example, if the developers for a given dApp are already hosting all their data and/or paying a pinning service to host the data saying to check that endpoint first seems fine and is inline with how many of them already operate by hard-coding a gateway endpoint provided by the pinning service they pay for storage

This would typically be done using the Delegated Routing API for which we have a lot of tooling and support in implementations. Practically, this which would involve setting a application specific endpoint somewhere, which — just like a specific provider maddr — can go down, or become stale eventually. But for as long as that delegated routing endpoint is up, it helps the app can map CIDs to provider maddrs. 

It seems to me that insisting that this is the recommended to do it, over encoding it along with the CID is overly pedantic, given that the two approaches are not all that dissimilar, with the one exception that your mixing "permanent" information, i.e. the CID, with impermanent, i.e. a specific provider maddr. 

> URIs become ephemeral. While encoding ipfs://bafyfoo used to not be ephemeral ipfs://bafyfoo?provider=<pinning-service-that-had-the-cid-when-the-link-was-made> is now ephemeral.

This is the crux of the debate here; and ultimately a question of where this boundary between ephemeral and permanent should be delineated. I happen to think it maps elegantly to query parameters. 

But more broadly, content routing is hard and adds a performance tax that undercuts adoption in scenarios where the benefits of verification and content-addressing are desired. 

Optimising for successful retrieval by CID should be a higher level goal, and think that the broad strokes proposed here advance this goal. Moreover, it paves a path for _incremental adoption_ of content addressing.

So my take is that that this we should just encode all of these caveats into this IPIP, e.g. strongly recommend against persistence of the provider hint in long term storage like on-chain.

IPIP/0504-provider-query-parameter.md
vasco-santos: Why a multiaddr with additional info could be a problem for DHT/IPNI? They should not try to match by multiaddr or anything like that. 
Anyway, there would be no reason to remove this info before passing over to those systems if that is a concern and they do not support this

IPIP/0504-provider-query-parameter.md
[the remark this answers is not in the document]
vasco-santos: Thanks for writing this @2color . This is exactly what I feel like! I wrote some use cases that can benefit directly from this https://github.com/vasco-santos/provider-hinted-uri/blob/main/EXPLORATION.md#use-cases

Note that some of them are actually depending on a follow up of this (adding `tags`) to the multiaddr, which for now is a different conversation, as also `tags` are optional ways to expand on this even further

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,242 @@
...
+
+If a CID is present in both a multi-dotted origin and in the path (even if they match), the URI MUST be rejected as ambiguous.
+
+---
+
+#### Query Parameter: `provider`
+
+- Name: `provider`
+- Type: URI Query Parameter (repeating allowed)
+- Value: Multiaddr string (`?provider=multiaddr`).
```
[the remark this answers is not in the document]
vasco-santos: I find surprising that `multiaddr` is assumed to be a concept of libp2p. I see space for `multiaddr` usage outside of libp2p as any other multiformat. 

Anyway, after syncing with @bumblefudge last week, I totally agree that we should also accept `string`  provider, while keeping `multiaddr` if starting by `/`.

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,265 @@
...
+- If the `provider` is encoded as a **fragment** (`#...`), it is only accessible to the client (browsers do not send fragments to the server).
+
+This distinction allows URI publishers to tailor behavior:
+
+- **Client-only mode:** Use a fragment (`#provider=...`) to ensure the server remains unaware of hint data. This is useful for privacy-preserving client apps or when hints are intended to guide only the client.
+- **Server-assisted mode:** Use query parameters (`?provider=...`) to allow the server to parse and act on provider hints. This may enable proxy behavior, similar to existing IPFS gateways like `ipfs.io` or `dweb.link`.
+
+Publishers of such URIs should consider the **security profile** and **trust assumptions** of their environment when deciding how to encode hints.
+
+This flexibility supports a spectrum of use cases—from fully local client-side fetch strategies to cooperative client-server resolution pipelines.
```
SgtPooki, line 253: Let's call out client security concerns here as well. i.e. If malicious or nosy tracking entity (nft marketplace?) starts encoding provider hints into all their URLS, every single client can be easily tracked. Another point for not prioritizing provider hints over routing system as called out by @lidel.

IPIP/0504-provider-query-parameter.md
```diff
@@ -0,0 +1,265 @@
...
+  - Enables seamless data exfiltration or migration from providers that do not participate in public discovery systems (e.g., not on Amino DHT or cid.contact). - Useful during outages or vendor lock-in scenarios.
+- **Custom Provider Prioritization**
+  - Developers can craft links that prefer self-hosted providers (e.g., local gateways, corporate infra) before falling back to public gateways or network discovery.
+
+#### 🚀 Performance-Sensitive Applications
+
+- **Reduced Latency on First Fetch**
+  - By embedding direct provider hints, clients can skip discovery lookups and go straight to fetching bytes—significantly lowering time to first byte, especially in cold-start scenarios.
+-	**Faster Initial Seeding**
+  - Clients like IPFS Desktop can opportunistically add hints to links they generate (e.g., using their peer ID or trusted gateway), enabling faster and more deterministic bootstrapping when others use those links.
```
lidel, line 223: :thought_balloon: alternative approach for this use case is for IPFS Desktop to do manual provide on Amino DHT of the CID that is being shared. This would not require hardcoding anything in shared URL.

IPIP/0504-provider-query-parameter.md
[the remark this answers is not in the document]
lidel: > What feels more accurate to me is to allow clients to decide implementation details, or to allow configuration of this, rather than the spec mandate a certain implementation behaviour.

Reality is, leaving implementation details of "evaluation order" out of spec **IS effectively leading to (A):**  even if someone initially implements (C), over time, motivated people will eliminate "annoying latency" by switching client code from (C) to (B), and then to "remove unnecessary lookups" switch from (B) or (A).

Next steps:

- [ ] the IPIP should clearly address risks that come with evaluation order choice, and inform reader about the externalized ecosystem cost listed above ( :orange_circle:  :red_circle: ). 
   -  Do not pretend downsides do not exist, but document them and make a compelling case in "User benefit" the benefits outweigh the identified cost for IPFS ecosystem long term. 
- [ ] do not ignore the topic of evaluation order, but address it in the spec head-on: 
  - a "SHOULD" recommendation for (C) with rationale of :green_circle: 
  - a  "MAY" for (B) with  :orange_circle: cost documented 
  - a "MUST NOT" for (A) with :red_circle: risks documented as reason why it is considered harmful
    - we can't block people from YOLOing, but with this, whatever implementation or part of ecosystem want to shut themselves and their users in the foot long erm, we can always point at spec they ignored

IPIP/0504-provider-query-parameter.md
vasco-santos: This is a good point, I will add it. Thanks @SgtPooki

IPIP/0504-provider-query-parameter.md
vasco-santos: Included this in https://github.com/ipfs/specs/pull/504/commits/3f6c1315ae037d324518b707047ead21194d04db

IPIP/0504-provider-query-parameter.md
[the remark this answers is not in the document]
vasco-santos: @lidel @SgtPooki just added this https://github.com/ipfs/specs/pull/504/commits/5382b97e554f9ceff36eda47d45e8ed7d291b367

Let me know if this is a good compromise on the direction for you

IPIP/0504-provider-query-parameter.md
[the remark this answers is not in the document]
aschmahmann: Thanks @vasco-santos for flagging the use cases. As I think I've called out the risk I'm most concerned with here is degradation of the ecosystem of URIs that have for the last 10 years been associated with "mainnet". Breaking them down into your use cases:
1. IPFS Content Provider: 🔴Problematic, as it is intentionally about not using content routing "clients have an option to reach content without relying on discovery systems"
2. Skipping Multi-hop discovery:  🔴 Seems mostly the same as IPFS Content Provider. However, if the issue is related to wanting to save a protocol negotiation round-trip this data can also just be included in the routing layer (e.g. IPNI providers have this today)
3. Pre-seeding & Edge Caching: 🟢 From what I can tell about this use case it has very little to do with mainnet (e.g. it's about internal nodes and applications not really tied to mainnet) and so doesn't interfere with my concerns
4. Subresource Integrity for Static Servers: 🟢This is unrelated to mainnet, as it's intention is to fetch from a single location. Note: This also seems to effectively match @2color's example.

> I've seen many situations ... If we don't formalise this in a spec, we end up pushing users to solving this in user/application space ... leading to fragmentation, and no ecosystem benefits from a conventional approach.

A thought on a potential compromise here that may address my concerns while still helping with the problem quoted above.

- Let this exist as effectively an RFC rather than a spec change. That we're working in an HTTP context (and opened the door to this around trustless-gateway behavior) seems to effectively mean that IPFS users are free to publish whatever custom headers, query parameters, url fragments, ... they want and associate with them whatever semantics they want where at a project / spec level.
     - This allows us to treat this a little like areas of the multiformats repos where we could be looser and mostly concern ourselves with name conflict management and helping implementers figure out what they likely want to implement (e.g. nobody thinks base256emoji is mandatory to implement, but they know the prefix is taken and what it means)
- Note in the RFC that this is not recommended for public network IPFS gateways such as mainnet

---

An additional / alternative compromise introduced to me by @Stebalien:

Since one of the major issues I flagged is the links becoming effectively ephemeral, introducing vendor lock-in, etc. this can largely be side-stepped by allowing introducing these hints at the DNS(Link) layer. Since those records are largely mutable anyway the risks are much lower.

The risks / downsides are still non-zero. For example, users in ENS land are using immutable names for versioning in ways that again resurface the problems documented here. However, it is IMO less bad than simply allowing the ephemeral links and location dependence in the immutable IPFS part of the stack.

IPIP/0504-provider-query-parameter.md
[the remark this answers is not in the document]
lidel: Update on the hypothetical above: I picked https://storacha.network as one of the example storage providers, and less than a year later it happened for real. Storacha wound down their IPFS service, stopped announcing content to the network on May 1, and w3s.link stopped serving on May 31.[^1]

The CIDs survived, that is the whole point of content addressing, and the data could be repinned elsewhere. A hint baked into a URL cannot. Every `?provider=` pointing at a dead endpoint, once it lands in a blogpost, a wiki citation, an NFT record or a crawler's index, becomes a permanent instruction to try a host that will never answer again. The cost is paid in latency by readers, and in junk traffic by whoever ends up owning that domain.

Not asking to close this, the idea still has value. But for whoever picks it up: providers going away is the normal end state of an endpoint, not an edge case, so **the evaluation order (A/B/C above) belongs in the spec text rather than in each implementer's judgement**.

[^1]: [An Update on Storacha, and Important News for You and Your Data](https://web.archive.org/web/20260503081756/https://medium.com/@storacha/an-update-on-storacha-and-important-news-for-you-and-your-data-15a5d10b7da0), April 2026: uploads disabled April 15, content no longer discoverable on the IPFS network from May 1, w3s.link gateways stopped serving May 31.

PR #507, opened 2025-06-16
docs: add paragraph describing HAMT fanout limit

achingbrain:
I came across [this comment](https://github.com/ipfs/boxo/blob/364c5040ec91ec8e2a61446e9921e9225704c34d/ipld/unixfs/hamt/hamt.go#L778) in the boxo source code.

It's worth documenting this constant so other implementations don't expose themselves to https://github.com/advisories/GHSA-q264-w97q-q778 accidentally.


achingbrain:
Looks like lots of bits of the spec fail the linter

PR #515, opened 2025-10-06
fix(ipns): missing xref

guillaumemichel:


guillaumemichel:
Error `No bibliographic reference for key "ipns-pubsub-router".` isn't fixed in the preview.

However, it seems that the build needs to run twice for circular cross-references to work:
  1. First build: Creates references.json with all spec metadata, but can't resolve cross-references yet
  2. Second build: Uses the existing references.json to resolve cross-references properly

PR #518, opened 2025-10-13
IPIP-518: URIs in Routing V1 API via Generic Schema

### LINKED ISSUE #377: Specify how API from IPIP-337 can support new protocols

lidel:
This IPIP extends `/routing/v1` HTTP API to allow HTTP(S) URLs alongside multiaddrs in `Addrs` field in a new Record type defined by `Schema=generic`.

Rationale in the IPIP document, TLDR is that allowing URLs will improve interoperability, and also remove hard dependency on Libp2p specifications (Multiaddr), allowing IPFS Ecosystem to implement HTTP-only stack where it makes sense, removing conversion step that is error prone.

Prototype implementation in boxo:
- https://github.com/ipfs/boxo/pull/1051

cc @aschmahmann @achingbrain @hsanjuan @gammazero  for visibility / initial feedback

## TODO

- [x] open PR with IPIP and gather feedback from community and stakeholders
- [ ] implement URI support in Boxo: https://github.com/ipfs/boxo/pull/1051
- [ ] implement URI support in Helia: TODO PR
- [ ] Boxo and Helia support shipped
- [ ] Kubo and [helia-verified-fetch](https://github.com/ipfs/helia-verified-fetch) support shipped
- [ ] [Someguy](https://github.com/ipfs/someguy) support shipped and deployed to [delegated-ipfs.dev](https://docs.ipfs.tech/concepts/public-utilities/#delegated-routing-endpoint)
- [ ] https://inbrowser.link support shipped
- [ ] ratify (merge this PR)

-----

Closes #377

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "name": "ipfs-specs-website",
       "version": "1.0.0",
       "dependencies": {
-        "spec-generator": "^1.6.1"
+        "spec-generator": "^1.7.0"
       }
     },
     "node_modules/@11ty/dependency-tree": {
@@ -5201,6 +5201,16 @@
         "markdown-it": "bin/markdown-it.js"
       }
     },
+    "node_modules/markdown-table": {
+      "version": "3.0.4",
+      "resolved": "https://registry.npmjs.org/markdown-table/-/markdown-table-3.0.4.tgz",
+      "integrity": "sha512-wiYz4+JrLyb/DqW2hkFJxP7Vd7JuTDm77fvbM8VfEQdmSMqcImWeeRbHwZjBjIFki/VaMK2BhFi7oUUZeM5bqw==",
+      "license": "MIT",
+      "funding": {
+        "type": "github",
+        "url": "https://github.com/sponsors/wooorm"
+      }
+    },
     "node_modules/marked": {
       "version": "12.0.2",
       "resolved": "https://registry.npmjs.org/marked/-/marked-12.0.2.tgz",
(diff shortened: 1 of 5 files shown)


aschmahmann (changes requested):
Thanks for the update here this looks pretty good to me, left a few comments / questions

PR #538, opened 2026-06-08
Edge IPFS

darobin:
Just playing around with this idea.

### DIFF
--- a/src/http-gateways/edge-ipfs.md
+++ b/src/http-gateways/edge-ipfs.md
@@ -1,0 +1,29 @@
+---
+title: Edge IPFS
+description: >
+  This is a lightweight approach to getting the benefits of IPFS — sharing verifiable
+  content – without having to run a long-lived node or taking part in complex,
+  stateful interactions.
+date: 2026-06-08
+maturity: wip
+editors:
+  - name: Robin Berjon
+    email: robin@berjon.com
+    url: https://berjon.com/
+    github: darobin
+    bluesky: "robin.berjon.com"
+    affiliation:
+        name: Supramundane Agency
+        url: https://supramundane.agency/
+tags: ['httpGateways', 'lowLevelHttpGateways']
+---
+
+Edge IPFS is a lightweight approach to getting the benefits of IPFS — sharing verifiable content –
+without having to run a long-lived node or taking part in complex, stateful interactions.
+It is intended to provide a smooth, simple path to adoption.
+
+The target use case is providers who want their content both addressable and discoverable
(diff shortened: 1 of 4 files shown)


src/http-gateways/edge-ipfs.md
```diff
@@ -0,0 +1,55 @@
...
+The target use case is providers who want their content both addressable and discoverable
+on IPFS, but only need to serve it (not coordinate with the network in real time). For
+these users, simplicity and low resource overhead matter more than full IPFS feature parity.
+This solution also works well for people who want to retrieve content from such providers
+with a very lightweight client implementation (doing very little more than HTTP).
+
+## Extension To Delegated Routing
+
+Making content discoverable can be achieved by offering a delegated routing endpoint
+([[http-routing-v1]]). An Edge IPFS server responds to routing requests for CIDs
```
2color, line 34: I would be more specific here about the endpoint only implementing section 4 of the spec, i.e., the [Content Routing API](https://specs.ipfs.tech/routing/http-routing-v1/#content-routing-api).

src/http-gateways/edge-ipfs.md
```diff
@@ -0,0 +1,55 @@
...
+that it knows about with a peer routing response the `Protocols` array of which
+must have a `rasl` entry.
+
+Correspondingly, there must then be a `rasl` field in the peer response object,
```
2color, line 38:
```suggestion
Correspondingly, there must then be a `rasl` field in the Peer Schema object,
```

[And link to it](https://specs.ipfs.tech/routing/http-routing-v1/#peer-schema)


One thing that's slightly awkward is that the peer-schema requires an ID:

> ID: the [Peer ID](https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md) as Multihash in Base58btc or CIDv1 with libp2p-key codec.

So the question is whether we want to require every RASL server to also have a key, or to work on adjusting the spec. I'm sure the latter will get push back, given that this assumption is baked in to a bunch of implementations — but still worth having the discussion, given that with HTTP, you are most likely relying on TLS and certificate authorities for transport encryption, rather than key based authentication, transport encryption.

lidel: May be a good opportunity to switch to a new schema that does not depend on libp2p concepts (peerids, multiaddrs), and allows returning provider info as URIs:

-  https://github.com/ipfs/specs/pull/518

mishmosh:
Per @lidel there is prior art in these 2 IPIPs, which offer a similar solution from the bottom up:

https://github.com/ipfs/specs/pull/504
https://github.com/ipfs/specs/pull/518

PR #542, opened 2026-07-28
delegated routing: resolve DNSADDR addresses before filtering

achingbrain:
Delegated HTTP Routing clients can specify address filters to limit returned providers to ones that listen on transports supported by the client.

`DNSADDR` addresses are a special case in that they can resolve to multiple addresses that may contain tuples from the filter, and the starting address is not likely to contain them.

This IPIP suggests adding a line to the spec that reminds the implementer they should resolve `DNSADDR` addresses to the final set before applying any address filters as otherwise valid, dialable providers would be omitted from the response.


lidel (approved):
This is more like a clarification of filtering mechanism (so LLMs make less mistakes) than spec change, so ok to merge without IPIP

go impl:
- https://github.com/ipfs/someguy/pull/174

lidel (approved):

PR #546, opened 2026-08-15
docs: require CAR fixtures for CIDs in specs

lidel:
A CID in a spec promises a reader can fetch it and check the text against reality. Nothing enforces that today. #499 showed the cost: the file-over-max-links vector was redefined, the new CID was never pinned, and the spec pointed at unretrievable data until a reader reported it.

- CIDs and content paths added by a PR need CAR coverage attached, scaled to what the text asks of the reader
- partial CARs are acceptable; a 1 GiB fixture whose claim is structural ships as a 51 KB CAR of root and the minimal set of  internal nodes referenced from spec
- prefer an existing tagged CAR in gateway-conformance, codec-fixtures or boxo over attaching a copy
- identity CIDs and deliberately unresolvable CIDs are exempt
- each zip carries a README covering what every CAR verifies
- documents partial CAR import, where kubo's default pin-roots fails on an incomplete DAG

Backfilled existing specs  + added missing ones in:
- [ipfs-specs-fixtures.zip](https://github.com/user-attachments/files/31090406/ipfs-specs-fixtures.zip)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,151 @@
+# Working on ipfs/specs
+
+Guidance for anyone editing this repository, human or agent.
+
+## Every CID and content path needs retrievable data
+
+A specification that names a CID is making a promise: a reader can fetch it and
+see what the text describes. That promise decays. Pinning services lapse, test
+nodes get reinstalled, and a CID that resolved when the PR merged returns
+nothing two years later. The reader is then stuck with a hash and no way to
+check the spec against reality.
+
+So: **a pull request that adds or changes a CID, or a content path such as
+`/ipfs/<cid>/dir/file.html`, must attach a `.zip` containing CAR files that
+cover it.** Attach the zip to the PR description, then link it from the spec
+text as an HTTPS URL so readers can reach the data without the PR.
+
+`.car` is not an allowed GitHub attachment type, so archive the CARs in a
+`.zip`. The limit is 25 MB per attachment.
+
+### How much data to include
+
+Match the coverage to what the text asks of the reader.
+
+| the spec says | ship |
+| --- | --- |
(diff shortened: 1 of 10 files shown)


mishmosh:
Generally looks good and agree with the principle behind this PR. Consider streamlining the edits to reduce cognitive load and make this more usable for first-time users.

**Suggestion:** Split the AGENTS.md edits into a summary + details so contributors can scan the requirement before diving into the details.

Summary:
   - Every CID in a spec must be backed by a CAR file (attached or linked from an upstream repo like gateway-conformance/boxo)
   - Include a README explaining what each CAR verifies
   - Exemptions: identity CIDs and deliberately unresolvable CIDs
   - Point to the worked example ([ipip-0499-test-fixtures.zip](https://github.com/user-attachments/files/31088958/ipip-0499-test-fixtures.zip))

Additional details:
- Producing CARs, partial DAG imports, verification scripts, pinning strategy can go in a later section.

lidel:
Done. Added a numbered rule list at the top, each rule linking to the section that explains it, with the worked example alongside.

The detail below stays long on purpose: `AGENTS.md` is loaded automatically by coding agents, and a rule a weaker model can argue itself out of stops working. It already sat after the requirement, so nothing moved.

Your comment also turned up two gaps: the file only covered fixtures despite its name, so it now says where the live specs are, that `src/` is not plain Markdown, and that an IPIP number is its PR number. And `template.md` was missing the pointer `ipip-template.md` has, which matters since new specs start from it.

I'm going to merge this because adding missing links to CARs with fixtures is way more important than AGENTS.md. 
Feel free to refine AGENTS.md in follow-up PR, this covers what I cared the most (missing CARs).

mishmosh:
Nice, thanks!

Issue #547, opened 2026-08-17
Sunset x-ipfs-path header and unsupported characters


achingbrain:
HTTP header values can only include characters from a limited set, notably not the full unicode smorgasbord.

The HTTP gateway spec doesn't say what to do when the `x-ipfs-path` header would end up containing these characters.

Should they be percent-escaped?  Or should the header be omitted entirely?

lidel:
I'd lean towards percent-escaping, but interop is tricky. 
It may break existing users who do not percent-decode, but still better than skipping I think?

achingbrain:
Percent encoding would be the least lossy way of doing it, though it may disrupt clients.  That said if you try to set a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) value with an invalid character it'll through so clients are already disrupted.

Percent encoding can also inflate header value lengths - what happens as the header value gets longer?  There's no header value limit in the HTTP spec but Apache limits header values to 8Kb, Tomcat limits the whole header to 8Kb.  More modern servers allow more.

lidel:
I've been thinking about it, looked at how this behaves today and I think we should percent-encode the whole value.

IIUC current behavior is already broken is silent ways for subser of non-ASCII paths 🙈 . boxo/Kubo put the decoded path in the header, so non-ASCII goes out as raw UTF-8 bytes, and JS `fetch` reads them back as garbled latin-1. Go's `net/http` also replaces control characters like `%0A` with a space, so data is lost. Nobody gets correct values for non-ASCII paths today, which means changing the format breaks less than it seems.

Proposal: the header carries the path in the same form as it appears in a URL.

- gateways SHOULD percent-encode the value per RFC 3986 (UTF-8 bytes; a literal `%` becomes `%25`)
- clients SHOULD percent-decode it before comparing or displaying
- gateways MAY omit the header if the encoded value is too long for common header size limits (encoding grows non-ASCII up to 3x, and 8K limits are still common)

Why full encoding, not partial escaping or a new format:

- plain ASCII paths do not change at all, so the common case keeps working
- it is lossless: one decode step gets the exact path back, with no guessing about which parts are encoded
- it is future-proof: any filename UnixFS allows (unicode, `%`, control characters) survives the round trip
- it matches the rest of the spec, which already uses RFC 3986 percent-encoding for `filename*`[^1]
- consumers I checked already cope: service-worker-gateway percent-decodes the value, and caches treat it as an opaque key

so, if we go this route, next steps:

[...]

PR #548, opened 2026-08-18
IPIP-548: Sunset x-ipfs-path header

achingbrain:
The `x-ipfs-path` header can contain characters outside of the US-ASCII character set which some clients will refuse to process.

Adding encoding rules to this header could break existing clients, instead add an `ipfs-uri` header that performs the same task but is a URI instead of a path so follows standard percent-encoding rules found in RFC 3986.

Futher discussion can be found in #547.


darobin (approved):

lidel:
(I'll be adding some extra [ipip-0548-test-fixtures.zip](https://github.com/user-attachments/files/31286627/ipip-0548-test-fixtures.zip))

lidel:
Thanks @achingbrain for kicking this off, `Ipfs-Uri` closes a real gap.

The header hid more edge cases than expected (`.`/`..` segments, `%2F`, names that already look percent-encoded, DNSLink IDNs, legacy base58 peer IDs), so I pushed clarifications here (9d187a6, 99afeee): the `Ipfs-Uri` section now builds on the `ipfs://` / `ipns://` URI specs and pins the exact serialization, `X-Ipfs-Path` is legacy-only and omitted when the content path is not ASCII-safe, and the fixtures table covers names across nine alphabets (CAR in the zip above).

Remaining before ratification:

- [x] interop tests: ipfs/gateway-conformance#301 (one test per failure class over three fixtures; the kubo-master report shows stock kubo as not yet conformant, as expected)
- [x] Go impl: ipfs/boxo#1209 (passes the conformance suite; sharness stays red until kubo's CORS sharness drops the legacy header; kubo follows with the next boxo bump)
- [x] JS impl: ipfs/helia-verified-fetch#376 (canonical serialization on top of #375; CI blocked by an unrelated repo-wide build issue)

Both implementations follow the spec as closely as possible; push back or relax if any of it feels too pedantic. I'm also leaning toward dropping `x-ipfs-path` by default and keeping it opt-in (as boxo does): the raw value is unsafe anyway. PTAL

lidel (approved):
Status update:

- gateway-conformance [v0.14.0](https://github.com/ipfs/gateway-conformance/releases/tag/v0.14.0) shipped with the IPIP-548 tests
- ipfs/boxo#1209 and ipfs/kubo#11437 are merged; the header will likely ship in Kubo v0.43.1 this or next week
- ipfs/helia-verified-fetch#376 covers the JS side

@achingbrain if no concerns we can flip this to ratified state and merge?

Issue #533, opened 2026-04-07
Change UnixFS dag-pb protobuf field ordering


achingbrain:
The `.proto` definition of a `PBNode` is this ([spec](https://ipld.io/specs/codecs/dag-pb/spec/#serial-format)):

```proto
message PBNode {
  // refs to other objects
  repeated PBLink Links = 2;

  // opaque user data
  optional bytes Data = 1;
}
```

Implementations are [expected](https://github.com/ipfs/specs/pull/531) to write the repeated `Links` messages to the output buffer first, then the `Data` field, even though the field IDs are ordered the opposite way (not disallowed by the protobuf spec, though some off the shelf encoders will write in ID order).

When the `PBNode` represents a directory, the `Links` objects could be either flat directory entries, or HAMT shard entries - the information needed to ascertain this is contained in the `Data` field.

This means when processing an incoming `PBNode` message, we typically read all of the `Links`, and then use the `Data` to decide how to process them.

If we are performing a graph traversal, a streaming parser would let us process the `PBNode` message as it arrives, and to select the `Link` message we wish to traverse through or resolve to, however this is not currently possible since we need to process the `Data` field before we can return a `Link`.  This can add significant overhead when there are many thousands of `Link`s.

We should allow ordering the `Data` field first, this would enable the streaming use-case and help with IPFS code running in resource-constrained environments such as web browsers.

Since [IPIP-499](https://github.com/ipfs/specs/pull/499) we now have CID profiles which could be an upgrade path for the network.  This proposal could be part of a `unixfs-v1-2026` profile.

PR #550, opened 2026-08-24
IPIP-550: PBNode field ordering

achingbrain:
Having `Link` messages before the `Data` message in `PBNode` messages means for HAMT shards you have to read every `Link` before you can process any of them, since the hash type/fanout values used to calculate the prefix each directory entry has is stored in the `Data` field at the end of the message.

This IPIP suggests allowing writers to encode the `Data` field first, this allows readers to use streaming parsers that can skip out on reading links if they have already reached the directory entry they are interested in.

Further discussion can be found in https://github.com/ipfs/specs/issues/533

## Revised/final state

Scope was reduced during review; the shape being ratified is:

- Reads are the point: readers SHOULD accept both `PBNode` field orders, and implementations interoperating with Mainnet MUST. Covered by tests: [gateway-conformance v0.14.1](https://github.com/ipfs/gateway-conformance/releases/tag/v0.14.1) (released), byte-exact regression tests in boxo (ipfs/boxo#1212) and Kubo (ipfs/kubo#11439), cross-implementation fixtures proposed in ipld/codec-fixtures#212.
- No new profile. `PBNode field order` is a profile parameter defaulting to `links-first`, and the IPIP-0499 tables now state it explicitly, so `unixfs-v0-2015` and `unixfs-v1-2025` keep their bytes and CIDs.
- Writing `Data`-first is a low-level opt-in that no profile enables: shipped in `@ipld/dag-pb` v4.2.0 (`fieldOrder` encode option), merged for the next boxo release and Kubo v0.43.1 (`Import.UnixFSPBNodeFieldOrder`).
- The DAG-PB codec spec permits both encode orders since ipld/ipld#383.
- unixfs.md gains a Profiles registry and test vectors covering both orders.

Nothing changes by default anywhere.

### DIFF
--- a/src/ipips/ipip-0550.md
+++ b/src/ipips/ipip-0550.md
@@ -102,7 +102,9 @@ N/a
 |---|---|---|
 | hello.txt, basic text file | `bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am` | `68656c6c6f0a` |
 | UnixFS directory containing `hello.txt`, with Data field before Links | `bafybeigqvyloizmfcdy6scaxnyltftzptaruqa3hnnplfzsbf4sqteiwlm` | `0a02080112330a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120968656c6c6f2e7478741806` |
+| UnixFS directory containing `hello.txt`, legacy format with Links before Data field | `bafybeigdcg7pksx2zk5336vrfsktjodlr4rbfz37qr3koc5xboxe5ekv24` | `12330a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120968656c6c6f2e74787418060a020801` |
 | HAMT Shard containing `hello.txt`, with Data field before Links | `bafybeicwgy2rlqmqqu3yy2tqvm2wbgdvy3snu4sbbv4wqpvpnoplpzxz74` | `0a250805121c80000000000000000000000000000000000000000000000000000000282230800212350a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120b444668656c6c6f2e7478741806` |
+| HAMT Shard containing `hello.txt`, legacy format with Links before Data field | `bafybeicjwkfslu7gwyywffvqgse5kiibojtktxcdqhgv7ldj5fjdacuceq` | `12350a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120b444668656c6c6f2e74787418060a250805121c800000000000000000000000000000000000000000000000000000002822308002` |
 
 ### Copyright
 
--- a/src/ipips/ipip-0550.md
+++ b/src/ipips/ipip-0550.md
@@ -89,10 +89,8 @@ ordering on read may be considered buggy.
 Two valid encodings for the same data are possible.
 
 - Same content, two CIDs which harms deduplication and CID-based denylists need to include both CIDs
-- Round trips from blocks to data and back change CIDs if the writer does not use the legacy format
+- Round trips from blocks to data and back change CIDs if the writer uses the legacy format
   - This can reveal history of writes in systems like MFS
-
-Note that all of the above also applied to the introduction of metadata in UnixFS v1.5 which was deemed acceptable.
 
 ### Alternatives
 
@@ -100,7 +98,11 @@ N/a
 
 ## Test fixtures
 
-N/a
+| File | CID (unixfs-v1-2025 profile) | Block (base16 encoded)  |
+|---|---|---|
+| hello.txt, basic text file | `bafkreicysg23kiwv34eg2d7qweipxwosdo2py4ldv42nbauguluen5v6am` | `68656c6c6f0a` |
+| UnixFS directory containing `hello.txt`, with Data field before Links | `bafybeigqvyloizmfcdy6scaxnyltftzptaruqa3hnnplfzsbf4sqteiwlm` | `0a02080112330a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120968656c6c6f2e7478741806` |
+| HAMT Shard containing `hello.txt`, with Data field before Links | `bafybeicwgy2rlqmqqu3yy2tqvm2wbgdvy3snu4sbbv4wqpvpnoplpzxz74` | `0a250805121c80000000000000000000000000000000000000000000000000000000282230800212350a24015512205891b5b522d5df086d0ff0b110fbd9d21bb4fc7163af34d08286a2e846f6be03120b444668656c6c6f2e7478741806` |
 
 ### Copyright
 
--- a/src/ipips/ipip-0550.md
+++ b/src/ipips/ipip-0550.md
@@ -49,6 +49,22 @@ a more efficient streaming parser for protobuf messages, since they will no
 longer need to read all of the `Link` messages before they can process any of
 them.
 
+### The `unixfs-v1-2026` profile
+
+Writing the `Data` field first will change the CID generated for a piece of
+content, so this needs to be an opt-in change.
+
+We introduce a new **named configuration profiles** similar those added in
+[IPIP-499](./ipip-0499.md).
+
+It inherits all settings from `unixfs-v1-2025` and adds a new parameter:
+
+| Parameter                     | `unixfs-v1-2026`     |
+| ----------------------------- | -------------------- |
+| Write PBNode Data field first | true (opt-in)        |
+
+If this parameter is not included in a profile it is assumed to be `false`.
+
 ## Design rationale
 
 Traversing HAMT shards is more expensive than it needs to be, which
--- a/src/ipips/ipip-0550.md
+++ b/src/ipips/ipip-0550.md
@@ -1,6 +1,6 @@
 ---
 title: "IPIP-0550: PBNode field ordering"
-date: 2026-08-24
+date: 2026-08-27
 ipip: proposal
 editors:
   - name: Alex Potsides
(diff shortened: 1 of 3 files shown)


src/ipips/ipip-0550.md
```diff
@@ -0,0 +1,85 @@
...
+Traversing HAMT shards will become faster in resource-constrained environments
+and inefficient runtimes.
+
+### Compatibility
+
+Protobuf has no requirement to write fields in any particular order, including
+not in the numerical order of the field IDs, so this is a backwards compatible
+change, unless custom protobuf parsers are used that expect fields defined in a
+certain order. The UnixFS spec does not disallow this so any parsers enforcing
+ordering on read may be considered buggy.
```
lidel: This is a bit tricky: (iiuc) the compatibility argument should be made against DAG-PB, not plain protobuf? The DAG-PB spec already answers half of this and conflicts with the other half[^1]:

- Read side is already covered: DAG-PB requires decoders to accept both orders ("as IPFS data exists in both forms"), and go-codec-dagpb and js-dag-pb both do (they only reject interleaved fields). So the last sentence can go; there are no buggy parsers to call out.
- Write side conflicts: the same rule mandates encoding `Links` before `Data`, so `Data`-first blocks would be readable everywhere but non-canonical DAG-PB.

Open question to answer in this IPIP: do we change the DAG-PB spec ([`ipld/ipld/specs/codecs/dag-pb/spec.md`](https://github.com/ipld/ipld/blob/972879ba6ff34eef01b9d5bda2e897a9d08e0dbd/specs/codecs/dag-pb/spec.md?plain=1#L61)) to match, or state here that `Data`-first blocks are valid UnixFS but non-canonical DAG-PB?

[^1]: [DAG-PB spec, Protobuf Strictness](https://ipld.io/specs/codecs/dag-pb/spec/#protobuf-strictness), rule 2: "Fields in the `PBNode` message must be encoded in the order as defined by the Protobuf schema above. Note that this order does not follow the field numbers. The decoder should accept either order, as IPFS data exists in both forms."

src/unixfs.md
```diff
@@ -111,6 +111,19 @@ More complex nodes use the `dag-pb` (`0x70`) encoding. These nodes require two s
...
+after the repeated `Links` field. This lead to inefficiencies processing HAMT
+shards as all links for a node must be read before the hash type and fanout
+values could be read, which in turns means that if the ready is looking for a
+specific path within the shard, they cannot abort reading links early.
+
+To support legacy data, implementations MUST be able to read and write `PBNode`
+messages in the legacy format as well as the current format described below.
+
+Field IDs were the same in the legacy format.
+:::
```
lidel: "MUST be able to read and write ... the legacy format" makes writing the legacy encoding a conformance requirement for every implementation. Is mandatory legacy write support intended, or should this be relaxed to SHOULD or MAY (read either, write one)?

src/ipips/ipip-0550.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Security
+
+N/a
```
lidel: Two valid encodings for the same node have consequences the Security section should cover:

- Same content, two CIDs. Dedup breaks across the two forms, and CID-based denylists match only one of them.
- Round-trips change CIDs. Tools that read a block and write it back flip the order (go-codec-dagpb always writes `Links` first), so the CID silently changes.
- Field order fingerprints the writer. In a DAG with mixed orders, new-order nodes reveal history: copy a legacy-ordered DAG into Kubo's MFS, edit one file, and only the rewritten nodes flip order, showing an observer which parts were changed and leaving a trail of user activity over time.

Even if the conclusion is that this is acceptable, the IPIP should say so explicitly.

src/ipips/ipip-0550.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Alternatives
+
+N/a
+
+## Test fixtures
+
+N/a
+
```
lidel, line 205: A wire-format change needs fixtures: sample `Data`-first blocks (a basic directory and a HAMT shard) with their CIDs, so implementations can verify decoding and interop tests can cover both orders.

lidel (commented):
Thanks, left some quick first-pass drive-by comments below.

src/ipips/ipip-0550.md
achingbrain: > do we change the DAG-PB spec to match or state here that Data-first blocks are valid UnixFS but non-canonical DAG-PB?

I've [opened a PR](https://github.com/ipld/ipld/pull/383) (nb. different repo) but this might be the most pragmatic way forward.

src/unixfs.md
achingbrain: Updated to MAY

src/ipips/ipip-0550.md
achingbrain: Updated the security section.

src/ipips/ipip-0550.md
lidel: Following up on https://github.com/ipld/ipld/pull/383#issuecomment-5409882629: once writers may emit `Data` first, the same input produces a different CID depending on the writer, which breaks determinism for users of `unixfs-v1-2025` from [IPIP-0499](https://specs.ipfs.tech/ipips/ipip-0499/).

So.. this IPIP should add field ordering as a profile parameter and define a new profile (e.g. `unixfs-v1-2026`) that opts into `Data`-first, with `unixfs-v1-2025` pinned to the legacy order? Existing users must keep their CIDs unless they opt in. Profiles are pre-existing mechanism for controlling and communicating that.

achingbrain: Yep, makes sense.

achingbrain: The profiles are defined in IPIP 499, how should `unixfs-v1-2026` be defined?  An IPIP to the IPIP?

lidel: IPIPs are historical docs, we don’t change them once merged.
I think separate IPIP about describing what `unixfs-v1-2026` is, is the way to go here.

Mind, this is a new feature, and ratification will require JS and GO implementation, conformance tests / fixtures.
Fine to open IPIP tho and park until interoperable impls exist.

achingbrain: I've added a section here about `unixfs-v1-2026`

src/ipips/ipip-0550.md
achingbrain: I've added fixtures - both legacy and `unixfs-v1-2026` style.

lidel:
Pushed 17444ef and b707207: the ordering is now defined per profile. Does this look sensible @achingbrain?

unixfs.md gains a Profiles registry (`unixfs-v0-2015`, `unixfs-v1-2025`, `unixfs-v1-2026`) showing both PBNode orderings with reader/writer rules, and Compatibility now matches DAG-PB codec behavior: decoders already accept both orders, and for UnixFS data this spec takes precedence on the write side.

Working code in Go:

- [gateway-conformance v0.14.1](https://github.com/ipfs/gateway-conformance/releases/tag/v0.14.1) ships this IPIP's fixtures (linked from the Test fixtures section) and tests that gateways serve `Directory` and `HAMTShard` nodes in both field orders (ipfs/gateway-conformance#304).
- Reads already work both ways in Go (go-codec-dagpb accepts either order), so only the write side needed code.
- ipfs/boxo#1212 and ipfs/kubo#11439 add opt-in write support: the `unixfs-v1-2026` profile name defined here plus a dedicated low-level option, verified byte-exact against the fixtures table. Defaults and existing profiles keep their bytes and CIDs.

That makes ratification relatively low-risk: whoever needs the streaming-friendly order can lock it in via the profile, and everyone else, legacy DAGs and libraries included, keeps preexisting behavior.

achingbrain:
Thanks for the updates, all looks good to me.

lidel:
I've opened:

- https://github.com/ipld/go-codec-dagpb/pull/103 but its a nice to have, optional, https://github.com/ipfs/boxo/pull/1212 does not need it, it uses self-contained encoder.
- https://github.com/ipld/codec-fixtures/pull/212 with regression tests to ensure read-interop with Rust etc continues to work. Also optional, boxo and gateway-conformance are enough for GO-JS interop.

lidel:
To keep the scope in check, pushed d99d0ee: dropped the `unixfs-v1-2026` profile from this IPIP.

A dated successor to `unixfs-v1-2025` reads as "the recommended latest" and invites unintentional adoption, effectively a new CIDv1 default, and this IPIP should not change any defaults. Instead, `PBNode field order` is now a regular profile parameter (`links-first` when absent). We should not touch ratified IPIPs, but just for clarity the IPIP-0499 tables state `links-first` explicitly so the preexisting profiles can't drift, and `data-first` writing stays a documented low-level opt-in that no profile enables (ipfs/boxo#1212, ipfs/kubo#11439, ipld/js-dag-pb#111).

The read side is unchanged and remains the priority: both orders formalized and covered by tests (gateway-conformance v0.14.1). Rationale is recorded under Alternatives; 

A future IPIP can still mint a `data-first` profile if real multi-implementation demand appears, and that review deserves to be slow. We just won't do it right now in this IPIP.

lidel:
Status across implementations, and why this is safe to ratify:

- Reads, the main point of this IPIP, already work everywhere and are now specified and tested: both orders are formalized here, exercised end to end in [gateway-conformance v0.14.1](https://github.com/ipfs/gateway-conformance/releases/tag/v0.14.1) (released), locked in with byte-exact regression tests merged in boxo (ipfs/boxo#1212) and Kubo (ipfs/kubo#11439), and proposed as cross-implementation fixtures in ipld/codec-fixtures#212.
- The DAG-PB codec spec now explicitly permits both encode orders (ipld/ipld#383, merged), so the layering conflict is gone.
- Opt-in writes are merged everywhere and default-off: already shipped in `@ipld/dag-pb` v4.2.0 (`fieldOrder` encode option), arriving in the next boxo release and Kubo v0.43.1 (`Import.UnixFSPBNodeFieldOrder`). No profile enables `data-first`, and the preexisting profiles pin `links-first` explicitly.

Nothing changes by default anywhere. Ratifying this documents behavior the ecosystem already relies on and keeps it from regressing.

To make this comment less boring, [this a Lemur](https://cdn.britannica.com/20/154120-050-78DE15C0/lemur.jpg) :point_left:

lidel (approved):

PR #554, opened 2026-09-17
docs: add note on packed/expanded encoding for scalar types

achingbrain:
Protobuf v3 flipped the default encoding type for non-string, non-byte repeated scalar fields from expanded to packed, but the protobuf def here predates all of that.

If you copy the protobuf def as-is and use a vanilla v3 encoder you get bytes that differ from the "official" dag-pb encoders.

Add a note about using non-packed decoding by default for CID stability.

Ref: https://protobuf.dev/programming-guides/encoding/#packed


lidel (approved):
lgtm. while at it, pushed two extra things:

- ddf5490 fix: correct File and Raw enum numbers in filesize rule
- b4ad54e docs: protobuf 3 field presence also changes CIDs

if it looks ok to you, feel free to merge

